PR #7, opened 2020-10-21
Backport internal pull request 64 and 72

davidheryanto:
PR 64:

Select experiment engine > select a client and one or more experiments > select a different client without clearing the client selection intermediately - API response error from Get Variables, that the experiment id is not found. This happens because the client id is first updated which causes the variables list to refresh (with the new client id and old experiment ids), before the experiment list is updated. This is temporarily fixed in the API by ignoring experiment ids that cannot be found when both client id and experiment ids are set, in the call to get experiment variables.

PR 72:

Check that service URLs are resolvable in the Turing router readiness check. This is to prevent the router from accepting requests before ensembler or enricher hosts are resolvable.

### DIFF
--- a/api/turing/service/experiment_service.go
+++ b/api/turing/service/experiment_service.go
@@ -418,8 +418,6 @@ func (es *experimentsService) getExperimentsWithCache(
 	for _, eID := range experimentIDs {
 		if e, found := expIDMap[eID]; found {
 			filteredExperiments = append(filteredExperiments, e)
-		} else {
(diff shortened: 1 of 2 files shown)


davidheryanto:
@krithika369 same as https://github.com/gojek/turing/pull/5 but GitHub somehow having issue with the reference after I update the repo visiblity from private to public

krithika369 (approved):
LGTM!

PR #8, opened 2020-10-21
Backport internal pull request 74

davidheryanto:
- Fix incorrect file path in Turing router Makefile for `build` target, so the router binary is built correctly
- Remove unused references to experiment engines not available in public Turing
- Remove Git configuration to clone private Turing repo now that the repo is public


krithika369 (approved):
LGTM!

PR #9, opened 2020-10-28
Backport internal pull request 79

davidheryanto:
Add Experiment field in the Turing API config to allow custom configuration for experiment runners and managers for different experiment engines.

PR #10, opened 2020-10-28
Refactor test http server in mission control unit test so the test is more deterministic

davidheryanto:
Test HTTP server in mission control unit test is currently started asynchronously in a separate goroutine

https://github.com/gojek/turing/blob/27cb7c65a0d7ef00066749ad64ae7bb244425b73/engines/router/missionctl/mission_control_test.go#L368-L377

There can be cases where this HTTP server is not ready yet but the tests start sending test requests resulting in unexpected test failures such as https://github.com/gojek/turing/pull/9/checks?check_run_id=1320100978

Screenshot from 2020-10-28 20-59-20

This pull request modifies the helper function such that the HTTP server starts in a blocking manner, ensuring that when test requests are sent, the test server is always ready.

> Also added some assertion to check that the HTTP response is not Nil before retrieving the body to make the test error message clearer rather than having panic due to segmentation violation.

### DIFF
--- a/engines/router/missionctl/mission_control_test.go
+++ b/engines/router/missionctl/mission_control_test.go
@@ -15,10 +15,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
+	"net/http/httptest"
 	"path/filepath"
 	"sort"
-	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


krithika369 (approved):
LGTM

PR #12, opened 2020-10-29
Backport internal pull request 81

krithika369:
In the Turing API's `getExperimentsWithCache`, if any experiment id requested is not found, return error.

This method is called by `ListVariables` and this change is a precautionary measure, to have the API return an error if info on any unknown experiment id is requested.


davidheryanto:
LGTM

davidheryanto (approved):

PR #13, opened 2020-11-02
[BUGFIX]: update keto policy to allow calls to /experiment-engines

romanwozniak:
Currently, the `allow-all` policy is only defined on the `"resources:mlp:experiment-engines:**"` resource. 

This doesn't cover calls to the root level resource (`"resources:mlp:experiment-engines"`) and this PR is fixing it.


krithika369 (approved):
LGTM. Thanks for the fix!

PR #14, opened 2020-11-03
Fix for Regression in Router's Outcome logging when response is unsuccessful

krithika369:
Internal MR # 43 changed the Turing router's outcome logging to use the timestamp at the time of response generation, rather than the one generated inside the logging goroutine, which may be delayed. That change introduced a regression since `logTuringRouterRequestSummary` was updated to be called only at the end of all steps, before writing a success response - in cases where one of the steps (Enricher / Router / Ensembler) failed, this method was no longer being called. This PR fixes this regression and adds a unit test case to test the request summary logging.

The changes have been tested on a local router, built from the internal repo.

### DIFF
--- a/engines/router/missionctl/handlers/http_handler.go
+++ b/engines/router/missionctl/handlers/http_handler.go
@@ -85,17 +85,23 @@ func (h *httpHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			err.Error())
 	}
 
-	// Read the request body
-	requestBody, err := ioutil.ReadAll(req.Body)
-	if err != nil {
-		h.error(ctx, rw, errors.NewHTTPError(err))
-		return
-	}
-
 	// Create response channel to store the response from each step. Allocate buffer size = 4
 	// (max responses possible, from enricher, experiment engine, router and ensembler respectively).
 	respCh := make(chan routerResponse, 4)
 	defer close(respCh)
+
+	// Defer logging request summary
+	var requestBody []byte
+	defer func() {
+		go logTuringRouterRequestSummary(ctx, ctxLogger, time.Now(), req.Header, requestBody, respCh)
+	}()
+
+	// Read the request body
+	requestBody, err = ioutil.ReadAll(req.Body)
+	if err != nil {
+		h.error(ctx, rw, errors.NewHTTPError(err))
+		return
+	}
 
 	// Enrich
 	var resp mchttp.Response
@@ -142,9 +148,6 @@ func (h *httpHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		}
 		payload = resp.Body()
 	}
-
-	// Write the result log
-	go logTuringRouterRequestSummary(ctx, ctxLogger, time.Now(), req.Header, requestBody, respCh)
 
 	// Write the json response to the writer
 	rw.Header().Set("Content-Type", resp.Header().Get("Content-Type"))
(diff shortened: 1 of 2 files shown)


engines/router/missionctl/handlers/http_handler_test.go
```diff
@@ -224,6 +253,97 @@ func TestHTTPServiceBadRequest(t *testing.T) {
...
 
+// TestLogRequestSummary checks that copyResponseToLogChannel is called with each piece of
+// info to be logged and finally, logTuringRouterRequestSummary is called.
+func TestLogRequestSummary(t *testing.T) {
+	logUtilsAllComponents := &testLogUtils{}
+	logUtilsBadEnricher := &testLogUtils{}
+
+	// Define check for goroutine called
+	checkLogRequestSummaryCalled := func(t *testing.T, l *testLogUtils) {
+		timeout := time.After(3 * time.Second)           // Wait for a maximum of 3 mins
```
krithika369, line 264: Ran this test case locally a 100 times to check that it is not flaky.

engines/router/missionctl/handlers/http_handler.go
```diff
@@ -85,18 +85,24 @@ func (h *httpHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 
+	// Create response channel to store the response from each step. Allocate buffer size = 4
+	// (max responses possible, from enricher, experiment engine, router and ensembler respectively).
+	respCh := make(chan routerResponse, 4)
+	defer close(respCh)
+
+	// Defer logging request summary
+	var requestBody []byte
+	defer func() {
+		go logTuringRouterRequestSummary(ctx, ctxLogger, time.Now(), req.Header, requestBody, respCh)
```
krithika369, line 96: Still uses the timestamp just after response write.

engines/router/missionctl/handlers/http_handler_test.go
davidheryanto: You mean seconds?

davidheryanto (approved):
I think our tests can be made simpler (e.g without the use of monkey patching and complex setup) if we have implemented http_handler to be more configurable (for instance, the way the result logger is injected inside http_handler). But I guess it's some improvement we can think of.

LGTM

engines/router/missionctl/handlers/http_handler_test.go
```diff
@@ -224,6 +255,97 @@ func TestHTTPServiceBadRequest(t *testing.T) {
...
+				logUtilsAllComponents.AssertCalled(t, "copyResponseToLogChannel", "enricher",
+					`{"value":"Init:Enrich"}`, "")
+				logUtilsAllComponents.AssertCalled(t, "copyResponseToLogChannel", "router",
+					`{"value":"Init:Enrich:Route"}`, "")
+				logUtilsAllComponents.AssertCalled(t, "copyResponseToLogChannel", "ensembler",
+					`{"value":"Init:Enrich:Route:Ensemble"}`, "")
+				checkLogRequestSummaryCalled(t, logUtilsAllComponents)
+			},
+		},
+		"log enricher error": {
```
davidheryanto, line 301: Should we also add tests for errors in other components like Ensembler?

krithika369: Ideally, yes. But I think the way we deal with the error response should generally be the same across each component. Since there is some asynchronous behaviour being tested, I didn't want to add repetitive tests for this.

PR #16, opened 2020-11-04
Add trace for response ensembling in Turing router's FanIn

krithika369:
This PR starts a span from the `Aggregate` method in FanIn, to measure the time taken to ensemble the responses.

**Before:**
Screenshot 2020-11-04 at 11 02 35 AM

<hr/>

**After:**
Screenshot 2020-11-04 at 11 02 24 AM

### DIFF
--- a/engines/router/missionctl/fiberapi/fan_in.go
+++ b/engines/router/missionctl/fiberapi/fan_in.go
@@ -12,9 +12,13 @@ import (
 	"github.com/gojek/turing/engines/router/missionctl/errors"
 	"github.com/gojek/turing/engines/router/missionctl/experiment"
 	"github.com/gojek/turing/engines/router/missionctl/instrumentation/metrics"
+	"github.com/gojek/turing/engines/router/missionctl/instrumentation/tracing"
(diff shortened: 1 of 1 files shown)


krithika369:
@romanwozniak I feel like implementing the tracing for response ensembling in this way is better, as opposed to introducing the ability to add interceptors to `FanIn`, for the following reasons:
* As you mentioned, the interceptors are currently built around the `Dispatch` method in the `Component`, which doesn't fit well with `FanIn`.
* The span does not cover the entire `Aggregate` method, but rather, after the responses are received, around the `collectResponses`. This is custom behavior, outside Fiber's purview anyway.

I suppose that having interceptors for `FanIn` would make (at least parts of) the existing interceptors reusable but I also think that since the implementation of the `FanIn`'s behaviour is up to the user, it is not really important. Let me know if you have other thoughts. Thanks!

romanwozniak:
@krithika369 okay, I gave it another look and it seems reasonable approach for now. LGTM

romanwozniak (approved):

PR #17, opened 2020-11-06
Add Turing Helm chart default values and Helm values documentation

davidheryanto:
Currently Turing Helm chart is missing the default values and documentation about what values can be provided by the user of Helm chart. This makes usage of chart not user friendly.

- This pull request add resonable default Helm values for relevant fields so that the users of the Helm chart does not need to always provide all the values. This will make deployment configuration simpler. 

- Markdown documentation for the Helm values is also provided, using [helm-docs](https://github.com/norwoodj/helm-docs) to render the document.

- Also removed Helm values in `test/e2e/turing.helm-values.yaml` which are already provided by default, in order to simplify configuration

### DIFF
--- a/chart/README.md
+++ b/chart/README.md
@@ -1,0 +1,74 @@
+# turing
+
+![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square)
+
+Turing: ML Experimentation System
+
+## Requirements
+
+| Repository | Name | Version |
+|------------|------|---------|
+| https://charts.bitnami.com/bitnami | postgresql | 8.9.8 |
+
+## Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| dbMigrations.image.tag | string | `"v4.7.1"` | Docker tag for golang-migrate Docker image https://hub.docker.com/r/migrate/migrate |
+| postgresql.persistence.enabled | bool | `true` | Persist Postgresql data in a Persistent Volume Claim  |
+| postgresql.postgresqlDatabase | string | `"turing"` | Database name for Turing Postgresql database |
+| postgresql.postgresqlPassword | string | `nil` | Password for Turing Postgresql database (required) |
+| postgresql.postgresqlUsername | string | `"turing"` | Username for Turing Postgresql database |
+| postgresql.resources | object | `{}` | Resources requests and limits for Turing database. This should be set  according to your cluster capacity and service level objectives. Reference: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ |
+| swaggerUi.apiHost | string | `"127.0.0.1"` | Host for the Swagger UI |
+| swaggerUi.basePath | string | `"/v1"` | Base URL path to serve the Swagger UI |
+| swaggerUi.image | object | `{"tag":"v3.24.3"}` | Docker tag for Swagger UI https://hub.docker.com/r/swaggerapi/swagger-ui |
+| swaggerUi.service.externalPort | int | `8080` | Swagger UI Kubernetes service port number |
+| swaggerUi.service.internalPort | int | `8081` | Swagger UI container port number |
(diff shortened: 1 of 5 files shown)


chart/README.md
```diff
@@ -0,0 +1,74 @@
...
+| turing.config.vault.address | string | `nil` | Vault server address (required) |
+| turing.config.vault.token | string | `nil` | Vault authentication token (required) |
+| turing.image.registry | string | `"docker.io/"` | Docker registry for Turing API image. User is required to override the registry for now as there is no publicly available Turing image |
+| turing.image.repository | string | `"turing"` | Docker image repository for Turing API |
+| turing.image.tag | string | `"latest"` | Docker image tag for Turing API |
+| turing.livenessProbe.path | string | `"/v1/internal/live"` | HTTP path for liveness check |
+| turing.readinessProbe.path | string | `"/v1/internal/ready"` | HTTP path for readiness check |
+| turing.resources | object | `{}` | Resources requests and limits for Turing API. This should be set  according to your cluster capacity and service level objectives. Reference: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ |
+| turing.service.externalPort | int | `8080` | Turing API Kubernetes service port number |
+| turing.service.internalPort | int | `8080` | Turing API container port number |
```
romanwozniak, line 73: It may be a good idea to have a separate section for the required config options. Something like `Configuration` (required values go here) and `Advanced Config Options` (the rest of values go here). 

This will make it easier for users to prepare minimal required values file for their deployments.

chart/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+# It is also currently used to generate README.md using helm-docs
+# https://github.com/norwoodj/helm-docs
+
+turing:
+  image:
+    # -- Docker registry for Turing API image. User is required to override
+    # the registry for now as there is no publicly available Turing image
+    registry: docker.io/
+    # -- Docker image repository for Turing API
+    repository: turing
```
romanwozniak, line 11: Doesn't seem crucial for now, cos we don't publish images yet, but I guess it's going to be `gojek/turing`.
Also, we should consider using ghcr.io as Merlin doesn, instead of docker.io

chart/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+      # -- Enable debugging for Fiber library used by Turing router
+      fiberDebugLog: true
+      # -- Enable custom metrics
+      customMetrics: true
+      jaeger:
+        # -- Enable Jaeger tracing
+        enabled: false
+        # -- Jaeger tracing collector endpoint 
+        collectorEndpoint: 
+      fluentd:
```
romanwozniak, line 63: I feel like this should be hidden behind a feature toggle. I.e. this configuration should not be required if BQ is not available as an available option for logging. Probably can be addressed in a separate MR

romanwozniak (approved):
This is cool :) I left few comments, but it lgtm

chart/README.md
davidheryanto: Yeah we can have another section in the documentation showing example deployments with minimal and advanced configuration. We can add this in another pull request.

I'd actually prefer not to have any required configuration by **default**. Any required configuration should require the user to **specifically enable** specific chart component. For instance, token key for NewRelic is required only if the user specifically enable NewRelic integration. 

But I think we need some refactoring in our config format and loading first.

chart/values.yaml
davidheryanto: Yeah sure, I'm planning to use GitHub container registry as well.

chart/values.yaml
davidheryanto: Yes, I think we can address this in a pull request regarding configuration format in Turing. Agree that all these plugins and extensions should be a feature toggle.

romanwozniak: Can we have tracked this in JIRA? either as a new card or a note to existing one about config refactoring

davidheryanto: Yes sure

PR #18, opened 2020-11-06
Add env var to set allowed origins for accessing the API server

krithika369:
This PR introduces a new env var `ALLOWED_ORIGINS` to the Turing API so that only trusted origins can make requests to the Turing API server via the browser. By default, this value will be set to `*` which is effectively the same as the previous `cors.AllowAll()`.

### DIFF
--- a/api/turing/cmd/main.go
+++ b/api/turing/cmd/main.go
@@ -93,11 +93,11 @@ func main() {
 		web.ServeReactApp(mux, cfg.TuringUIConfig.Homepage, cfg.TuringUIConfig.AppDirectory)
 	}
 
-	c := cors.New(cors.Options{
+	log.Infof("Listening on port %d", cfg.Port)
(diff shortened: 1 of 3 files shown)


romanwozniak (approved):
I'm not entirely ok with `c` being used as a var name here `c := cors.New(...)`, but I assume it comes from the `rs/cors` examples/docs page

krithika369:
> I'm not entirely ok with `c` being used as a var name here `c := cors.New(...)`, but I assume it comes from the `rs/cors` examples/docs page

Sì. I changed it now:
```
        corsHandler := cors.New(cors.Options{
		AllowedOrigins: cfg.AllowedOrigins,
	})
	if err := http.ListenAndServe(cfg.ListenAddress(), corsHandler.Handler(mux)); err != nil {
		log.Errorf("Failed to start turing-api: %s", err)
	}
```

PR #20, opened 2020-11-12
Change Router Prometheus Histogram buckets

krithika369:
This PR changes the Prometheus Histogram buckets defined by the Turing router for capturing custom metrics (request latencies) such that there is more sensitivity at lower values and, at the same time, larger values are accommodated.

### DIFF
--- a/engines/router/missionctl/instrumentation/metrics/prometheus.go
+++ b/engines/router/missionctl/instrumentation/metrics/prometheus.go
@@ -24,12 +24,11 @@ type PrometheusHistogramVec interface {
 }
 
 // requestLatencyBuckets defines the buckets used in the custom Histogram metrics defined by Turing
-var requestLatencyBuckets = append(
-	append(
-		[]float64{2, 4, 6, 8, 10, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100, 120, 140, 160, 180, 200},
-		prometheus.LinearBuckets(250, 50, 15)...,
-	), []float64{1000, 2000, 5000, 10000, 20000, 50000, 100000}...,
-)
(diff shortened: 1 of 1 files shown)


engines/router/missionctl/instrumentation/metrics/prometheus.go
```diff
@@ -23,6 +23,14 @@ type PrometheusHistogramVec interface {
 	GetMetricWith(prometheus.Labels) (prometheus.Observer, error)
 }
 
+// requestLatencyBuckets defines the buckets used in the custom Histogram metrics defined by Turing
+var requestLatencyBuckets = append(
+	append(
+		[]float64{2, 4, 6, 8, 10, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100, 120, 140, 160, 180, 200},
+		prometheus.LinearBuckets(250, 50, 15)...,
+	), []float64{1000, 2000, 5000, 10000, 20000, 50000, 100000}...,
```
romanwozniak, line 31: I'm not sure if there are any benefits of defining it in this way. It's better to just make one array with all the values, so it's easier to understand what buckets are there.

For example, `knative` even add a comment with all the values that would appear here listed, when they use a custom function for generating buckets"
https://github.com/knative/serving/blob/39974804647d382c06e293f9b610435afec1b5cb/vendor/knative.dev/pkg/webhook/stats_reporter.go#L130

romanwozniak (approved):
LGTM

engines/router/missionctl/instrumentation/metrics/prometheus.go
krithika369: I thought it was more readable than having a long list of numbers. But I changed it now.

PR #21, opened 2020-11-12
Fix text overflow bug in the UI panels

krithika369:
This PR fixes a bug in the UI panels where the panel would expand beyond the max screen width when attempting to accommodate a large text.

Screenshot 2020-11-12 at 5 11 56 PM

The problem arises when applying text truncation inside a child flexbox, from the auto-calculated min width of the container exceeding that of the parent which would break the layout. The recommended fix is to set `min-width` to 0 on the child container.

More info: https://css-tricks.com/flexbox-truncated-text/

### DIFF
--- a/ui/src/router/components/configuration/components/section.scss
+++ b/ui/src/router/components/configuration/components/section.scss
@@ -1,5 +1,13 @@
 @import "~@elastic/eui/src/global_styling/mixins/typography";
 
+// When text is truncated in a child flex element, the auto-calculated min width is wider than
+// its parent container which can break the layout. To prevent this, the min-width attribute
+// is expected to be set to a fixed value.
+// Ref: https://css-tricks.com/flexbox-truncated-text/
+.euiFlexItem {
+  min-width: 0;
+}
+
 .euiPanel--configSection {
   .euiDescriptionList.euiDescriptionList--responsiveColumn {
     > * {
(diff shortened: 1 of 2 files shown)


ui/src/router/components/configuration/components/section.scss
```diff
@@ -1,5 +1,13 @@
 @import "~@elastic/eui/src/global_styling/mixins/typography";
 
+// When text is truncated in a child flex element, the auto-calculated min width is wider than
+// its parent container which can break the layout. To prevent this, the min-width attribute
+// is expected to be set to a fixed value.
+// Ref: https://css-tricks.com/flexbox-truncated-text/
+.euiFlexItem {
+  min-width: 0;
+}
```
romanwozniak, line 9: Despite adding this to `ui/src/router/components/configuration/components/section.scss`, this style would be applied to all of the `.euiFlexItem` everywhere in the app, which can potentially lead to unexpected side effects. 

Would be better to at least nest it under `.euiPanel--configSection` or `euiDescriptionList`

romanwozniak (approved):
LGTM

ui/src/router/components/configuration/components/section.scss
krithika369: Thanks for pointing it out. I don't think this can be nested under `.euiPanel--configSection` or `euiDescriptionList` because the style must be applied to the FlexItem that encompasses the ConfigSection. There are many `EuiFlexItem`s in the DOM tree and from playing around, it seems like this property must be set on the second and third level flex items. I added a separate class to these components now.

Screenshot 2020-11-12 at 6 16 05 PM

PR #23, opened 2020-11-17
Introduce env var TEST_ECHO_IMAGE for API e2e tests

krithika369:
This PR copies the relevant changes from internal MR 92, that attempts to minimize the dependency on Docker Hub, in the internal CI / CD pipelines, due to the recent rate limit imposed by Docker. As a result, the env var `TEST_ECHO_IMAGE` has been added to potentially allow the tests to use a different image (from a different repo), among other changes. However, the image name is set to `kennethreitz/httpbin` by default and no changes are done to the Github pipelines at the moment.

### DIFF
--- a/api/e2e/test/00_all_e2e_test.go
+++ b/api/e2e/test/00_all_e2e_test.go
@@ -4,10 +4,11 @@ package e2e
 
 import (
 	"fmt"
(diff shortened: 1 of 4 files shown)


davidheryanto (approved):
LGTM

PR #24, opened 2020-11-18
Add support for file-based configuration in Turing API

davidheryanto:
Currently, Turing API retrieves all configuration values only from environment variables. This is easy to use when we have simple configuration format and reasonable defaults for most configuration keys. As Turing API config format can become more complex (e.g more nesting in the config structure), it can become difficult and error-prone to only use environment variables for setting all the config values. In addition, Turing API config should be able to support configuration for extensions (such as configuration for different experiment engines). Providing values for these extensions which have different config structure can be difficult using only environment variables.

This pull request:
1. Add file-based config for Turing API. The configuration values will be derived in this order (from low to high precedence):
   1. Default configuration values provide reasonable default (e.g. 5432 as the Postgres database port number)
   2. Config values from file (json and yaml are supported). If multiple files are provided, the subsequent config files
       will override the matching keys in previous config files.
   3. Configuration from environment variables. Environment variable name with matching key will override the config value.
       Underscore will be used to separate nested keys in environment variable. E.g config key `DbConfig.Port` can be 
       set with environment variable `DBCONFIG_PORT`.

> Example config file for reference and documentation is provided : https://github.com/davidheryanto/turing/blob/add-support-for-file-based-config/api/turing/config/example.yaml

2. Config files are specified with `-config` flag in Turing cmd. Multiple files can be specified like so 
    ```
    -config=/path/to/config.yaml -config=/path/to/config-2.yaml
    ```

3. Update Turing Helm chart to work with the new file-based config. 
    - The Helm value for `turing.config` will represent the config object for Turing API. The config object format in Helm
      value is the same as the format in the config yaml file. This is to ease usability and make documentation simpler.
    - Add extra fields to allow more custom use cases for e.g.
      - extraArgs: allow users to use multiple config files
      - extraEnvs: allow users to set environment variable to override config value from file
      - extraVolumes: allow users to mount extra secrets already provisioned separately

> This folllows the approach used in https://github.com/hashicorp/vault-helm in order to allow flexible configuration
    when using the Helm chart.


api/turing/config/config.go
```diff
@@ -32,85 +40,100 @@ func (qty *Quantity) Decode(value string) (err error) {
...
 	//
 	// The configuration follows the following format to support different experiment engines
 	// and to allow custom configuration for different engines:
 	// { "<experiment_engine>": <experiment_engine_config>, ... }
 	//
 	// For example:
-	// { "experiment_engine_a": `{"client": "foo"}`, "experiment_engine_b": `{"apikey": 12}` }
-	Experiment map[string]string `envconfig:"experiment"`
+	// { "experiment_engine_a": {"client": "foo"}, "experiment_engine_b": {"apikey": 12} }
+	Experiment map[string]interface{}
```
davidheryanto, line 75: Updated from `map[string]string` to `map[string]interface{}` so that in the config file, this can be specified like so:
```
Experiment:
  expA:
    key: val
```
rather than
```
Experiment:
   expA: {"key":"val"}
```

api/turing/config/config.go
```diff
@@ -122,72 +145,169 @@ type RouterDefaults struct {
...
+	return config, nil
+}
+
+// setDefaultValues for all keys in Viper config. We need to set values for all keys so that
+// we can always use environment variables to override the config keys. In Viper v1, if the
+// keys do not have default values, and the key does not appear in the config file, it cannot
+// be overridden by environment variables, unless each key is called with BindEnv.
+// https://github.com/spf13/viper/issues/188
+// https://github.com/spf13/viper/issues/761
+func setDefaultValues(v *viper.Viper) {
```
romanwozniak, line 255: Just a blunt idea: since we support multiple config files, wouldn't it bu easier, clearer to have a default config file with listed default values that is implicitly loaded before any other user-passed config files?

api/turing/config/config.go
```diff
@@ -32,85 +40,100 @@ func (qty *Quantity) Decode(value string) (err error) {
...
-	TuringEncryptionKey string               `split_words:"true" required:"true"`
-	AlertConfig         *AlertConfig         `envconfig:"alert"`
-	MLPConfig           *MLPConfig
-	TuringUIConfig      *TuringUIConfig `envconfig:"turing_ui"`
+	Port                int `validate:"required"`
+	AllowedOrigins      []string
+	AuthConfig          *AuthorizationConfig
+	DbConfig            *DatabaseConfig   `validate:"required"`
+	DeployConfig        *DeploymentConfig `validate:"required"`
+	RouterDefaults      *RouterDefaults   `validate:"required"`
```
romanwozniak, line 57: General question: is it sufficient to have just a presence check (`required`) for config fields or do we need more thoughtful validation?

romanwozniak (approved):
Added few comments, but overall it looks really cool :)

api/turing/cmd/main.go
```diff
@@ -19,11 +21,34 @@ import (
...
+	var configFlags configFlags
+	flag.Var(&configFlags, "config", "Path to a configuration file. This flag can be specified multiple "+
+		"times to load multiple configurations.")
+	flag.Parse()
+
+	if len(configFlags) < 1 {
+		log.Panicf("Must specify at least one config path using -config")
+	}
+
+	cfg, err := config.FromFiles(configFlags...)
```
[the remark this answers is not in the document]
davidheryanto: Yeah `Load` sounds better. I will change it to `Load`. Was thinking of adding some options to the `Load` function so users can opt out loading from default or env variable but I thought we can leave it for later.

api/turing/config/config.go
davidheryanto: Yes my original intention is actually to have a default config yaml file that is always loaded. This can also serve as a documentation. 

However, if we want to have this default file that is always loaded, when we build the Turing API binary, we want to embed this file in the binary. There are numerous way to do this e.g the good old `go-bindata`. However this will require an additional step in the build process to embed the default file, before running `go build` step. I'd rather not have to modify the build step if possible.

Actually I am kinda waiting for the embed feature to be natively available in go1.16
https://github.com/golang/go/issues/41191

Another way is to have this default specified in say `default.go` like
```go
defaultConfig := `
DbConfig:
  Host: localhost
  Port: 5432
  ... and so on ..
`
```
But I guess this approach is not that much better than our current way of defining defaults with `viper.setDefaults` as both are not very explicit with users having to browse the source code to check the default.

api/turing/config/config.go
davidheryanto: No it's not sufficient. I'll leave this updated in another pull request.
https://github.com/gojek/turing/issues/25

api/turing/config/config.go
romanwozniak: Is it really necessary to embed the default config file? Would it be possible to just keep it side by side with the binary, in the same way as we do with `swagger.yaml` file?

davidheryanto: That actually is possible because we copy the swagger.yaml explicitly in the Dockerfile
https://github.com/gojek/turing/blob/main/api/Dockerfile#L32
And we run Turing binary in the directory where we copy swagger.yaml to.

But I imagine if I only have a Turing API server binary, then it becomes an assumption that the directory where I run this Turing binary from have the respective yaml files, which I think can be a surprise for some people.

davidheryanto: I will merge this in for now. We can add default configuration based on file later. But I believe the default config is usually embeded in the binary itself.

https://github.com/prometheus/prometheus/blob/4eca4dffb89d9da927f1e7884efe85cbd6dc30f3/config/config.go#L71

https://github.com/k0sproject/k0s/blob/14f348158fc8aba111bb074face2338fd2a05849/pkg/apis/v1beta1/cluster.go#L114

PR #26, opened 2020-11-21
Bugfix for Gitops Alert Creation

krithika369:
This PR fixes 3 issues with the Alert creation for Turing Routers.
* `api/turing/service/alert_service.go` - Per the [Prometheus Alert specs](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/), the `groups` key was previously not added. This has been fixed and the unit test case that exercises alert creation in Gitlab has been updated, to compare the content of the alert.
* `api/turing/models/alert.go` - In the internal alerting system used, the Alert Group names are expected to be unique across files. Appending the router name and metric name to the alert group name generated, to achieve this.
* `ui/src/config.js`, `ui/src/router/alerts/edit/EditAlertsView.js` - For the internal set up in Gojek, the alert environment is expected to match `appConfig.environment` (replacing 'dev' with 'development'), as opposed to the router's deployment environment that ties to k8s clusters. This has been fixed. Both this fix and the original implementation may not cater to some OSS users, but perhaps this can be addressed when the need arises.


ui/src/config.js
```diff
@@ -49,3 +49,8 @@ export const sentryConfig = {
 export const monitoringConfig = {
   dashboardUrl: process.env.REACT_APP_ROUTER_MONITORING_URL || ""
 };
+
+export const alertConfig = {
+  environment:
+    appConfig.environment === "dev" ? "development" : appConfig.environment
```
davidheryanto, line 55: I think we should not have this hardcoded environment name conversion logic in the codebase, should just use `appconfig.environment`. Any name conversion from dev to development (or prod to production, p to production etc) should probably be done separately, as these naming conventions usually are specific to diferent organisation/users. We can add a TODO comment or issue for this.

davidheryanto (approved):
LGTM

PR #27, opened 2020-11-21
Add option for default docker registry for use in DockerDeploymentPanel

krithika369:
A previous [PR](https://github.com/gojek/turing/pull/22) set Docker Hub as the default repository for selecting images in the Enricher/Ensembler configuration panels. This PR introduces the environment variable `REACT_APP_DEFAULT_DOCKER_REGISTRY` which can be used to pick a different registry (from the list in `REACT_APP_PRIVATE_DOCKER_REGISTRIES`) as the default, for display in the UI.


ui/src/providers/docker/context.js
```diff
@@ -10,7 +10,7 @@ export const DockerRegistriesContextProvider = ({ children }) => {
       inputDisplay: registry
     })),
     {
-      value: "",
+      value: "docker.io",
```
krithika369, line 13: If using an empty string, we will require the handling logic to change in [SelectDockerImageComboBox](https://github.com/gojek/turing/blob/main/ui/src/components/form/docker_image_combo_box/SelectDockerImageComboBox.js#L6), since the repo name and image are being saved and passed around in the same variable, and an empty string should imply that the default registry should be chosen (in this case, Docker Hub can never be chosen as the registry without entering an image value). This issue is described in more detail in the previous PR: https://github.com/gojek/turing/pull/22

ui/src/config.js
```diff
@@ -36,6 +36,9 @@ export const appConfig = {
   privateDockerRegistries: process.env.REACT_APP_PRIVATE_DOCKER_REGISTRIES
     ? process.env.REACT_APP_PRIVATE_DOCKER_REGISTRIES.split(",")
     : [],
+  defaultDockerRegistry: process.env.REACT_APP_DEFAULT_DOCKER_REGISTRY
+    ? process.env.REACT_APP_DEFAULT_DOCKER_REGISTRY
+    : "docker.io", // User Docker Hub as the default
```
romanwozniak, line 41: `process.env.REACT_APP_DEFAULT_DOCKER_REGISTRY || "docker.io"` could be a better way for specifying default values

romanwozniak (approved):

PR #29, opened 2020-12-03
Get additional info on deployment failures

krithika369:
This PR implements the API and UI changes for storing additional info on deployment failures in the DB and displaying them as a part of the `Activity Log` panel in the Turing UI.

### API

When a Knative service fails to deploy within the configured timeout, we attempt to get one of the following logs from the Kubernetes resources, before purging them:
* `user-container` termination message - When a container fails to become ready and the pod repeatedly restarts, the termination message holds the relevant logs ([ref](https://github.com/knative/serving/blob/master/pkg/reconciler/revision/resources/deploy.go#L151))
* Knative service status condition - In cases where the pod was never created (such as bad image name for the user-container), the Knative service's statuses may contain some useful info. So, we retrieve all the status transitions from the service.

These changes are implemented in `api/turing/cluster/controller.go`.

### UI

Since container termination logs can be long (up to 2048 characters), the Activity Log panel has been modified to allow for collapsing the logs. By default, only the first 3 lines are visible and the rest can be seen by expanding the log. To achieve this, the message is now being wrapped into an `ExpandableContainer`.
* `ui/src/components/expandable_container/ExpandableContainer.js` - Add properties `isScrollable` (decides whether the content is scrollable and a scrollbar should be added) and `toggleKind` (either "button" or "text", which will use `EuiButton` and `EuiLink`, respectively).

**isScrollable=true, toggleKind="button" (default)**
Screenshot 2020-12-07 at 3 56 41 PM

**isScrollable=false toggleKind="text" (event item collapsed)**
Screenshot 2020-12-07 at 3 57 23 PM

**isScrollable=false toggleKind="text" (event item expanded)**
Screenshot 2020-12-07 at 3 57 38 PM

* `ui/src/router/activity_log/events_list/EventsList.js` - Wrap event message into an `ExpandableContainer`
* Styling changes so that the Version History and Activity Log panels are of equal width.

### DIFF
--- a/api/turing/cluster/controller.go
+++ b/api/turing/cluster/controller.go
@@ -20,7 +20,7 @@ import (
 
 	rest "k8s.io/client-go/rest"
 
-	knativeapis "knative.dev/pkg/apis"
+	knapis "knative.dev/pkg/apis"
 	"knative.dev/pkg/kmp"
 	knservingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 	knservingclientset "knative.dev/serving/pkg/client/clientset/versioned"
@@ -491,13 +491,13 @@ func (c *controller) waitKnativeServiceReady(
 		case <-ctx.Done():
 			terminationMessage := c.getKnativePodTerminationMessage(svcName, namespace)
 			if terminationMessage == "" {
-				// Pod was not created (as with invalid image names), get most recent status message
-				// from the knative service
+				// Pod was not created (as with invalid image names), get status message from the knative service.
+				// We expect a condition of type ConfigurationsReady that is not met (IsFalse).
 				svc, err := services.Get(svcName, metav1.GetOptions{})
 				if err != nil {
 					terminationMessage = err.Error()
 				} else {
-					failedCondition := svc.Status.GetCondition(knativeapis.ConditionType("ConfigurationsReady"))
+					failedCondition := svc.Status.GetCondition(knapis.ConditionType("ConfigurationsReady"))
 					if failedCondition != nil && failedCondition.IsFalse() {
 						terminationMessage = failedCondition.GetMessage()
 					} else {
--- a/api/turing/cluster/controller.go
+++ b/api/turing/cluster/controller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/kubernetes"
@@ -20,7 +21,6 @@ import (
 
 	rest "k8s.io/client-go/rest"
 
-	knapis "knative.dev/pkg/apis"
 	"knative.dev/pkg/kmp"
 	knservingv1alpha1 "knative.dev/serving/pkg/apis/serving/v1alpha1"
 	knservingclientset "knative.dev/serving/pkg/client/clientset/versioned"
@@ -491,21 +491,15 @@ func (c *controller) waitKnativeServiceReady(
 		case <-ctx.Done():
 			terminationMessage := c.getKnativePodTerminationMessage(svcName, namespace)
 			if terminationMessage == "" {
-				// Pod was not created (as with invalid image names), get status message from the knative service.
-				// We expect a condition of type ConfigurationsReady that is not met (IsFalse).
+				// Pod was not created (as with invalid image names), get status messages from the knative service.
 				svc, err := services.Get(svcName, metav1.GetOptions{})
 				if err != nil {
 					terminationMessage = err.Error()
 				} else {
-					failedCondition := svc.Status.GetCondition(knapis.ConditionType("ConfigurationsReady"))
-					if failedCondition != nil && failedCondition.IsFalse() {
-						terminationMessage = failedCondition.GetMessage()
-					} else {
-						terminationMessage = "cause unknown"
-					}
+					terminationMessage = getKnServiceStatusMessages(svc)
 				}
 			}
-			return fmt.Errorf("timeout waiting for service %s to be ready: %s", svcName, terminationMessage)
+			return fmt.Errorf("timeout waiting for service %s to be ready:\n%s", svcName, terminationMessage)
 		case <-ticker.C:
 			svc, err := services.Get(svcName, metav1.GetOptions{})
 			if err != nil {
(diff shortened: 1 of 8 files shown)


ui/src/router/activity_log/events_list/EventsList.js
```diff
@@ -30,7 +31,9 @@ export const EventsList = ({ events, status }) => {
           timelineIcon={timelineIcon(event)}
           type="update"
           timestamp={<DateFromNow date={event.created_at} size={"s"} />}>
-          {event.message}
+          <ExpandableContainer maxHeight={43} buttonSize="s">
```
krithika369, line 34: Height=43 shows 3 lines of text nicely.

ui/src/router/components/configuration/components/section.scss
```diff
@@ -4,7 +4,7 @@
 // its parent container which can break the layout. To prevent this, the min-width attribute
 // is expected to be set to a fixed value.
 // Ref: https://css-tricks.com/flexbox-truncated-text/
-.euiFlexItem.euiFlexItem--childFlexPanel {
+.euiFlexItem--childFlexPanel {
```
krithika369, line 7: `.euiFlexItem.euiFlexItem--childFlexPanel` is unnecessarily restrictive, should be enough to look at "euiFlexItem--childFlexPanel".

ui/src/router/history/HistoryView.js
```diff
@@ -6,11 +6,11 @@ import { ListRouterVersionsView } from "../versions/list/ListRouterVersionsView"
 export const HistoryView = ({ router, ...props }) => {
   return (
     <EuiFlexGroup direction="row">
-      <EuiFlexItem grow={5}>
+      <EuiFlexItem grow={1} className="euiFlexItem--mediumPanel">
```
krithika369, line 9: Without the minimum width property applied by  `euiFlexItem--mediumPanel`, Activity Log panel grows to occupy most of the space.

api/turing/cluster/controller.go
```diff
@@ -488,7 +489,23 @@ func (c *controller) waitKnativeServiceReady(
...
+			if terminationMessage == "" {
+				// Pod was not created (as with invalid image names), get most recent status message
+				// from the knative service
+				svc, err := services.Get(svcName, metav1.GetOptions{})
+				if err != nil {
+					terminationMessage = err.Error()
+				} else {
+					failedCondition := svc.Status.GetCondition(knativeapis.ConditionType("ConfigurationsReady"))
+					if failedCondition != nil && failedCondition.IsFalse() {
+						terminationMessage = failedCondition.GetMessage()
```
davidheryanto: Maybe we can use this const for the "ConfigurationsReady" condition type?
https://github.com/knative/serving/blob/7398ed7b7e45bb751e6aa475a8d2f2ab15398e44/pkg/apis/serving/v1/service_types.go#L101

I think the variable name `failedCondition` can be misinterpreted, one would think `failedCondition.IsFalse()` means condition is successful hence the configuration is Ready, when it is actually not.

Can we maybe use `configurationReadyCondition` or something similar as the variable name?

api/turing/cluster/controller.go
```diff
@@ -488,7 +489,23 @@ func (c *controller) waitKnativeServiceReady(
...
+					} else {
+						terminationMessage = "cause unknown"
```
davidheryanto, line 504: In this case, in addition to saying the cause is unknown, can we maybe also provide a bit more context, maybe just listing all the Condition statuses for the Knative service object?

ui/src/components/expandable_container/ExpandableDescriptionList.scss
```diff
@@ -2,7 +2,7 @@
 
 .expandableContainer__childWrapper {
   height: 0;
-  overflow-y: auto;
+  overflow: hidden;
```
krithika369, line 5: Without this, we see unnecessary scroll:

Screenshot 2020-12-03 at 12 32 17 PM

Screenshot 2020-12-03 at 12 30 53 PM

api/turing/cluster/controller.go
krithika369: Thanks, David. I changed this so that we dump all the status conditions from the service if the pod termination message is not available.

ui/src/router/activity_log/events_list/EventsList.scss
```diff
@@ -0,0 +1,3 @@
+.activityLogEvent {
```
krithika369, line 1: This is so that the "\n" in the activity log event is rendered as a newline (every Type.. is a new line).
Screenshot 2020-12-03 at 4 18 14 PM

romanwozniak (commented):
Cool, that you re-used existing component, I've added some comments regarding this. 

But ideally I'd prefer to have this implemented as "Show More" link (similar to this component https://www.npmjs.com/package/react-show-more-text). This is more obvious UX compared to an arrow-styled button next to text.

ui/src/components/expandable_container/ExpandableDescriptionList.scss
krithika369: Yeah the second screenshot is from the Docker Env Variables. If you want to see the scroll there (I thought it was weird to have that and the Expand option), I will apply this style change only for the activity log's expandable container.

ui/src/components/expandable_container/ExpandableContainer.js
```diff
@@ -1,12 +1,23 @@
...
+  EuiSpacer
+} from "@elastic/eui";
 import classNames from "classnames";
 import { useToggle } from "@gojek/mlp-ui";
 
 import "./ExpandableDescriptionList.scss";
 import useDimension from "../../hooks/useDimension";
 
-export const ExpandableContainer = ({ maxHeight = 300, children }) => {
+export const ExpandableContainer = ({
```
krithika369, line 15: Docker env var container will use button toggle + scrollbar. Activity log will use text toggle + no scroll.

ui/src/components/expandable_container/ExpandableDescriptionList.scss
krithika369: Now the scroll behavior is controlled by the presence of the class `scrollableContainer`.

krithika369:
@romanwozniak another look for the UI changes? I think I've addressed all your comments. Updated the PR description with new screenshots as well. Thanks!

ui/src/components/expandable_container/ExpandableContainer.js
```diff
@@ -16,25 +27,38 @@ export const ExpandableContainer = ({ maxHeight = 300, children }) => {
...
-      <div
-        className="expandableContainer__childWrapper"
-        style={{
-          height: isExpanded
-            ? contentHeight
-            : Math.min(contentHeight, maxHeight)
-        }}>
-        <div ref={contentRef}>{children}</div>
-      </div>
+      <EuiFlexGroup direction="row" gutterSize="xs" alignItems="left">
```
romanwozniak, line 30: Do you still need to wrap this into a flex group?

romanwozniak (approved):
LGTM

ui/src/components/expandable_container/ExpandableContainer.js
krithika369: Removed.

PR #31, opened 2020-12-07
Add option for Protobuf format in Router's Kafka Result Logger

krithika369:
This PR adds an option to log events in Protobuf format, when using the Kafka Result Logger in the Turing Router. The env var `APP_KAFKA_SERIALIZATION_FORMAT` may be set to one of 'json' or 'protobuf'.

The `.proto` definition resides in the internal Gojek proto repository and has been copied to `engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto`.

The required Golang classes have been generated using `protoc --go_out=$OUT_DIR ./TuringResultLog.proto` and placed in the same directory.

These changes have been tested end to end (write to a topic in Protobuf format, consume the message and decode) locally.


engines/router/missionctl/log/resultlog/kafka.go
```diff
@@ -130,25 +126,101 @@ func (l *KafkaLogger) write(turLogEntry *TuringResultLogEntry) error {
...
+		EventTimestamp string            `json:"event_timestamp"`
+		Request        requestLogEntry   `json:"request"`
+		Experiment     *responseLogEntry `json:"experiment,omitempty"`
+		Enricher       *responseLogEntry `json:"enricher,omitempty"`
+		Router         *responseLogEntry `json:"router,omitempty"`
+		Ensembler      *responseLogEntry `json:"ensembler,omitempty"`
+	}{
+		RouterVersion:  routerVersion,
+		TuringReqID:    turingReqID,
+		EventTimestamp: e.timestamp.Format(time.RFC3339Nano),
```
krithika369, line 151: This was `Ts` earlier, to match the BQ schema. But changing this to `EventTimestamp` to match the internal Kafka message convention.

engines/router/missionctl/turingctx/context.go
```diff
@@ -23,6 +23,12 @@ func NewTuringContext(parent context.Context) context.Context {
 	return context.WithValue(parent, turingReqIDKey, reqID.String())
 }
 
+// NewTestTuringContext returns a context for testing, with the Turing Request ID set to
+// the given value
+func NewTestTuringContext(parent context.Context, turingReqID string) context.Context {
```
krithika369, line 28: Add utility method for use in `TestNewProtobufKafkaLogEntry`.

davidheryanto:
@krithika369 Is there a reason why we don't include the proto definition in this repo? The proto file is a good documentation for the message schema, and allows us to easily review and update the schema.

krithika369:
@davidheryanto makes sense. I will copy it over here alongside the `.pb.go`, once the internal MR with the proto definition is reviewed and merged.

engines/router/missionctl/log/resultlog/kafka.go
davidheryanto: I think it's good if the default behaviour of our various log exporters to different sinks (Kafka, BigQuery, Fluentd) is consistent with the schema of our log message.

If we are to use this `TuringResultLog.proto` as the schema of our log message, 
```proto
message TuringResultLogMessage {
  string turing_req_id = 1;
  google.protobuf.Timestamp event_timestamp = 2;
  string router_version = 3;
  Request request = 4;
  Response experiment = 5;
  Response enricher = 6;
  Response router = 7;
  Response ensembler = 8;
}
```

I think it is reasonable to expect that if someone choose BigQuery as the LogSink, there should be a column called `event_timestamp` rather than having it automatically renamed to `ts`.

engines/router/.env.development
```diff
@@ -29,6 +29,7 @@ APP_FLUENTD_TAG=response.log
 # For result logging to Kafka
 APP_KAFKA_BROKERS=localhost:9093
 APP_KAFKA_TOPIC=turing-result-log-test
+APP_KAFKA_SERIALIZATION=json
```
davidheryanto, line 32: Does `APP_KAFKA_SERIALIZATION_FORMAT` sound reasonable?
I think it conveys the meaning better

engines/router/missionctl/log/resultlog/kafka.go
```diff
@@ -130,25 +126,101 @@ func (l *KafkaLogger) write(turLogEntry *TuringResultLogEntry) error {
...
-	logEntry := &kafkaLogEntry{
-		RouterVersion: routerVersion,
-		TuringReqID:   turingReqID,
-		Timestamp:     e.timestamp.Format(time.RFC3339Nano),
-		Request:       e.request,
-		Experiment:    getTuringResponseOrNil(e.responses, ResultLogKeys.Experiment),
-		Enricher:      getTuringResponseOrNil(e.responses, ResultLogKeys.Enricher),
-		Router:        getTuringResponseOrNil(e.responses, ResultLogKeys.Router),
-		Ensembler:     getTuringResponseOrNil(e.responses, ResultLogKeys.Ensembler),
+	kafkaMessage := &struct {
```
davidheryanto, line 139: Since we are using proto3 which supports canonical encoding in JSON https://developers.google.com/protocol-buffers/docs/proto3#json

Should we use the same log message object (defined in the proto), then convert it to JSON using the common proto tooling? Instead of redefining the JSON structure manually, which we must always ensure is
consistent with the proto message definition.

engines/router/missionctl/log/resultlog/kafka.go
krithika369: I've added a card to track this internally. We need to handle this update for existing users of BQ logging.

engines/router/.env.development
krithika369: Sure, updated it.

engines/router/missionctl/log/resultlog/kafka.go
krithika369: The request/response bodies were initially defined as `string` fields in the Proto, to avoid having to unmarshal them into a struct. And if using string, there is additional escaping when we marshal the proto structure's fields to json. So I decided to keep the 2 separate.

But I've now updated the  request/response body fields to use `google.protobuf.Struct`, and both the JSON and Protobuf logging to use the same `TuringResultLogMessage` struct.

engines/router/missionctl/log/resultlog/kafka_test.go
```diff
@@ -84,60 +85,74 @@ func TestNewKafkaLogger(t *testing.T) {
...
-		Ensembler: nil,
-	}, *logEntry)
+			"experiment": {
+				"response": {"key": "experiment_data"}
+			},
+			"enricher": {
+				"response": {"key": "enricher_data"}
+			},
+			"router": {
+				"response": {"key": "router_data"}
```
davidheryanto, line 115: Do you think we should add a test case where error is not empty?

davidheryanto (approved):
LGTM

Good to also have some comments for the proto message in the .proto file describing what is the message used for.

engines/router/missionctl/log/resultlog/kafka_test.go
krithika369: Done

PR #32, opened 2020-12-08
Add Serialization property to Kafka Config in Turing API

krithika369:
This PR adds the `SerializationFormat` property to the Kafka Config processed by the Turing API, which will allow for setting this value on the Turing Router (implemented in [PR](https://github.com/gojek/turing/pull/31)).

Originally, Kafka result logging was only supported in JSON format. With this change, result logging to Kafka topics can be configured for one of 'json' or 'protobuf' formats.

**Note on DB Migration:**
When these changes are integrated and deployed, the data in the existing Turing database would need to be updated to set `serialization_format: "json"` in the Kafka Config, for the router versions using Kafka as the result logger.

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -1095,12 +1095,18 @@ definitions:
     required:
     - brokers
     - topic
+    - serialization
     properties:
       brokers:
         type: "string"
         description: Comma-separated list of host and port pairs that are the addresses of the Kafka brokers.
       topic:
         type: "string"
+      serialization:
+        type: "string"
+        enum:
+        - "json"
+        - "protobuf"
 
   Alert:
     type: "object"
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -1095,14 +1095,14 @@ definitions:
     required:
     - brokers
     - topic
-    - serialization
+    - serialization_format
     properties:
       brokers:
         type: "string"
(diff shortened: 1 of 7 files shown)


romanwozniak (approved):
LGTM. I'm not quite clear on that, but shouldn't you add a migration here: https://github.com/gojek/turing/tree/main/api/db-migrations so it's applied in E2E test?

romanwozniak:
I just noticed that the same `SerializationFormat` constants are defined in `engines/router/missionctl/config/config.go` in https://github.com/gojek/turing/pull/31/files. Can we consider re-using the same types/constants to avoid any possible discrepancies?

krithika369:
> LGTM. I'm not quite clear on that, but shouldn't you add a migration here: https://github.com/gojek/turing/tree/main/api/db-migrations so it's applied in E2E test?

The DB columns remain the same only the value in `router_versions.log_config.kafka_config` needs to be updated. E2E tests are not affected by this.

krithika369:
> I just noticed that the same `SerializationFormat` constants are defined in `engines/router/missionctl/config/config.go` in https://github.com/gojek/turing/pull/31/files. Can we consider re-using the same types/constants to avoid any possible discrepancies?

Can. I will make another PR to clean it up along with `ResultLogger` types and some other mentions of experiment engine types.

PR #34, opened 2020-12-09
Add GitHub Action for Helm chart release

davidheryanto:
Add GitHub action to release Helm chart. Action is based on
https://github.com/helm/chart-releaser-action

On every push to master, if the Helm chart version is updated, a new GitHub release named `turing-chart-VERSION` will be created. The `index.yaml` for the Helm chart repository will also be updated to reflect this new version. The `index.yaml` will be served using GitHub pages. Once merged, this Helm chart repo will be added to https://artifacthub.io (currently the official site for publishing Helm charts)

Example of Turing chart published on staging artifacthub.
https://staging.artifacthub.io/packages/helm/repo1607517787/turing

The requirements.yaml file is removed because currently the postgresql chart dependency is already embedded in this repo so the requirements.yaml file is not required. Having that file will fail the GitHub action to publish the Helm chart.

### DIFF
--- a/.github/workflows/helm-chart.yaml
+++ b/.github/workflows/helm-chart.yaml
@@ -1,0 +1,34 @@
+name: helm-chart
+
+on:
+  push:
+    branches:
+    - master
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Install Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.0
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.1.0
+        env:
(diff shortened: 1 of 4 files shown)


krithika369:
This is cool! 

Question - What does "if the Helm chart version is updated," mean? Does this mean when the [version](https://github.com/gojek/turing/blob/main/infra/chart/Chart.yaml#L4) in `Chart.yaml` is updated or when there is any update to the chart definition?

davidheryanto:
The [chart releaser](https://github.com/helm/chart-releaser) tool can detect any changes in the chart templates since the last release. If changes are detected (e.g. we change the default replica value in the template from 1 to 2), the chart releaser tool will release the new Chart with version defined in the `Chart.yaml` file.

Normally, every time the chart templates are updated, one should always update the version in `Chart.yaml` as well. When this happens, the chart-releaser tool will work as expected, releasing a new version of the chart.

If the chart templates are updated, but the version is unchanged, the GitHub action will fail when trying to release the chart, because the release for the version already exists. 

We can correct this by updating the version in `Chart.yaml` 

_In rare case, when we want to **replace** the chart that was previously released (hence the version is unchanged even with changes in chart templates), we can manually delete the Chart release in GitHub and re-run the GitHub action_.

krithika369 (approved):
LGTM

PR #35, opened 2020-12-16
Update revision variable for Grafana Dashboard 

davidheryanto:
When the linked Grafana dashboard from Turing UI support revision variables with _All_ selection, the linked dashboard will now:
- Show metrics for all revision when current page is at router level.
- Show metrics for a specific revision when current page is at router version level.


ui/src/hooks/useMonitoring.js
```diff
@@ -17,6 +17,9 @@ const getMonitoringLink = (
   routerVersion
 ) => {
   templateSettings.interpolate = /{{([\s\S]+?)}}/g;
+  // If current page is at router details (not viewing a specific router version), the monitoring link will show
+  // metrics for all revisions. Else, the monitoring link will show metrics only for a "specific" revision.
+  routerVersion = /routers\/\d+\/details/.test(window.location.pathname) ? "$__all" : routerVersion
```
romanwozniak, line 22: This looks unnecessarily hack-ish. 

I'd suggest to just update the contract of `getMonitoringLink` and allow `routerVersion` to be `undefined`. And when it's `undefined`, then use `$__all` for a version param.

Then, you can change the calling side and remove `config.version` here:
https://github.com/gojek/turing/blob/b5da2ee2ef5c480e7a6ddbc44b60fd574e31373e/ui/src/router/details/components/page_navigation/RouterDetailsPageNavigation.js#L40

romanwozniak (changes requested):

ui/src/hooks/useMonitoring.js
davidheryanto: Updated

ui/src/hooks/useMonitoring.js
```diff
@@ -33,7 +33,7 @@ export const useMonitoring = () => {
     (envName, routerName, revision) => {
       const clusterName = getEnvironmentCluster(envName, environments);
       const projectName = !!project ? project.name : undefined;
-
+      revision = !revision ? "$_all" : revision
```
krithika369, line 36: Perhaps this line can be absorbed into the function signature with a default argument?
```
const getMonitoringLink = (
  clusterName,
  projectName,
  routerName,
  routerVersion="$__all"
)
```

krithika369 (approved):
Left one small suggestion. LGTM.

romanwozniak (approved):

ui/src/hooks/useMonitoring.js
davidheryanto: Ok sure

PR #38, opened 2020-12-22
Add config to set alert playbook URL

davidheryanto:
This playbook URL will be set as one of the annotations in the alert spec, which can then be used to provide extra information to the alert recipient.

### DIFF
--- a/api/turing/api/appcontext.go
+++ b/api/turing/api/appcontext.go
@@ -104,6 +104,7 @@ func NewAppContext(
 		}
 		appContext.AlertService = service.NewGitlabOpsAlertService(db, gitlabClient, cfg.AlertConfig.GitLab.ProjectID,
 			cfg.AlertConfig.GitLab.Branch, cfg.AlertConfig.GitLab.PathPrefix)
+		service.AlertPlaybookURL = cfg.AlertConfig.PlaybookURL
 	}
 
 	// Initialize OpenAPI validation middleware
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -173,8 +173,9 @@ type VaultConfig struct {
 }
 
 type AlertConfig struct {
-	Enabled bool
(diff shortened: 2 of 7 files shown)


api/turing/api/appcontext.go
```diff
@@ -104,6 +104,7 @@ func NewAppContext(
 		}
 		appContext.AlertService = service.NewGitlabOpsAlertService(db, gitlabClient, cfg.AlertConfig.GitLab.ProjectID,
 			cfg.AlertConfig.GitLab.Branch, cfg.AlertConfig.GitLab.PathPrefix)
+		service.AlertPlaybookURL = cfg.AlertConfig.PlaybookURL
```
krithika369, line 107: This feels a bit ad-hoc to me. I guess `AlertPlaybookURL` is not a part of `gitlabOpsAlertService` because it is common to other types of alert service implementations? Can setting this be made the responsibility of the constructor in the Alert Service? So in the appcontext init we just do:
```
if cfg.AlertConfig.Enabled {
    appContext.AlertService, err := serice.NewAlertService(db, cfg.AlertConfig)
}
```
I guess this will also be closer to how we initialize the experiment managers in `NewExperimentsService`. Wdyt?

krithika369 (approved):
Left one comment. Other than that it LGTM.

krithika369:
@davidheryanto did you mean to merge this PR? Not sure if you noticed the comment: https://github.com/gojek/turing/pull/38#discussion_r547060572 :)

api/turing/api/appcontext.go
davidheryanto: yeah sounds good. I will update that in another PR

davidheryanto:
I will add that in another PR. I tried auto merge feature in this PR so it was automatically merged before I address your comment :)

krithika369:
@davidheryanto ah I see.. Will watch out for that in the future :)

PR #36, opened 2020-12-16
Update packages with vulnerability as reported by npm

davidheryanto:
Screenshot from 2020-12-16 11-15-49

Update applied with: `npm audit fix --force`

Along with the upgrade there are lint errors. Add fixes for these lint errors:
- import/no-anonymous-default-export
- react-hooks/exhaustive-deps

PR #39, opened 2020-12-22
Fix UI error due to invalid type

davidheryanto:
Pull request https://github.com/gojek/turing/pull/36 updates `react-scripts` package to v4, which enables eslint rule `react-hooks/exhaustive-deps`.

To conform to this rule, this patch is added:

```diff
- const rootHandler = useCallback(new StackableFunction([], handleChanges), [handleChanges]);
+ const rootHandler = useCallback(() => new StackableFunction([], handleChanges), [handleChanges]);
```
But this introduces UI error when editing a router:
Screenshot from 2020-12-22 11-40-30

This PR updated the patch to use `useMemo` instead of `useCallback`

### DIFF
--- a/ui/src/components/form/context.js
+++ b/ui/src/components/form/context.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from "react";
+import React, { useCallback, useState, useMemo } from "react";
 import { set } from "./utils";
 import { StackableFunction } from "./functions/stackable_function";
 import { useOnChangeHandler } from "./hooks/useOnChangeHandler";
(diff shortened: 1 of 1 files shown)


krithika369:
Thanks, @davidheryanto. I'm trying to refactor this so we don't have to ignore this check. In the meantime, this LGTM.

krithika369 (approved):

ui/src/components/form/context.js
```diff
@@ -19,7 +19,9 @@ export const FormContextProvider = ({ data: initData, ...props }) => {
     [setData]
   );
 
-  const rootHandler = useCallback(() => new StackableFunction([], handleChanges), [
+  // TODO: The eslint rule below is disabled. Consider refactoring to conform to the rule.
```
krithika369, line 22: @davidheryanto after checking the code, my understanding is that we want the `new StackableFunction` object to be:
* The argument to the `useOnChangeHandler` call in the line below ([ref](https://github.com/gojek/turing/blob/main/ui/src/components/form/context.js#L26))
* Recomputed (i.e., recreated) only when `handleChanges` is updated (so that the `useEffect` defined inside the `useOnChangeHandler` hook will not be continuously triggered)

So, instead of `useCallback` which returns a callback function, I believe `useMemo` should be used:
```
import React, { useCallback, useMemo, useState } from "react";
...
const rootHandler = useMemo(() => new StackableFunction([], handleChanges), [
    handleChanges
  ]);
```

This passes the ES lint checks and fixes the bug as well. @romanwozniak wdyt?

romanwozniak: looks legit :)

PR #40, opened 2020-12-22
Refactor Alert Service Initialization

davidheryanto:
- Alert service constructor now accepts config object rather than many arguments that makes initialization error prone
- Having config object in the alert service allows additional alert config to be added without changing the constructor. For example this allows the alert service to reference the playbook URL in the alert config.
- Init GitLab client in alert service constructor so the appcontext has more focused responsibility


krithika369 (approved):
LGTM. Thanks!

PR #41, opened 2020-12-22
Update aggregation period for alerts to match the period in the dashboard

davidheryanto:
Currently our dashboard shows router metrics for error rate, latency, throughput with an aggregation window of 1m. However the alert was based on aggregation window of 5m. What is shown in the dashboard, therefore, may not correspond to the condition for the alert to be active. This PR matches the aggregation window used in both alert and dashboard to 1m.


krithika369 (approved):
LGTM!

romanwozniak (approved):

PR #42, opened 2021-01-04
Update alert service to include dashboard URL in the generated alert

davidheryanto:
Update `AlertService` so that the generated alert has a link to the Grafana dashboard to debug the triggered alert. This dashboard link can be configured from the Turing API config.

- Update `AlertService` interface
  - `Save,Update,Delete` methods now require router argument because router info is needed to get project and environment name for alert dashboard template values
  - Add `MLPService` dependency to `AlertService`. MLP service is needed to fetch project and environment for the given router alert
  - Add `GetDashboardURL` method that returns the dashboard link for an alert

- Add `DashboardURLTemplate` to Turing API config. This template will be used to generate dashboard link for the alert


krithika369:
@davidheryanto Instead of adding `MLPService` as a dependency of `AlertService`, can we deal with the required operations at `AlertsController`? Reasons:
* `AlertsController` already has a reference to the `MLPService` and we are already doing things like [getProjectFromRequestVars](https://github.com/gojek/turing/blob/main/api/turing/api/alerts_api.go#L27) in the Alert API handlers (which is currently ignored).
* Letting the API handlers interface with multiple services (instead of one service to another) is in line with the current implementations. For example, to build a router model and then deploy it, we need the `MLPService`, `CryptoService`, `EventsService`, etc., but all of these are done in the controllers and the services themselves are focussed.
* This will also prevent the need for moving `PodLog` to `models` package (originally, `models` package was meant to be focussed on the DB representations. `PodLog` doesn't fall in this category.)

What do you think?

davidheryanto:
OK I will move the MLP service to AlertController. I was under the impression that the controller package is for REST controller and hence should simply does REST related functions, like parsing request and response data. But since you pointed out that it wasn't the case, then I think we can do like what you suggested.

davidheryanto:
@krithika369 I removed MLPService dependency from AlertService and move PodLog back to service package.

api/e2e/test/01_create_router_test.go
```diff
@@ -4,12 +4,12 @@ package e2e
 
 import (
 	"fmt"
+	"github.com/gojek/turing/api/turing/service"
```
krithika369, line 7: Normally, the `goimports` linter would complain about this ordering. I just realised that `make lint` doesn't include the e2e directory - `golangci-lint` only works when there is a .mod file in the root, for the files in that module and the code is compilable. Not sure that there is a way to get this working for the e2e folder at the moment.

I usually have goreturns or goimports enabled in the IDE for auto-formatting on save file.

api/e2e/test/helpers_api_test.go
```diff
@@ -6,13 +6,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"github.com/gojek/turing/api/turing/service"
```
krithika369, line 9: Same comments as api/e2e/test/01_create_router_test.go.

Why has this been changed here? Can we run `goimports` on this file? May be the `make fmt` target can be adjusted to include this directory.

api/turing/api/alerts_api.go
```diff
@@ -104,7 +112,11 @@ func (c *AlertsController) UpdateAlert(r *http.Request, vars map[string]string,
 	updateAlert := body.(*models.Alert)
 	updateAlert.ID = alert.ID
 	updateAlert.Service = c.getService(*router)
-	if err := c.AlertService.Update(*updateAlert, email); err != nil {
+	dashboardURL, err := c.getDashboardURL(c.AlertService.GetDashboardURLTemplate(), alert, router, nil)
```
krithika369, line 115: Do you think we can get the `project`, `environment`, `router` and `routerVersion` in the controller and call `AlertService` to get the dashboard URL directly instead of getting the template from Alert Service and formatting it here?

So in the AlertService interface, we can have something like:
```
    GetDashboardURL(
        alert *models.Alert,
        project *mlp.Project,
        environment *merlin.Environment,
	router *models.Router,
	routerVersion *models.RouterVersion,
    )
```
instead of
```
    GetDashboardURLTemplate() template.Template
```

I think this may be more re-usable if we are populating the Dashboard URL from the API in other responses as an annotation (say, Router / Router Version info will append the dashboard link for the UI to use as well), then this `AlertService.GetDashboardURL` may be called from other controllers with the necessary DB model data, most of which they may already have. wdyt?

api/turing/service/alert_service.go
```diff
@@ -17,17 +18,30 @@ import (
...
+	config               config.AlertConfig
+}
+
+// dashboardURLValue will be passed in as argument to execute dashboardURLTemplate.
+type DashboardURLValue struct {
+	Environment string // environment name where the router is deployed
+	Cluster     string // Kubernetes cluster name where the router name is deployed
+	Project     string // MLP project name where the router is deployed
+	Router      string // router name for the alert
+	Revision    string // router revision number
```
krithika369, line 44: May be `Version` is better to use here instead of `Revision`, since the Turing App UI uses that terminology? It also corresponds to the RouterVersion DB model's Version field.

api/turing/api/alerts_api.go
krithika369: Btw, will there be another PR for letting the UI fetch the dashboard link from the API so we can remove [this](https://github.com/gojek/turing/blob/main/ui/src/hooks/useMonitoring.js#L13)?

krithika369 (commented):
Thanks for the changes @davidheryanto. This mostly LGTM. Left some small comments about formatting and the placement of GetDashboardURL(). The PR description is a bit outdated.

api/e2e/test/01_create_router_test.go
davidheryanto: I've updated the golangci.yml config file to include Go source with e2e build tags, so the linting will run on e2e package as well.

api/turing/api/alerts_api.go
davidheryanto: OK I've updated AlertService with GetDashboardURL method.

Yes for the UI to be able to get the dashboard URL from the server, we agreed to add the annotations field to Router and RouterVersion object. But this requires db migration and changes in schema. So I think it's better to have it in a separate pull request.

api/e2e/test/00_all_e2e_test.go
```diff
@@ -16,9 +16,9 @@ import (
 const (
 	deploymentWaitTimeoutSeconds = 200
 	deleteTimeoutSeconds         = 20
-	fluentdSyncIntervalSeconds   = 10
+	fluentdSyncIntervalSeconds   = 10 // nolint:deadcode,varcheck,unused
```
krithika369, line 19: Since these vars are not used in the OSS version of the tests, do you think we can just remove them?

api/e2e/test/helpers_bigquery_test.go
```diff
@@ -17,6 +17,8 @@ import (
 //
 // If at least <minRowCount> rows is returned, the no of rows retrieved will be returned.
 // Else if timeout is reached, error will be non nil.
+//
+// nolint:deadcode,unused
```
krithika369, line 21: Same comment as above - can we simply remove this? The e2e folder is replicated internally anyway.

krithika369 (approved):
Left some minor comments/questions on clean up. The rest LGTM. Thanks @davidheryanto !

api/turing/service/alert_service.go
krithika369: @davidheryanto Not sure you noticed this earlier comment. It's a small thing. I personally think `Version` is better here but I don't have a strong opinion.

api/e2e/test/01_create_router_test.go
krithika369: I didn't realise this was possible. Thanks!

api/turing/service/alert_service.go
davidheryanto: OK yeah sure version sounds more consistent

api/e2e/test/00_all_e2e_test.go
davidheryanto: OK I'll remove it

davidheryanto:
@krithika369 Renamed revision to version and removed the extra nolint annotation

krithika369:
Thanks, @davidheryanto. LGTM.

PR #43, opened 2021-01-05
Prevent caching of index.html when API serves the React App

krithika369:
Cache busting in React Apps is typically achieved by the following:
* `build/static/` directory with the CSS, Js and other static files contain a hash of the file contents as the file name suffix. This is handled by `npm run build` for the CRA. ([ref CRA doc](https://create-react-app.dev/docs/production-build))
* `index.html` - this file references the static assets and it is advised that this file is not cached, to prevent serving the older versions of an app ([ref Github Issue](https://github.com/facebook/create-react-app/issues/1910))

For the latter, Chrome usually adds the `max-age=0` header in the request, which, although supposed to trigger revalidation, does not guarantee it. Further, navigations in the same browser sessions, usage of forward/back buttons, etc. may not usually trigger this validation.

Screenshot 2021-01-05 at 10 43 23 AM

To handle these scenarios, web servers usually set the following cache directives in the response headers:
* must-revalidate - If a cache is fresh, use it and if not, validation MUST be triggered
* no-cache - Page is cacheable but will not be used. Every request triggers a revalidation, even if the cached resource is fresh.
* no-store - Page not cacheable

Per the CRA docs, the `no-cache` directive will suffice in our scenario. But, it is a common (conservative) practice to set a combination of these headers, including `expires=0` ([ref SO answer](https://stackoverflow.com/a/2068407)).

Screenshot 2021-01-05 at 12 19 22 PM

This PR adds the response headers described above to the React App files served by the API.

### DIFF
--- a/api/turing/web/web.go
+++ b/api/turing/web/web.go
@@ -7,6 +7,9 @@ import (
 
 func FileHandler(path string) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
+		// Ref: https://create-react-app.dev/docs/production-build/#static-file-caching
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
+		w.Header().Set("Expires", "0") // For proxies
 		http.ServeFile(w, r, path)
 	}
 
--- a/api/turing/web/web.go
+++ b/api/turing/web/web.go
@@ -5,11 +5,13 @@ import (
 	"path"
 )
 
-func FileHandler(path string) http.Handler {
+func FileHandler(path string, disableCaching bool) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
-		// Ref: https://create-react-app.dev/docs/production-build/#static-file-caching
-		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
-		w.Header().Set("Expires", "0") // For proxies
+		if disableCaching == true {
+			// Ref: https://create-react-app.dev/docs/production-build/#static-file-caching
+			w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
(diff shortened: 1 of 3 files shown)


davidheryanto:
Should we disable caching of index.html only rather than all the served files (bandwith and client response time can be improved if larger javascript and css files are cached). For other assets like js and css files, I believe every build will append a unique hash to the filenames so accessing old assets should not be an issue?

davidheryanto (approved):

krithika369:
> Should we disable caching of index.html only rather than all the served files (bandwith and client response time can be improved if larger javascript and css files are cached). For other assets like js and css files, I believe every build will append a unique hash to the filenames so accessing old assets unexpectedly should not be an issue?

I may be missing something here. But I had the same question and tested this on Chrome and it seems that only the document type is re-fetched with every hit of the refresh button (as we want) and not the .css, .js files (the response header's date shows the older date for these files, say, from when a hard refresh was done).

davidheryanto:
OK sure then we can leave it as it is :)

api/turing/web/web.go
```diff
@@ -7,6 +7,9 @@ import (
 
 func FileHandler(path string) http.Handler {
 	fn := func(w http.ResponseWriter, r *http.Request) {
+		// Ref: https://create-react-app.dev/docs/production-build/#static-file-caching
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
+		w.Header().Set("Expires", "0") // For proxies
```
romanwozniak, line 12: This `FileHandler` function is generic one and is not specific to React or CRA. For example, it's also being used for serving `swagger.yaml` file. 

It would be better to parametrize it  with either `disableCahche bool` or `headers http.Header` arguments.

PR #44, opened 2021-01-08
Update Turing Result Log proto

krithika369:
This PR updates the existing `TuringResultLog.proto` with some breaking changes and refactors all existing loggers to use the schema generated from this proto.

### Proto Changes
The changes below have been added to keep the result log schema simple, partly owing to some inconveniences in the conversion of the proto types to a BigQuery compatible schema.
* Request/Response body has been changed from `google.protobuf.Struct` to `string`. The presence of the generic `Value` type in Structs results in a complex structure when generating the BQ schema from the proto. Additionally, the original implementation for the BQ logger used the `string` type for the request/response body and this change to the proto keeps it close to the schema from the original design.
* Request header has been changed from `map<string, google.protobuf.ListValue>` to `map<string, string>` since ListValue has the same inconvenience outlined with Struct when generating the BQ schema.

### Refactoring
In addition to the Kafka logger, all other existing loggers (Console, BQ and Fluentd) will also use the Struct generated from the Proto for logging, for consistency and code reusability. This eliminates the individual log entry types added for each logger. Instead, all of them will use `resultlog.TuringResultLogEntry` which is an alias for the message format defined in the proto. 

Depending on the format in which the message must be written to the log destination API (bytes / struct / ...), `TuringResultLogEntry.MarshalJSON()` or `TuringResultLogEntry.Value()` will be used and any further formatting changes required will be done by the individual loggers (for example, console logger requires an array of key-value pairs which will be obtained by post-processing the result of `TuringResultLogEntry.Value()`).

#### Behaviour Changes
The unification of the result logging schema causes the following changes in the observed end results:
* For the Console / Kafka loggers
    - the request/response bodies will now be represented as JSON formatted strings
    - the request header will be a map of strings
* For the BQ / Fluentd loggers, request `header` will be an array of records (with fields `key` and `value`)
* For the Console / BQ / Fluentd loggers field `event_timestamp` will replace `ts` to store the timestamp info
* The BQ table schema no longer recognizes `REQUIRED` fields - all fields are `NULLABLE`. (Proto3 eliminates the optional/required classifications for the fields. Although this can be achieved with special annotations for the BQ schema, we can ignore this for simplicity.)

#### Customer Impact

Deploying this PR will result in the following backwards-incompatible changes for the users of the result logging feature in Turing, when they attempt to deploy a new version of their existing routers. The differences and the recommended migration strategy are detailed below.

##### BigQuery Logging

Field | Change
------|--------
ts | Renamed as `event_timestamp` (like `ts`, `event_timestamp` will be used to partition the data by day)
request.header | Type change from `string` (JSON representation) to `repeated record` with fields `key` (header attribute name) and `value` (comma-delimited header attribute values)
(all) | All columns are NULLABLE (`turing_req_id`, `event_timestamp` and `request` were previously REQUIRED)

The BQ table schema may be found in `testdata/bq_turing_result_schema.json`.

**Recommended Migration strategy:** Start writing results to a new BQ destination table (will be created by the Turing app automatically, with the new schema) and (optionally) use a SQL query / script to migrate the data from the old table.

##### Kafka Protobuf

Field | Change
------|--------
request.header | Type change from `map<string, google.protobuf.ListValue>` to `map<string, string>`
request.body | Type change from `google.protobuf.Struct` to `string` (utf-8 encoded JSON body)
experiment.response, enricher.response, router.response, ensembler.response | Type change from `google.protobuf.Struct` to `string` (utf-8 encoded JSON response)

The new proto may be found in `TuringResultLog.proto`.

**Recommended Migration strategy:** Start writing results to a new Kafka topic.


engines/router/missionctl/log/resultlog/bigquery.go
```diff
@@ -14,103 +13,65 @@ import (
 	"context"
 	"encoding/json"
 	"strings"
-	"time"
 
 	"cloud.google.com/go/bigquery"
+	"go.einride.tech/protobuf-bigquery/encoding/protobq"
```
krithika369, line 18: This custom library has more functionality for schema inference from the Protobuf struct (for example, conversion of proto maps -> BQ records). See: https://github.com/googleapis/google-cloud-go/issues/1119

engines/router/missionctl/testdata/bq_turing_result_schema.json
```diff
@@ -1,87 +1,120 @@
 [
```
krithika369, line 1: Generated from the proto using https://github.com/GoogleCloudPlatform/protoc-gen-bq-schema

engines/router/missionctl/log/resultlog/fluentd.go
```diff
@@ -17,15 +17,11 @@ type fluentdClient interface {
 // fluentd backend.
 type FluentdLogger struct {
 	tag          string
-	bqLogger     BigQueryLogger
```
krithika369, line 20: Previously, this dependency existed so that the BQ logger (for streaming API) could be used to format the message in the same way for the batch logging with Fluentd. This is not important anymore since we're using the common `TuringResultLogEntry` methods.

krithika369:
@davidheryanto / @romanwozniak This PR assumes the proto added in our internal repo is satisfactory. I'm also still in the midst of testing these changes end to end (will remove the 'Draft' status when that's done) but the PR is largely ready for review. Thanks!

krithika369:
I've done some local / e2e testing of these changes, as detailed in the internal MR. Clearing the Draft status for this PR.

engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto
```diff
@@ -5,46 +5,42 @@ package turing;
...
 option go_package = "github.com/gojek/turing";
 
-import "google/protobuf/struct.proto";
 import "google/protobuf/timestamp.proto";
 
 message Request {
-    // The header from the incoming request to the Turing router
-    map<string, google.protobuf.ListValue> header = 1;
+    // The header from the incoming request to the Turing router, in JSON format
+    string header = 1;
```
davidheryanto, line 12: I think map<string,string> is more commonly used data type to represent HTTP headers.

Example from Google APIs:
https://github.com/googleapis/googleapis/blob/master/google/cloud/scheduler/v1/target.proto#L61
https://github.com/googleapis/googleapis/blob/master/google/actions/sdk/v2/webhook.proto#L56

Also since according to HTTP 1.1 spec https://tools.ietf.org/html/rfc2616#page-31
http message header consists of header field name and field value, so map<string,string> looks reasonable to represend this data. I think we should use JSON string representation when the format of the data can be arbitraty object.

krithika369: Hi David, I agree that `map<string,string>` is better for this. With that, the resulting header would be of the following BQ schema: 
```
{
    "name": "header",
    "type": "RECORD",
    "mode": "REPEATED",
    "description": "The header from the incoming request to the Turing router",
    "fields": [
     {
      "name": "key",
      "type": "STRING",
      "mode": "NULLABLE"
     },
     {
      "name": "value",
      "type": "STRING",
      "mode": "NULLABLE"
     }
    ]
   }
```

And this is the excerpt from the proto's Go library:
```
Header map[string]string `protobuf:"bytes,1,rep,name=header,proto3" json:"header,omitempty" protobuf_key:"bytes,1,opt,name=key,proto3" protobuf_val:"bytes,2,opt,name=value,proto3"`
```

They both match. But, the Go BQ library's dynamic schema inference on the `TuringResultLogMessage` struct fails ([ref](https://godoc.org/cloud.google.com/go/bigquery#StructSaver)), resulting in the below error:
```
repeated schema field header requires slice or array, but value has type map[string]string
```

I'm guessing it doesn't have extensive support for the protobuf annotations.

Our current implementation of the BQ logger uses `string` for representing the header, so I decided to keep it (not saying that this is correct or better). I don't see any major benefit of changing this to a map. However, if we feel strongly that this is the convention, I can handle it for BQ/Fluentd Loggers. Wdyt?

krithika369: The special handling would go [here](https://github.com/krithika369/turing/blob/update_turing_reslog_proto/engines/router/missionctl/log/resultlog/bigquery.go#L31), in the BQ's [ValueSaver](https://godoc.org/cloud.google.com/go/bigquery#ValueSaver) implementation. (I'm already using the ValueSaver method instead of StructSaver because the latter generates column names like `turingReqId` instead of `turing_req_id`.)

davidheryanto: Personally I think BigQuery specific issue should not affect the protobuf schema of our message. After all, BigQuery is only one of the many storage destinations. Some other storage destinations may have other requirements, so it's almost impossible to satisfy all the requirements unless we make all the types generic (like bytes or string). I think this responsibility should fall on the mapper/tooling to convert between Proto schema and BigQuery schema.

The benefit of having more specific type is that it is clearer from the consumer of the message the data format to expect. If we just use (json) string, without looking at the examples or documentation, these formats will be acceptable as headers
```json
{
  "Host": "http://example.com",
  "Content-Type": "json"
}
```
or 
```json
[
  {"key": "Host", "value": "http://example.com"},
  {"key": "Content-Type", "value": "json"}
]
```

But I guess if BigQuery schema compatibility (with certain tooling) is important in our case, we can use string type, but we need to provide more documentation regarding the JSON string representation of the HTTP headers.

krithika369: Ok, makes sense. I'll update the schema to `map<string, string>`.

davidheryanto: oh just wondering will `map<string, string>` cause issue with the schema generated from Beast?

krithika369: No, Beast supports this as expected (generated BQ schema will be `repeated` `record` with fields 'key' and 'value').

engines/router/missionctl/log/resultlog/console.go
```diff
@@ -15,16 +15,20 @@ func newConsoleLogger() *ConsoleLogger {
 // write logs the given TuringResultLogEntry to the console
 func (*ConsoleLogger) write(turLogEntry *TuringResultLogEntry) error {
 	// Get context-specific logger
-	logger := log.WithContext(*turLogEntry.ctx)
-	// Add request and responses
+	logger := log.Glob()
```
krithika369, line 18: context is no longer stored in the `TuringResultLogEntry`. The Turing Request ID that was being added with the contextual logger is now directly stored as a field in this struct.

engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto
```diff
@@ -5,46 +5,42 @@ package turing;
...
 
 message Response {
-    // The response from a Turing component (Enricher / Experiment Engine / Router / Ensembler).
-    google.protobuf.Struct response = 1;
-    
+    // The JSON response from a Turing component (Enricher / Experiment Engine / Router / Ensembler), UTF-8-encoded.
+    string response = 1;
+
     // The error from a Turing component, when a successful response is not received.
     string error = 2;
```
davidheryanto, line 23: Is this error always non-empty when the respose is not ok? i.e if both response and error is empty string, means it is a successful response with no error rather than erronous response with no error detail.

davidheryanto (approved):
LGTM

We just need to document the migration process when the changes are deployed for existing users. 

For example how to use this updated proto schema in BigQuery when they have existing table, knowing that our proto changes cannot reuse the existing BigQuery table because it involves changing existing column data type (which BigQuery does not support)

engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto
krithika369: Yes, this error will always be non-empty when the response is empty or not 2XX ([router](https://github.com/gojek/turing/blob/main/engines/router/missionctl/mission_control.go#L229), [enrich/ensemble](https://github.com/gojek/turing/blob/main/engines/router/missionctl/mission_control.go#L154)) since Turing wraps the error received with more info / creates a new error object.

krithika369:
Thanks @davidheryanto. Added some info on the customer impact to the PR description. Merging.

PR #45, opened 2021-01-08
Version diff view

romanwozniak:
This PR updates Turing UI and adds a versions comparison view to make it simpler to see the differences between two arbitrary router versions.

### Demo:
https://user-images.githubusercontent.com/1886194/104061657-ad3f8c00-5201-11eb-8919-3bbaf2bd15ec.mp4

### Details:
This PR touches two parts:
- `History` tab at the router details page:
   - The relative proportions of versions list and activity log views are changed from 1:1 to 4:3
   - Actions column in versions list table is updated with a custom logic of rows selection (for comparison) and controls for navigating to the comparison view
- `VersionComparisonView` was added:
   - It leverages on `react-diff-viewer` component to show line-by-line diff in the text representation of version configurations
   - New `RouterVersion` model class was added to incapsulate some logic of prettifying and cleaning raw version config data


ui/src/router/versions/components/RouterVersionActions.js
```diff
@@ -29,9 +29,10 @@ export const RouterVersionActions = ({
         label: "Deploy",
         name: "Deploy this version",
         icon: "importAction",
-        available: config =>
-          !(isActiveConfig(config) && routerStatus !== Status.UNDEPLOYED),
-        enabled: () => routerStatus !== Status.PENDING,
+        available: () => true,
```
krithika369, line 32: This logic would mean that if all versions of a router deployment have failed, the "deploy" action is not activated for the current config. It's a minor thing but I'm wondering if the check should be `routerStatus === Status.DEPLOYED` instead of the `routerStatus !== Status.UNDEPLOYED`.

Screenshot 2021-01-11 at 11 16 28 PM

ui/src/services/version/RouterVersion.js
```diff
@@ -0,0 +1,92 @@
...
+        resource_request: this.resource_request
+      },
+      experiment_engine: {
+        type: this.experiment_engine.type,
+        ...(this.experiment_engine.type !== "nop"
+          ? {
+              config: {
+                client: {
+                  id: this.experiment_engine.config.client.username,
+                  passkey: "[hidden]"
```
krithika369, line 30: If the passkey changed between versions, do you think it would be of interest to the users? If the passkey is set, we can may be compute another hash over the already encrypted string and use it for compare?

krithika369 (commented):
Looks cool! Added some small questions, will take a closer look in the morning.

ui/src/components/page_navigation/PageNavigation.js
```diff
@@ -21,7 +26,7 @@ export const PageNavigation = ({ tabs, actions, selectedTab, ...props }) => (
             {...(tab.href
               ? { href: tab.href, target: "_blank" }
               : { onClick: () => props.navigate(`./${tab.id}`) })}
-            isSelected={tab.id === selectedTab}
+            isSelected={selectedTab.startsWith(tab.id)}
```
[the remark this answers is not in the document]
romanwozniak: In order to make tabs active for nested paths as well. I.e. `History` should be an active tab for `/routers/:id/history` route, but also for `/routers/:id/history/compare/:versionA/:versionB`

ui/src/router/versions/components/RouterVersionActions.js
romanwozniak: Agree

ui/src/services/version/RouterVersion.js
romanwozniak: I was thinking about this, but I believe that this hashed string makes no sense to user at all. 
Instead, there should something simpler, such as just `[hidden]/[changed]` or similar.

> may be compute another hash over the already encrypted string
curious, why do we need to compute another hash?

krithika369: >why do we need to compute another hash?

I thought you had concerns with displaying the encrypted string as is in case it was long or insecure (not that the user can't get hold of this string otherwise).

Then, should we just call the key "encrypted_passkey" instead of "passkey" and put down the actual contents so it can still be compared?

ui/src/router/versions/components/RouterVersionActions.js
krithika369: @romanwozniak The change should be `routerStatus === Status.DEPLOYED`, not `!==`

ui/src/services/version/RouterVersion.js
romanwozniak: Okay, done

ui/src/router/versions/list/ListRouterVersionsTable.js
```diff
@@ -22,15 +21,70 @@ export const ListRouterVersionsTable = ({
...
+        return { ...items };
+      });
+    },
+    [setSelection]
+  );
+
+  const onShowDiff = useCallback(() => {
+    const selected = Object.values(selection);
+    if (selected.length === 2) {
+      const [leftSelection, rightSelection] = selected;
```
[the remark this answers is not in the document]
romanwozniak: I thought that it's fine to have it this way for simplicity, but since you brought it up, I updated the code to show versions in the order they were selected in the table

PR #46, opened 2021-01-11
Add data migration script for kafka log config

davidheryanto:
[Recent update](https://github.com/gojek/turing/pull/31) in Turing router add support for multiple Kafka serialization format: json or protobuf
This requires an update to the existing `log_config` data in router version table.  `kafka_config` needs to have `serialization_format` field. This migration ensures that

### DIFF
--- a/api/db-migrations/000007_log_config_kafka_serialization_format.down.sql
+++ b/api/db-migrations/000007_log_config_kafka_serialization_format.down.sql
@@ -1,14 +1,5 @@
 -- Remove 'serialization_format' field in 'kafka_config'
 
 UPDATE router_versions
-SET log_config = (SELECT json_build_object(
-    'log_level', log_config -> 'log_level',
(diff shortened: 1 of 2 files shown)


romanwozniak (approved):
LGTM

api/db-migrations/000007_log_config_kafka_serialization_format.down.sql
```diff
@@ -0,0 +1,14 @@
+-- Remove 'serialization_format' field in 'kafka_config'
+
+UPDATE router_versions
```
krithika369, line 3: Perhaps we can use `jsonb_set()` to simplify this, so we don't have to explicitly specify the fields that wont change: https://www.postgresql.org/docs/11/functions-json.html

davidheryanto: Updated

krithika369 (approved):
LGTM!

PR #47, opened 2021-01-18
Add optional `annotations` field into `Route` to have better integration between Turing and gojek/merlin

romanwozniak:
### Context:
When users configure routers in Turing, they can either input any arbitrary URL or select one of Merlin model endpoints (if Merlin integration is enabled) as the router's endpoint. 

When Merlin endpoint is selected as one of the router endpoints, we want to make it easier for users to navigate to the model details page in Merlin console.

### Changes:
- API:
  - Added optional `Annotations` field into `models/Route` entity to store a link to the model details page in Merlin
- UI:
  - Updated `MerlinEndpointsProvider` to show only `serving` endpoints 
  - Updated `MerlinEndpointsProvider` to also include an annotation with the id of the corresponding model
  - Updated `RoutesConfigTable` to show the link to Merlin model details page if `annotations` contain model id

### UI:
Screenshot 2021-01-18 at 20 46 49


api/turing/models/routes.go
```diff
@@ -11,12 +11,14 @@ import (
...
 type Route struct {
-	// Id of the route
+	// ID of the route
 	ID string `json:"id"`
 	// Type of the route
 	Type string `json:"type"`
 	// Endpoint to query
 	Endpoint string `json:"endpoint"`
+	// EndpointInfoURL (optional) holds URL to endpoint details page
+	EndpointInfoURL string `json:"endpoint_info_url"`
```
krithika369, line 21: I find the idea of an `EndpointInfoURL` kind of ad-hoc. Also, in the rare event that the Merlin app server is moved, all the DB records need to be updated.

Would it be better to create a `MerlinEndpoint` type that can be `nil` where N/A, and can otherwise store the model id? Then, for display, the UI can format the endpoint URL by calling the relevant Merlin APIs if needed. Wdyt?

davidheryanto: Another approach we can do is having a field like `annotations` with type `map[string]string` to add this addtional information to a Router object. 

The annotation approach gives more flexibility to add/remove the extra information in the future and makes it less coupled to other objects types. If we do not need to enforce having this extra field or having a relationship on that field, we can consider this approach.

romanwozniak: > I find the idea of an `EndpointInfoURL` kind of ad-hoc. Also, in the rare event that the Merlin app server is moved, all the DB records need to be updated.
> 
> Would it be better to create a `MerlinEndpoint` type that can be `nil` where N/A, and can otherwise store the model id? Then, for display, the UI can format the endpoint URL by calling the relevant Merlin APIs if needed. Wdyt?

I didn't want to introduce another point of coupling between Turing and Merlin, that's why decided to go with a generic `EndpointInfoURL`. But yes, maybe `Annotations`, suggested by David, would be a better idea.

romanwozniak: @davidheryanto @krithika369 I addressed your comments, please give it another look

davidheryanto: Ok looks good. Is there gonna be another PR to update the databasa schema and test the annotation field gets written to the database. Currently I believe this new annotation field will not be written to the database.

krithika369 (approved):
LGTM

api/turing/models/routes.go
romanwozniak: `annotations` field, as well as any other fields of `Route` struct are being written into the database as json blob. The test for (de-)serialization of entire `Route` struct into json was updated correspondingly.

PR #49, opened 2021-01-26
[BugFix] Fix version comparison of docker ensemblers

romanwozniak:
The comparison of docker ensemblers in Version Deff View was working incorrectly, because the wrong schema of `ensembler` JSON object was used in `toPrettyYaml` method. This is fixed now

Screenshot 2021-01-26 at 10 13 35


krithika369 (approved):

krithika369:
LGTM

PR #50, opened 2021-01-26
[BugFix] fix monitoring link when no specific revision is selected

romanwozniak:
The placeholder for "any" revision in the monitoring dashboard should be `$__all`, instead of `$_all`


krithika369 (approved):
LGTM

PR #54, opened 2021-04-28
Batch Ensembling API: Creation of batch ensembling jobs.

ashwinath:
As part of @romanwozniak PR: https://github.com/gojek/turing/pull/52, we are building a ensembler with a user defined model.

The user flow is as follows:
1. User creates a MLFlow model. (Not part of this PR)
2. User submits the ensembler via POST `/projects/{project_id}/ensemblers` endpoint. (Not part of this PR)
3. User creates a batch ensembling job that will be executed in the future. (This PR)

This PR includes:
1. Creation of ensembling job via POST `/projects/{project_id}/jobs`.
2. OpenAPI validation of multiple Swagger YAML files.
3. Improve developer experience; DB migrations are now symlinked in `turing/infra/docker-compose/dev/turing/db-migrations/`

Not in scope of this PR:
1. Building the OCI image
2. The orchestration of the Ensembling job.


romanwozniak (commented):
@ashwinath thanks for the MR, I've started reviewing it, but haven't managed to complete it and will do it tomorrow. But you can start answering the questions I asked

api/db-migrations/000009_ensembling_job.up.sql
```diff
@@ -0,0 +1,26 @@
...
+    'terminated',
+    'failed_submission',
+    'failed_building'
+);
+
+CREATE TABLE IF NOT EXISTS ensembling_jobs
+(
+    id               serial PRIMARY KEY,
+    name             varchar(50) NOT NULL,
+    version_id       integer,
```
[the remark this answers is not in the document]
ashwinath: Let me remove this, I was confused between ensembler and ensembling job at the start and didn't remove it in the end. I was thinking if we could use multiple versions of the same ensembler, not ensembling job.

api/swagger-batch.yaml
```diff
@@ -119,27 +117,42 @@ paths:
...
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+      requestBody:
+        description: A JSON object containing information about the ensembler.
+        required: true
+        content:
+          application/json:
+            schema:
+              oneOf:
```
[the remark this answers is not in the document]
ashwinath: It could be some other job in the future but I'll remove this because it's only one kind of job for now

api/swagger-batch.yaml
```diff
@@ -173,13 +187,55 @@ components:
             artifact_uri:
               type: "string"
               format: "uri"
+
     EnsemblingJob:
       description: A JSON object that represents an ensembling job for batch experiment use cases
       properties:
         id:
           $ref: "#/components/schemas/Id"
+        name:
```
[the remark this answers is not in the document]
ashwinath: This should be user supplied, it's easier for them to keep track if they can provide some name to it.

api/swagger-batch.yaml
```diff
@@ -197,12 +253,15 @@ components:
...
-          items:
+          type: "object"
+          additionalProperties:
             $ref: "#/components/schemas/EnsemblingJobPredictionSource"
         ensembler:
           $ref: "#/components/schemas/EnsemblingJobEnsemblerSpec"
+        sink:
+          $ref: "#/components/schemas/EnsemblingJobSink"
 
+    # roman
```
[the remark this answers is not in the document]
ashwinath: Woops! I was quite confused when I wrote this part so I put some comment on who wrote which part. Somehow I named some variables exactly the same as you did when I was writing this part.

api/swagger-batch.yaml
```diff
@@ -215,26 +274,27 @@ components:
 
     Dataset:
       type: "object"
+      oneOf:
+        - $ref: "#/components/schemas/BigQueryDataset"
```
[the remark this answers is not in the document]
ashwinath: Without that `oneOf` it does not show the example properly.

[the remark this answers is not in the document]
ashwinath: There are no tags in this site but look for `Mapping Type Names` in https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

The example still uses an `oneOf` here.

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,75 @@
...
+	var errResp *Response
+	var project *mlp.Project
+	if project, errResp = c.getProjectFromRequestVars(vars); errResp != nil {
+		return errResp
+	}
+
+	request, ok := body.(*models.EnsemblingJob)
+	if !ok {
+		return BadRequest("Unable to parse body as ensembling job", "could not unmarshal request body")
+	}
```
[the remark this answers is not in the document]
ashwinath: yeah actually this code will not be executed since there is a check prior to this

api/swagger-batch.yaml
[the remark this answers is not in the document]
romanwozniak: The example you've posted doesn't contain `bq_config` property, and that's my point

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,75 @@
...
+		request.EnsemblerID,
+		service.EnsemblersFindByIDOptions{
+			ProjectID: &request.ProjectID,
+		},
+	)
+	if err != nil {
+		return NotFound("ensembler not found", err.Error())
+	}
+
+	_, err = c.MLPService.GetEnvironment(request.EnvironmentName)
```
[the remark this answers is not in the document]
ashwinath: I'm not too sure what you mean by this. How would we determine the environment if that is the case? Could you point me to the configuration at the backend?

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return BadRequest(
+			"invalid environment",
+			fmt.Sprintf("environment %s does not exist", request.EnvironmentName),
+		)
+	}
+
+	// Save ensembling job
+	err = c.EnsemblingJobService.Save(request)
```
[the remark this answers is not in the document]
ashwinath: During the building of the OCI image, we can query the ensemblers table to get the artifacts URI. There is no need to duplicate the information and store it in the ensembling_jobs table since it is linked by ensembler_id foreign key.

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
romanwozniak: https://github.com/gojek/merlin/blob/ec8a8e2b0c3a8e6345c1e8cfeef63d5d03011e88/api/api/prediction_job_api.go#L53

api/turing/api/ensembling_job_api_it_test.go
```diff
@@ -0,0 +1,330 @@
...
+	merlin "github.com/gojek/merlin/client"
+	mlp "github.com/gojek/mlp/client"
+	"github.com/gojek/turing/api/turing/models"
+	"github.com/gojek/turing/api/turing/service"
+	"github.com/gojek/turing/api/turing/service/mocks"
+	"github.com/stretchr/testify/mock"
+	"gotest.tools/assert"
+)
+
+func generateEnsemblingJobFixtureJSON() string {
```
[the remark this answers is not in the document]
ashwinath: Agreed on `$.infra_config.image_ref` will be removing this as this is only needed to build the OCI image and submit the job to Kubernetes. Will be fixing `$.ensembler_config.spec.ensembler.uri` to the local directory, this should be a fixed location because in the Dockerfile I have copied the artifact into the local file directory.

api/turing/api/ensembling_job_api_it_test.go
```diff
@@ -0,0 +1,330 @@
...
+		path                 string
+		expected             *Response
+		ensemblersService    func() service.EnsemblersService
+		ensemblingJobService func() service.EnsemblingJobService
+		mlpService           func() service.MLPService
+		vars                 RequestVars
+		body                 string
+	}{
+		{
+			name:     "nominal flow",
```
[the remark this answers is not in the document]
ashwinath: Ah I didn't notice this, I just relied on my text editor to generate the boilerplate code. Will change my boilerplate code as well.

api/swagger-batch.yaml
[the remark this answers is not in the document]
ashwinath: I'm confused, the example I posted was with your suggestion. Is it supposed to be this way?

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
ashwinath: Alright, if thats the case I'll just get the default from a configuration file rather than the database.

api/db-migrations/000009_ensembling_job.up.sql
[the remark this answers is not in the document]
romanwozniak: There is no concept of an ensembler version (at least for now), so yes, this is redundant.

api/swagger-batch.yaml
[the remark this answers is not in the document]
romanwozniak: Can we keep this optional? Meaning, if the job name is not provided, then the job name is generated, based on the ensembler name + unique identifier (maybe ts?)

api/swagger-batch.yaml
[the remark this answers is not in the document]
romanwozniak: Oh, I got it wrong. Can you post how does example look like without my suggestion? :)

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,94 @@
...
+		},
+	)
+	if err != nil {
+		return NotFound("ensembler not found", err.Error())
+	}
+
+	// Ensembler URI will be a local directory
+	// Dockerfile will build copy the artifact into the local directory.
+	// See engines/batch-ensembler/app.Dockerfile
+	artifactFolderName, err := getEnsemblerFolderName(ensembler)
```
romanwozniak, line 50: This should be done at the service layer, not controller layer. Controller should only be responsible for (de)serialization and validation of the user provided data

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
romanwozniak: You have already ensembler object here, fetched from the database. Why would you need to fetch it again if the `create job` action is effectively `submit job` and the docker image is being built at the same time when the job is saved?

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		return BadRequest("ensembler not supported", err.Error())
+	}
+	request.EnsemblerConfig.EnsemblerConfig.Spec.Ensembler.Uri = fmt.Sprintf("/home/spark/%s", artifactFolderName)
```
romanwozniak, line 54: `"/home/spark/` should be moved into a named variable

test/e2e/setup-infra.sh
```diff
@@ -188,7 +188,7 @@ function install_turing() {
     cd "$script_dir/../.."
 
     kubectl apply -f ./test/e2e/turing.mockserver.yaml
-    helm install turing ./infra/chart -f ./test/e2e/turing.helm-values.yaml --wait
+    helm upgrade --install turing ./infra/chart -f ./test/e2e/turing.helm-values.yaml --wait
```
romanwozniak, line 191: What is the rational?

infra/docker-compose/dev/turing/db-migrations
```diff
@@ -0,0 +1 @@
+../../../../api/db-migrations/
```
romanwozniak, line 1: I vaguely remember that there was an issue with using symlinks with docker. Have you verified that this works as expected?

ashwinath: What was the issue? I just did `docker-compose up -d` and it works.

test/e2e/setup-infra.sh
ashwinath: I'll remove this, I remember now. There was one run where the helm release couldn't install because it had the same release name. In retrospect I think it was because i called it twice. Will revert this

api/turing/service/ensembling_job_service_test.go
```diff
@@ -0,0 +1,201 @@
...
+			assert.Equal(t, found.EnvironmentName, ensemblingJob.EnvironmentName)
+			assert.Equal(t, models.State("pending"), ensemblingJob.Status)
+			assert.Equal(t, found.InfraConfig, ensemblingJob.InfraConfig)
+			assert.Equal(t, found.EnsemblerConfig, ensemblingJob.EnsemblerConfig)
+		})
+	})
+}
+
+func TestFindPendingJobsAndUpdateIntegration(t *testing.T) {
+	t.Run("insertion with no errors", func(t *testing.T) {
```
romanwozniak, line 164: Test name should be updated accordingly

api/turing/service/ensembling_job_service.go
```diff
@@ -0,0 +1,70 @@
...
+	result := query.First(&ensemblingJob)
+
+	if err := result.Error; err != nil {
+		return &ensemblingJob, err
+	}
+
+	return &ensemblingJob, nil
+}
+
+func (s *ensemblingJobService) FindPendingJobs(limit int) ([]*models.EnsemblingJob, error) {
```
romanwozniak, line 55: Can you follow the convention I defined in `EnsemblerSerivce`? `FindPendingJobs` should be replaced with `List`, that accepts `options` and `options are` a composition of `PaginationOptions` and `Status` field (so you can fetch not just pending jobs, but jobs of any status)

api/turing/service/ensembling_job_service_test.go
```diff
@@ -0,0 +1,201 @@
...
+			assert.NoError(t, err)
+			assert.NotEqual(t, models.ID(0), ensemblingJob.ID)
+
+			// Query pending ensembling jobs
+			ensemblingJobs, err := ensemblingJobService.FindPendingJobs(10)
+			assert.NoError(t, err)
+			assert.Equal(t, 1, len(ensemblingJobs))
+
+			queriedEnsemblingJob := ensemblingJobs[0]
+			assert.Equal(t, models.State("pending"), queriedEnsemblingJob.Status)
```
romanwozniak, line 182: `models.State("pending")` as well as other job statuses should be replaced with a named constants

api/turing/service/ensembling_job_service.go
```diff
@@ -0,0 +1,70 @@
...
+	query := s.db.Order("id").Limit(limit).Where("status = ?", models.JobPending)
+	var ensemblingJobs []*models.EnsemblingJob
+	err := query.Find(&ensemblingJobs).Error
+	return ensemblingJobs, err
+}
+
+func (s *ensemblingJobService) UpdateJobStatus(id models.ID, status models.State, errString string) error {
```
romanwozniak, line 62: `status models.State` – so is it `state` or `status`?

romanwozniak, line 62: Can `UpdateJobStatus` be replaced just with `Save` method? \

api/turing/service/ensembling_job_service_test.go
romanwozniak: Actually, it's already defined as `models.JobPending`

api/turing/middleware/openapi_validation_test.go
```diff
@@ -157,18 +161,21 @@ func TestOpenAPIValidationValidate(t *testing.T) {
 
 func TestNewOpenAPIV2Validation(t *testing.T) {
```
romanwozniak, line 162: Func name should be updated accordingly

api/turing/middleware/openapi_validation.go
```diff
@@ -61,19 +81,27 @@ func NewOpenAPIV2Validation(swaggerYamlFile string, options OpenAPIValidationOpt
...
+		route, pathParams, _ := router.FindRoute(r.Method, r.URL)
+		input := &openapi3filter.RequestValidationInput{
+			Request:    r,
+			PathParams: pathParams,
+			Route:      route,
+			Options:    openapi.options,
+		}
+		err := openapi3filter.ValidateRequest(context.Background(), input)
+		if err != errors.New("invalid route") || err == nil {
+			// Make sure that it is the correct error message, ignore invalid route error unless that is the only error
```
romanwozniak, line 100: Is it possible to skip validation if `router.FindRoute` emits nil (can't find the route in the given spec)? So in this case you don't need to check if error is `invalid route`

api/turing/middleware/openapi_validation.go
```diff
@@ -46,10 +50,26 @@ func NewOpenAPIV2Validation(swaggerYamlFile string, options OpenAPIValidationOpt
 	if err != nil {
 		return nil, err
 	}
+	var routers []*openapi3filter.Router
+	routers = append(
+		routers,
+		openapi3filter.NewRouter().WithSwagger(v3Swagger),
+	)
```
romanwozniak, line 57: `routers` can be defined and initialised in a single line
```go
routers := []*openapi3filter.Router{openapi3filter.NewRouter().WithSwagger(v3Swagger)}
```

api/turing/api/ensembling_job_api_test.go
```diff
@@ -0,0 +1,339 @@
...
+}
+
+func createPyFuncEnsembler(id int) models.EnsemblerLike {
+	return &models.PyFuncEnsembler{
+		GenericEnsembler: &models.GenericEnsembler{
+			Model:     models.Model{ID: 1},
+			Type:      models.EnsemblerTypePyFunc,
+			ProjectID: 1,
+		},
+		ArtifactURI: "gs://d-gods-mlp/ensembler",
```
romanwozniak, line 160: Can you rename `d-gods-mlp` into something generic (i.e. backet-name)? Remember that that's OSS project

romanwozniak (changes requested):
@ashwinath I'm done with the review, please check if the comments make sense and try to address them. Thanks!

infra/docker-compose/dev/turing/db-migrations
romanwozniak: Okay 👌

api/swagger-batch.yaml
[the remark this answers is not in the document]
ashwinath: Screenshot 2021-05-17 at 1 10 00 PM

api/swagger-batch.yaml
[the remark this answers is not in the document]
ashwinath: Alright, I think that's fair too.

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
ashwinath: Alright will resave the data into ensembling job.

api/turing/api/ensembling_job_api.go
ashwinath: Ok, I was thinking the service was more of a facade pattern to do all connections to the database. Will move all logic into service instead.

api/turing/service/ensembling_job_service_test.go
ashwinath: The rationale here is that we don't want to reference the same constant that was being sent to the database. What if `models.JobPending` = `jobnotpending` instead. This test is supposed to catch this small (no reprecussion) bug as well.

api/turing/service/ensembling_job_service.go
ashwinath: I'll rename all to state

api/turing/middleware/openapi_validation.go
ashwinath: Because this is in two routers, other validation error messages should be taken priority.

Scenario one, GET `/api/router-1-route`:
router 1: `some other error`
router 2: `invalid route`

This scenario will fail without the because `err != errors.New("invalid route")` it will reply `invalid route` instead of some other error.

Scenario two, GET `/api/router-2-route`:
router 1: `invalid route`
router 2: `some other error`

This scenario will pass because router 2 (end of the loop) is the last error and `some other error` will appear.

Invalid route will be the catch all situation. The best we can do to cut short the loop is when it returns another error / no error during the loop. Else we have to test all routers.

That said, line 98 should be an assignment operator rather than a short variable declaration operator. Will change this.

api/swagger-batch.yaml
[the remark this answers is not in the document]
romanwozniak: okay, this looks good

api/turing/service/ensembling_job_service_test.go
romanwozniak: Does it matter what's the value of `models.JobPending` if in all the codebase you use the same constant to check the state of the job? `models.JobPending` can be stored as `123` or `😎😇🤗`, it doesn't matter as long as you stick to the convention, that `models.JobPending` represents jobs in pending state

api/turing/api/ensembling_job_api.go
romanwozniak: No, it's not. There are other services, that do more than interacting with DB, for instance: [outer_deployment_service.go](https://github.com/gojek/turing/blob/main/api/turing/service/router_deployment_service.go) or [crypto_service.go](https://github.com/gojek/turing/blob/main/api/turing/service/crypto_service.go)

api/turing/service/ensembling_job_service.go
ashwinath: Well state is a reserved keyword so I'll change all to status.

api/turing/middleware/openapi_validation.go
romanwozniak: I'd prefer more explicit way of detecting if the endpoint is not described in any of specs. Something like this:

```go
            for idx, router := range openapi.routers {
		route, pathParams, _ := router.FindRoute(r.Method, r.URL)
                if route == nil && idx == len(openapi.routers) - 1 {
                     // endpoint is not described
                     return fmt.Errorf("Route `%s %s` is not described in openapi spec",  r.Method, r.URL)
                }
		input := &openapi3filter.RequestValidationInput{
			Request:    r,
			PathParams: pathParams,
			Route:      route,
			Options:    openapi.options,
		}
		if err := openapi3filter.ValidateRequest(context.Background(), input); err != nil {
                    return err
                }
           }
       return nil
```

api/turing/service/ensembling_job_service_test.go
ashwinath: Yeah I agree with your point but let's say there was a problem in production, we take a peep into the database to see whats going on and we find that the status is `123`, it's one extra mental hoop to go through the codebase to find out what `123` means. Nonetheless, I don't have a strong enough opinion so I'll change this.

api/turing/api/ensembling_job_api.go
ashwinath: Yah I noticed this as well.

api/turing/middleware/openapi_validation.go
ashwinath: There's a small error there because once there is a valid route, you should always return the error since we know now that that error is valid.
```go
if err := openapi3filter.ValidateRequest(context.Background(), input); err != nil {
		return err
}
```

Anyway here's one that is working that should address your concerns.
```go
// Validate the request against the OpenAPI spec
func (openapi *OpenAPIValidation) Validate(r *http.Request) error {
	for idx, router := range openapi.routers {
		route, pathParams, _ := router.FindRoute(r.Method, r.URL)
		if route == nil && idx == len(openapi.routers)-1 {
			// endpoint is not described
			return fmt.Errorf("Route `%s %s` is not described in openapi spec", r.Method, r.URL)
		}
		input := &openapi3filter.RequestValidationInput{
			Request:    r,
			PathParams: pathParams,
			Route:      route,
			Options:    openapi.options,
		}
		return openapi3filter.ValidateRequest(context.Background(), input)
	}
	// Will not reach this line.
	return nil
}
```

romanwozniak: yup, lgtm

api/turing/service/ensembling_job_service.go
ashwinath: Yes I have addressed this.

api/turing/middleware/openapi_validation.go
ashwinath: Done

api/turing/api/ensembling_job_api_test.go
ashwinath: Yup renamed all gojek specific stuff out.

api/turing/middleware/openapi_validation_test.go
ashwinath: Updated

ashwinath:
Will fix the tests tomorrow.

api/turing/service/ensembling_job_service.go
ashwinath: done

api/turing/service/ensembling_job_service_test.go
ashwinath: done

ashwinath: done

api/turing/api/ensembling_job_api.go
ashwinath: done

api/turing/api/ensembling_job_api_it_test.go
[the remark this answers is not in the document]
ashwinath: done

api/turing/service/ensembling_job_service_test.go
ashwinath: done

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,85 @@
...
+	ensemblerDirectory, err := c.EnsemblingJobService.GetEnsemblerDirectory(ensembler)
+	if err != nil {
+		return BadRequest("ensembler not supported", err.Error())
+	}
+	request.EnsemblerConfig.EnsemblerConfig.Spec.Ensembler.Uri = ensemblerDirectory
+
+	// Populate ensembler artifact URI, error can be ignored since the type check is done prior
+	artifactURI, _ := c.EnsemblingJobService.GetArtifactURI(ensembler)
+	request.InfraConfig.ArtifactURI = artifactURI
+	request.InfraConfig.EnsemblerName = ensembler.GetName()
```
romanwozniak, line 64: All this logic should be done inside the `EnsemblingJobService`, not at the controller level. You just need to pass an instance of `ensembler` into the relevant method of that service and populate all the necessary fields from it inside of that method

romanwozniak: Ref: https://github.com/gojek/merlin/blob/8f52ce14c73de3a3efcafb9a9bca40138ee38c89/api/api/prediction_job_api.go#L58

api/turing/api/ensembling_job_api_it_test.go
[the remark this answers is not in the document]
romanwozniak: The other comment about the use of `ensembler_config` is not addressed or commented. Can it be `job_config` or something?

api/turing/service/pagination.go
```diff
@@ -29,3 +33,22 @@ func PaginationScope(options PaginationOptions) func(db *gorm.DB) *gorm.DB {
 		return db
 	}
 }
+
+// CreatePaginatedResults is a helper function that helps to create paginated results
+func CreatePaginatedResults(options PaginationOptions, count int, results interface{}) *PaginatedResults {
```
romanwozniak, line 38: Good idea 👍. But this probably can be package-private, don't expect it to be used outside of `service` package

api/turing/service/ensembling_job_service.go
```diff
@@ -0,0 +1,148 @@
...
+// EnsemblingJobFindByIDOptions contains the options allowed when finding ensembling jobs.
+type EnsemblingJobFindByIDOptions struct {
+	ProjectID *models.ID
+}
+
+// EnsemblingJobListOptions holds query parameters for EnsemblersService.List method.
+type EnsemblingJobListOptions struct {
+	PaginationOptions
+	ProjectID *models.ID
+	Status    models.Status
```
romanwozniak, line 25: Should this be `*models.Status` as well? You can also add validation tags, such as `validate:"oneof=pending running ..."`

api/turing/service/ensembling_job_service.go
```diff
@@ -0,0 +1,148 @@
...
+
+// EnsemblingJobService is the data access object for the EnsemblingJob from the db.
+type EnsemblingJobService interface {
+	Save(ensembler *models.EnsemblingJob) error
+	FindByID(id models.ID, options EnsemblingJobFindByIDOptions) (*models.EnsemblingJob, error)
+	List(options EnsemblingJobListOptions) (*PaginatedResults, error)
+	GetDefaultEnvironment() string
+	GenerateDefaultJobName(ensemblerName string) string
+	GetEnsemblerDirectory(ensembler models.EnsemblerLike) (string, error)
+	GetArtifactURI(ensembler models.EnsemblerLike) (string, error)
```
romanwozniak, line 36: I'm not onboard with the structure of this service. 
1. `GenerateDefaultJobName` `GetEnsemblerDirectory ` and `GetArtifactURI ` are probably only required in the context of ensembler creation. Hence, you don't need them to be defined at the service level, but instead, you would probably need `CreatePredictionJob` method that incapsulates all this logic.
2. Ensembling job can only work with pyfunc ensemblers, so it doesn't make sense to define the signature of these methods as `(ensembler models.EnsemblerLike)`, but instead it should accept `ensembler models.PyfuncEnsembler` and type check (and casting) should be done at the controller level (and relevant `BadRequest` response should be sent if user is trying to submit a job with non-pyfunc ensembler)

romanwozniak (changes requested):
Some of my previous questions were not addressed or answered, please check them + few more from this round of review. thanks!

api/turing/service/ensembling_job_service.go
ashwinath: Agreed, It's starting to come to me now how this repo is structured. Have implemented the suggestions.

api/turing/api/ensembling_job_api.go
ashwinath: done

api/turing/service/pagination.go
ashwinath: done

api/turing/service/ensembling_job_service.go
ashwinath: Not possible with the linter throwing errors saying that the line is too long.
```
turing/service/ensembling_job_service.go:25: line is 134 characters (lll)
        Status    *models.Status `validate:"oneof=pending running terminating terminated completed failed failed_submission failed_building"`
```

Will omit this suggestion.

api/turing/api/ensembling_job_api_it_test.go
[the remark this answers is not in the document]
ashwinath: Done, have changed this to job_config.

ashwinath:
Hold on just give me a minute, I just realised that some of the comments are not reflected in this page.

api/swagger-batch.yaml
```diff
@@ -308,3 +408,40 @@ components:
...
+          type: string
+        resources:
+          $ref: "#/components/schemas/EnsemblingResources"
+
+    EnsemblingResources:
+      type: "object"
+      properties:
+        requests:
+          $ref: "#/components/schemas/EnsemblingResource"
+        limits:
```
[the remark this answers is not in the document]
ashwinath: Do we want to enforce requests == limits though? Seems rather opinionated and Kubernetes does not limit this as well.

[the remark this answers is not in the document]
ashwinath: Okay on second thoughts we do want to enforce this because we want to control the infrastructure usage.

api/swagger-batch.yaml
```diff
@@ -308,3 +408,40 @@ components:
...
+          $ref: "#/components/schemas/EnsemblingResource"
+
+    EnsemblingResource:
+      type: "object"
+      properties:
+        cpu:
+          type: string
+        memory:
+          type: string
```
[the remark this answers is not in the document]
ashwinath: ok, will be adding in the suggestions.

api/turing/api/appcontext.go
```diff
@@ -102,11 +105,17 @@ func NewAppContext(
...
-	appContext.OpenAPIValidation, err = middleware.NewOpenAPIV2Validation(
-		cfg.SwaggerFile,
+	for _, swaggerFile := range cfg.SwaggerV3Files {
+		if _, err = os.Stat(swaggerFile); os.IsExist(err) {
+			return nil, errors.Wrapf(err, fmt.Sprintf("Swagger v3 spec file not found %s", swaggerFile))
+		}
+	}
+	appContext.OpenAPIValidation, err = middleware.NewOpenAPIValidation(
+		cfg.SwaggerV2File,
+		cfg.SwaggerV3Files,
```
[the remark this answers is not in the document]
ashwinath: I changed it into a struct instead so it's clearer to iterate. I personally prefer having fixed key names and just taking values from them.

api/swagger-batch.yaml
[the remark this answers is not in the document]
ashwinath: done

api/swagger-batch.yaml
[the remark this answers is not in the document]
ashwinath: done

api/turing/api/appcontext.go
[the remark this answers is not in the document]
ashwinath: done

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,71 @@
...
+	switch v := ensembler.(type) {
+	case *models.PyFuncEnsembler:
+		pyFuncEnsembler = ensembler.(*models.PyFuncEnsembler)
+	default:
+		return BadRequest("only pyfunc ensemblers allowed", fmt.Sprintf("ensembler type given: %T", v))
+	}
+
+	ensemblingJob, err := c.EnsemblingJobService.CreateEnsemblingJob(request, projectID, pyFuncEnsembler)
+	if err != nil {
+		return BadRequest("could not create request", err.Error())
```
romanwozniak, line 55: nitpicking: this likely to be `InternalServerError`, rather than `BadRequest`

api/turing/api/ensembling_job_api.go
```diff
@@ -0,0 +1,71 @@
...
+	body interface{},
+) *Response {
+	var errResp *Response
+	var project *mlp.Project
+	if project, errResp = c.getProjectFromRequestVars(vars); errResp != nil {
+		return errResp
+	}
+
+	// Check is done in api.handlers
+	request, _ := body.(*models.EnsemblingJob)
```
romanwozniak, line 31: nitpicking: `request` -> `job` makes more sense to me

api/turing/service/ensembling_job_service.go
```diff
@@ -0,0 +1,151 @@
...
+	Status    *models.Status
+}
+
+// EnsemblingJobService is the data access object for the EnsemblingJob from the db.
+type EnsemblingJobService interface {
+	Save(ensembler *models.EnsemblingJob) error
+	FindByID(id models.ID, options EnsemblingJobFindByIDOptions) (*models.EnsemblingJob, error)
+	List(options EnsemblingJobListOptions) (*PaginatedResults, error)
+	CreateEnsemblingJob(
+		request *models.EnsemblingJob,
```
romanwozniak, line 34: nitpicking: `request` -> `job`

api/turing/service/ensembling_job_service.go
romanwozniak: You can exclude it from linter: https://golangci-lint.run/usage/false-positives/#nolint

romanwozniak (approved):
@ashwinath thanks for addressing my comments. I left few minor remarks, but it LGTM 🚀

api/turing/api/ensembling_job_api.go
ashwinath: you're right, this should be a 5xx error.

api/turing/service/ensembling_job_service.go
ashwinath: Alright removed the linting error and put that line back in.

ashwinath:
Thanks for the time spent on the review, will merge it in when the tests passes.

PR #56, opened 2021-04-30
Fix Golang package of turing ensember.

ashwinath:
Fixes:
Readme has some discrepencies between actual API and example API for the batch ensembler.
Initialise Go module for the batch ensembler so that the generated go code can be in the main module.
Namespaced Go package to be consistent with gojek/turing.


engines/batch-ensembler/README.md
```diff
@@ -56,7 +56,7 @@ spec:
         options:
           viewsEnabled: "true"
           materializationDataset: dataset
-    joinOn:
+    join_on:
```
romanwozniak, line 59: Why is that? There is no a consistent naming convention with YAML (so both camel-case and snake-case are fine), but I use camel-case here (for instance, there is also `bqConfig ` at line:38).
Kubernetes uses camel case as well.

ashwinath: Oh okay I didn't realise this, will change this back to camel.

romanwozniak (approved):
LGTM

PR #57, opened 2021-05-11
Add Graceful Shutdown

anonychun:
Shutdown gracefully shuts down the server without interrupting any active connections. Shutdown works by first closing all open listeners, then closing all idle connections, and then waiting indefinitely for connections to return to idle and then shut down. If the provided context expires before the shutdown is complete, Shutdown returns the context’s error, otherwise it returns any error returned from closing the Server’s underlying

### DIFF
--- a/api/turing/cmd/main.go
+++ b/api/turing/cmd/main.go
@@ -1,10 +1,14 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net/http"
+	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/rs/cors"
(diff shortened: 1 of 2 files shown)


krithika369 (changes requested):
Thanks @anonychun for this PR! I added some comments inline. I have a few more requests.

1. It would be useful to update the PR description with the Go doc that your changes are based on, such as [this](https://golang.org/pkg/net/http/#Server.Shutdown), for anyone wondering if this is a common pattern or custom logic. :)
2. With this change, we have another thing to take care of, concerning the deployment. Our Kubernetes [readiness and liveness probes](https://github.com/gojek/turing/blob/main/infra/chart/templates/deployment.yaml#L61) are both based on the HTTP server. During the graceful shutdown, the server would no longer accept new connections and the liveness probe would fail. If the failure threshold for the liveness probe is reached, the pod would be forcefully killed anyway ([ref](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/)). So, it's good to configure the default values to cater to this. A good explanation on this behaviour can be found here: https://github.com/RisingStack/kubernetes-graceful-shutdown-example
3. Similar to #2 (the previous configs was for the API server), the probes are defined for the router that is deployed by the API module [here](https://github.com/gojek/turing/blob/main/api/turing/cluster/models.go#L101). Ideally, the threshold for the readiness probe is lower than the liveness probe.

Do let us know if you need any further clarification on this.

romanwozniak:
Hey @anonychun, thanks for this PR! Is it possible to combine this functionality into a single function to be shared between `api` and `router` to reduce the code duplication?

romanwozniak:
Closing this PR since there is no interest on it from the contributor.

PR #58, opened 2021-05-19
Make maxAllowedReplica config value configurable from env vars

romanwozniak:
As per title, currently it's hardcoded to `10`, but 10 could be not enough for some of the use cases, so we would like to make it more flexible and configured at the build time


krithika369 (approved):
LGTM

PR #59, opened 2021-05-24
Turing SDK: Define and save Pyfunc Ensemblers

romanwozniak:
### Context:
This MR introduces a new `sdk` module, that contains a skeleton of Turing SDK, related to work with pyfunc ensemblers.

### Changes:
- Reformatted openapi specs (and broke down definitions into multiple fiels), to define a minimal spec, that contains only these API methods, that are used by SDK
- Generated REST client using [openapi-generator](https://github.com/OpenAPITools/openapi-generator) (hence, everything under [sdk/turing/generated](https://github.com/romanwozniak/turing/tree/sdk_save_ensembler/sdk/turing/generated) can be reviewed with less attention to details)
- Added `ProjectsController` into turing-api to exposed MLP projects to Turing SDK
- Moved PyFuncEnsembler interface from `engines/batch-ensembler` into `sdk` and added `sdk` as a dependency on `batch-ensembler`
- Defined SDK methods sufficient for listing projects, creating/updating/listing pyfunc ensemblers
- Added GitHub workflow for building/publishing SDK to pypi repo
- Added a [quickstart](https://github.com/romanwozniak/turing/tree/sdk_save_ensembler/sdk/samples/quickstart) example on how to use SDK

### To be done before merged:
- Unit tests
- Add Python docs where missing


.github/workflows/e2e.yaml
```diff
@@ -22,6 +29,11 @@ jobs:
       with:
         go-version: "1.14"
 
+    - name: Reclaim some disk space
+      run: |
+        sudo apt-get clean
+        df -h
```
ashwinath, line 35: Don't think this is required.

api/turing/it/database/database.go
```diff
@@ -40,15 +40,15 @@ func migrate(db *sql.DB, dbName string) (*sql.DB, error) {
 
 	if migrations, err := gomigrate.NewWithDatabaseInstance(fmt.Sprintf("file://%s", migrationsFolder),
 		dbName, driver); err != nil {
-		return nil, err
+		return db, err
```
ashwinath, line 43: What is the reason behind this change?

api/turing/middleware/authorization.go
```diff
@@ -65,6 +66,32 @@ func (a *Authorizer) Middleware(next http.Handler) http.Handler {
 	})
 }
 
+func (a *Authorizer) FilterAuthorizedProjects(
+	user string,
+	projects []mlp.Project,
+	action string,
+) ([]mlp.Project, error) {
+	projectIDs := make([]string, 0)
+	var allowedProjects []mlp.Project
```
ashwinath, line 75: Nitpick: I think moving this variable declaration just before using it improves readability. i.e moving this line to above 88.

sdk/tests/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def json_serializer(o):
+    if isinstance(o, (date, datetime)):
+        return o.isoformat()
+    if isinstance(o, client.model_utils.ModelNormal):
+        return o.to_dict()
+
+
+project_1 = client.models.Project(
```
krithika369, line 12: Curious - is this common practice to define test data in the `__init__.py` rather than pytest fixtures ?

api/turing/service/ensembler_service.go
```diff
@@ -93,5 +98,5 @@ func (service *ensemblersService) Save(ensembler models.EnsemblerLike) (models.E
 	if err := result.Error; err != nil {
 		return nil, err
 	}
-	return ensembler, nil
+	return service.FindByID(ensembler.GetID(), EnsemblersFindByIDOptions{})
```
ashwinath, line 101: Why do you call FindByID again? Calling db.Save(ensembler) should mutate the ensembler value, shouldn't it?

ashwinath: Also, the error returned by this method will be confusing as a Save method might return a FindByID error.

ashwinath:
I have reviewed the Go code for now, will review the SDK code on Thursday.

sdk/turing/_base_types.py
```diff
@@ -0,0 +1,78 @@
...
+        self._id = id
+        self._created_at = created_at
+        self._updated_at = updated_at
+
+    @property
+    def id(self) -> int:
+        return self._id
+
+    @id.setter
+    def id(self, id: int):
```
krithika369, line 21: Why do we have these setters? Do we expect to use them or is it for completeness?

krithika369 (approved):
It's a LARGE MR. 😅

Left some questions / minor comments but on the whole LGTM. Thanks!

sdk/turing/ensembler.py
```diff
@@ -0,0 +1,314 @@
...
+
+        :param features: pandas.Series, containing a single row with input features
+        :param predictions: pandas.Series, containing a single row with all models predictions
+                `predictions['model-a']` will contain prediction results from the model-a
+        :param treatment_config: dictionary, representing the configuration of a treatment,
+                that should be applied to a given record. If the experiment engine is not configured
+                for this Batch experiment, then `treatment_config` will be `None`
+
+        :returns ensembling result (one of str, int, float, double or array)
+        """
```
ashwinath, line 32: missing a `pass` here?

sdk/turing/session.py
```diff
@@ -0,0 +1,162 @@
...
+        """
+        Update existing ensembler
+        """
+        return EnsemblerApi(self._api_client).update_ensembler(
+            project_id=ensembler.project_id,
+            ensembler_id=ensembler.id,
+            ensembler=ensembler)
+
+
+active_session: TuringSession = TuringSession(
```
ashwinath, line 138: what's this for?

api/ensemblers.yaml
```diff
@@ -0,0 +1,171 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Ensembler"
+
+  "/projects/{project_id}/ensemblers/{ensembler_id}":
+    get:
+      tags: [ "Ensembler" ]
+      operationId: "GetEnsemblerDetails"
+      summary: Returns Ensembler"s details by its ID
```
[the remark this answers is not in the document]
ashwinath: Hey it's an apostrophe!

sdk/Makefile
```diff
@@ -0,0 +1,25 @@
...
+
+env:
+	pip install -r requirements.txt
+
+dev: env
+	pip install -r requirements.dev.txt
+
+gen-client:
+	rm -rf turing/generated
+	@openapi-generator generate \
```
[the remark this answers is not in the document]
romanwozniak: Yeah, `README.md` is brewing 🍵

engines/batch-ensembler/ensembler/components/experimentation.py
```diff
@@ -1,58 +0,0 @@
-from abc import abstractmethod
```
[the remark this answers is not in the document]
romanwozniak: It was only temporarily added into `engines/batch-ensembler` until the SDK package is ready. It's expected from users to install `turing-sdk` to implement (and save) their pyfunc ensemblers, and not the `engines/`batch-ensembler` (which is just a spark application, that is executed inside spark cluster)

api/openapi-batch.yaml
```diff
@@ -0,0 +1,174 @@
...
+      summary: Returns a list of Ensembling jobs
+      parameters:
+        - in: path
+          name: project_id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+      responses:
+        "200":
+          description: TBU
```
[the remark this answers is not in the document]
romanwozniak: Not exactly, it was already updated in `api/swagger-batch.yaml` by @ashwinath, but I'll clean it up in a separate PR this mess with openapi/swagger specs that we have now after I pulled @ashwinath's PR

api/openapi-sdk.yaml
```diff
@@ -0,0 +1,13 @@
+openapi: 3.0.3
+info:
+  title: Turing Minimal Openapi Spec for SDK
+  version: 0.0.1
+servers:
+  - url: localhost:8080/v1
+paths:
+  "/projects":
```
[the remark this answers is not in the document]
romanwozniak: Likely, yes. In this case you wouldn't need "turing session", but rather can do everything within "mlp session", so the SDK can send calls directly to mlp api

sdk/samples/quickstart/quick_start.py
```diff
@@ -0,0 +1,76 @@
...
+    turing.set_url(turing_api)
+    turing.set_project(project)
+
+    # List projects
+    projects = turing.Project.list()
+    for p in projects:
+        print(p)
+
+    # Implement new pyfunc ensembler
+    class MyEnsembler(turing.ensembler.PyFunc):
```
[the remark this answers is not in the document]
romanwozniak: Did it for the explicitness of the flow (i.e. to define "Implement new pyfunc ensembler" as a separate step in this process). But yeah, maybe it's not necessary

sdk/samples/quickstart/quick_start.py
```diff
@@ -0,0 +1,76 @@
...
+        ensembler_instance=MyEnsembler(),
+        conda_env={
+            'channels': ['defaults'],
+            'dependencies': [
+                'python=3.7.0',
+                "cookiecutter>=1.7.2",
+                "numpy"
+            ]
+        },
+        code_dir=["../samples"],
```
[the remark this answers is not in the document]
romanwozniak: It doesn't matter. this code dir is just uploaded together with the pickled ensembler and added into the PYTHONPATH. So if the `code_dir` is defined `"../samples"` than it can be referred inside the ensembler code as `from samples.quickstart.quick_start import main`

sdk/tests/__init__.py
romanwozniak: It's wild wild west, everyone is doing it in their own way :)

but on a serious note, I didn't want to make it as individual fixtures (such as `project_1`, `project_2`...) because then they will have to be included into the test method's argument list and it could potentially grow arguments list unbearable long. 
I'll re-evaluate what can be made as a fixture after I'm done with all the tests

sdk/turing/_base_types.py
romanwozniak: For completeness

.github/workflows/e2e.yaml
romanwozniak: yeah, maybe it's not. e2e test is still failing time to time even with this explicit clean-up step, so maybe it's unrelated

api/turing/it/database/database.go
romanwozniak: In some scenarios, it was possible that this snippet will throw nil pointer. For instance, if the migration fails (because of typo in the migration file), this line will return `nil` and then there will be a panic at https://github.com/gojek/turing/blob/b67c48b7e1b1777f89ec6c6860e3965c516fff9f/api/turing/it/database/database.go#L77, when `.Close()` is called on `nil`

api/turing/middleware/authorization.go
romanwozniak: agreed

sdk/turing/session.py
romanwozniak: To store the information about what project the user is working under and to provide the bridge between SDK methods, exposed to the user, and API endpoints

sdk/turing/ensembler.py
romanwozniak: 💯

api/turing/service/ensembler_service.go
romanwozniak: > Why do you call FindByID again? Calling db.Save(ensembler) should mutate the ensembler value, shouldn't it?

Only the fields, that were mutated explicitly (i.e. new `name` is set and then saved), but not for fields, updated automagically (`updated_at`).

romanwozniak: > Also, the error returned by this method will be confusing as a Save method might return a FindByID error.

That's less of the concern. It was done the same way for other services: https://github.com/gojek/turing/blob/b67c48b7e1b1777f89ec6c6860e3965c516fff9f/api/turing/service/router_service.go#L71

.github/workflows/e2e.yaml
romanwozniak: you are right, there is no any visible effect of this command, removed

api/turing/service/ensembler_service.go
ashwinath: Hmm I wonder if gorm allows for the postgres keyword `RETURNING` though. But not consequential

sdk/turing/session.py
ashwinath: But what if the turing is not hosted on the user's machine? Importing this file will cause this to run and throw an exception?

api/turing/service/ensembler_service.go
romanwozniak: No, gorm doesn't support it: https://github.com/go-gorm/gorm/issues/4088

sdk/turing/session.py
romanwozniak: It's just a default session, it's expected that user will initialize it with a proper Turing API before using:
```python
turing.set_url("http://turing.com/api", use_google_oauth=True)
```

If user has skipped this step, then the error will only occur when user will try to call Turing API with one of available methods, for instance `turing.Project.list()`. The error is not being thrown just on import.

ashwinath (approved):
Just some E2E tests to fix.

sdk/Makefile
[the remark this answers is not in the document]
romanwozniak: ✅

sdk/tests/__init__.py
romanwozniak: This is addressed now too

romanwozniak:
> Just some E2E tests to fix.

The test is flaky likely due to limited storage available in runner (but this theory has to be proved).

PR #60, opened 2021-05-31
Refactor swagger/openapi specs

romanwozniak:
#### Context
With more methods added into Turing API, it's getting harder to maintain `swagger.yaml` (which has grown to ~1400 lines even before recent API additions). This PR is an attempt to modularize spec definitions and also upgrade them to OAS3.0.


api/.env.development
```diff
@@ -31,4 +31,3 @@ MLP_URL=http://mlp.example.com/api/v1
 MLP_ENCRYPTION_KEY=
 
 TURING_ENCRYPTION_KEY=test_encryption_key
-SWAGGER_FILE=./swagger.yaml
```
krithika369, line 34: Why is this removed? Is it because the default value exists and this is not really needed? https://github.com/gojek/turing/blob/main/api/turing/config/config.go#L326

api/openapi/common.yaml
```diff
@@ -0,0 +1,34 @@
...
+      type: "integer"
+      format: "int32"
+
+    FieldSource:
+      type: "string"
+      enum:
+        - "header"
+        - "payload"
+
+    pagination.Paging:
```
krithika369, line 20: Why is this schema named like this, instead of the old `PaginatedResults` ?

api/turing/config/example.yaml
```diff
@@ -104,7 +104,9 @@ TuringUIConfig:
   Homepage: /turing
  
 # File path to find OpenAPI v2 specfication file
```
krithika369, line 106: This should be v3 ?

infra/chart/templates/deployment.yaml
```diff
@@ -111,7 +111,7 @@ spec:
         - name: LAYOUT
           value: "BaseLayout"
         - name: SWAGGER_JSON
-          value: "/app/swagger.yaml"
+          value: "/app/openapi.yaml"
         - name: PORT
           value: "{{ .Values.swaggerUi.service.internalPort }}"
         - name: API_HOST
```
krithika369, line 117: Is anything about this configuration expected to change with OpenAPI V3, since we are using the server URL instead of host and basepath?

krithika369 (approved):
LGTM - it looks well organised, left some questions.

api/.env.development
romanwozniak: It's because with recent changes from @ashwinath, we introduced support of multiple spec files, so the validation of API endpoints can be done among all of them. And with that, `SWAGGER_FILE` is no longer in use (the config key was replaced with `SwaggerFiles` of `array` type). 

But now I'm going to revert it because this proposed specs split makes it possible to construct a fairly large single spec file while keeping it maintainable.

api/openapi/common.yaml
romanwozniak: `PaginatedResults` is more like a composite type of `paging` (information about current page number and total number of pages) + `results` (array with data objects). This `pagination.Paging` is just a schema of a former.

But if the question is more about using this `pagination.` prefix, then I'm just experimenting with it to achieve a better grouping of types between packages. (There is a chance that this will be changed later)

api/turing/config/example.yaml
romanwozniak: The description should definitely be changed, but it not necessarily need to be `v3`, because currently both v2 and v3 are supported. More explanations are here: https://github.com/gojek/turing/pull/60/files/bfe9b89b7e6e9ee1e8d9a06bddab56a3eec3ca90#r643222969

infra/chart/templates/deployment.yaml
romanwozniak: Likely, yes. As well as the deployment chart and its `initContainer` script. Will address this

api/.env.development
romanwozniak: Reverted it now

infra/chart/templates/deployment.yaml
romanwozniak: Done

PR #62, opened 2021-06-07
Bump version of Turing's chart

romanwozniak:
In the [previous MR](https://github.com/gojek/turing/pull/60), the Helm chart of Turing was changed, however the version of it remained the same. This resulted in the [failure](https://github.com/gojek/turing/actions/runs/897393012) of the Github workflow, caused by the fact, that Git tag with the current version already exists in the repo. 

This is an expected behaviour of [chart-releaser-action](https://github.com/helm/chart-releaser-action), because it checks the diff of the chart itself and creates a new Github release whenever the changes are detected. If release with the same chart version already exists, then this translates into an observed failure. Meaning, that Chart's version should always be updated if chart files were changed.

### DIFF
--- a/infra/chart/Chart.yaml
+++ b/infra/chart/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.1.0
+version: 0.1.1


ashwinath (approved):

ashwinath:
Actually I would think that we should only generate the helm charts on tagging rather than tracking master branch.

krithika369:
> Actually I would think that we should only generate the helm charts on tagging rather than tracking master branch.

The [Helm releaser action](https://github.com/helm/chart-releaser-action) creates a release (which in turn creates a tag) every time there is a change to the chart version:
_"... by checking each chart in your project, and whenever there's a new chart version, creates a corresponding GitHub release named for the chart version"_

@ashwinath So it's kind of the reverse to what you're saying. But I see your point.

With our Github project, we don't have other tags or releases right now - for example, release of the docker images only happens in our internal CD pipeline. I don't know if we will ever start supporting that on the public repo (if we do, we may run into issues of breaking changes between the Helm chart and other tagged artifacts), but until then, perhaps the current set up will be ok.

krithika369:
Will merge this PR to fix the failure status on the repo's CI.

PR #63, opened 2021-06-08
List/Get Ensembling job endpoints

ashwinath:
This PR introduces the get and list endpoints for ensembling jobs.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,7 +21,7 @@ jobs:
 
     steps:
     - name: Maximize build space
-      uses: ashwinath/maximize-build-space@master
+      uses: easimon/maximize-build-space@b4d02c14493a9653fe7af06cc89ca5298071c66e
       with:
         root-reserve-mb: 40000
         swap-size-mb: 1024
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -20,6 +20,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Maximize build space
+      uses: ashwinath/maximize-build-space@master
+      with:
+        root-reserve-mb: 40000
+        swap-size-mb: 1024
+        remove-dotnet: 'true'
+        remove-android: 'true'
+        remove-haskell: 'true'
 
     - name: Check out code
       uses: actions/checkout@v2
--- a/api/openapi/jobs.yaml
+++ b/api/openapi/jobs.yaml
@@ -10,6 +10,7 @@ paths:
   "/projects/{project_id}/jobs":
     post:
       tags: *tags
+      operationId: "CreateEnsemblingJob"
       summary: Submit an Ensembling job.
       parameters:
         - in: path
@@ -27,6 +28,30 @@ paths:
       responses:
         "202":
           description: Accepted.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/EnsemblingJob"
+
+  "/projects/{project_id}/jobs/{id}":
+    get:
+      tags: *tags
+      operationId: "GetEnsemblingJob"
+      summary: Get an existing Ensembling job.
+      parameters:
+        - in: path
+          name: project_id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+        - in: path
+          name: id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+      responses:
+        "200":
+          description: A JSON object representing an Ensembling Job
           content:
             application/json:
               schema:
--- a/api/openapi/jobs.yaml
+++ b/api/openapi/jobs.yaml
@@ -32,6 +32,33 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/EnsemblingJob"
+    get:
+      tags: *tags
+      operationId: "ListEnsemblingJobs"
+      summary: Returns a list of ensembling jobs that belong to the project
+      parameters:
+        - in: path
+          name: project_id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+        - in: query
+          name: page
+          schema:
+            type: integer
+            default: 1
+        - in: query
+          name: page_size
+          schema:
+            type: integer
+            default: 10
+      responses:
+        "200":
+          description: A JSON object, that represents paginated results response
+          content:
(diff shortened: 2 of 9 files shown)


api/openapi/jobs.yaml
```diff
@@ -10,6 +10,7 @@ paths:
   "/projects/{project_id}/jobs":
     post:
       tags: *tags
+      operationId: "CreateEnsemblingJob"
```
krithika369, line 13: Why has this been introduced?

api/turing/api/ensembling_job_api.go
```diff
@@ -67,5 +118,22 @@ func (c EnsemblingJobController) Routes() []Route {
...
+			method:  http.MethodGet,
+			path:    "/projects/{project_id}/jobs",
+			handler: c.ListEnsemblingJobs,
+		},
 	}
 }
+
+// GetEnsemblingJobOptions is the options that can be parsed
+// from query params for the GET ensembling job method
+type GetEnsemblingJobOptions struct {
```
krithika369, line 136: Nitpick: For consistency with the [Ensemblers API](https://github.com/gojek/turing/blob/main/api/turing/api/ensemblers_api.go#L179), can we name this `EnsemblingJobPathOptions`?

api/turing/service/ensembling_job_service.go
```diff
@@ -21,8 +21,8 @@ type EnsemblingJobFindByIDOptions struct {
 // EnsemblingJobListOptions holds query parameters for EnsemblersService.List method.
 type EnsemblingJobListOptions struct {
 	PaginationOptions
-	ProjectID *models.ID
-	Status    *models.Status `validate:"oneof=pending running terminating terminated completed failed failed_submission failed_building"` //nolint
+	ProjectID *models.ID `schema:"project_id" validate:"required"`
```
krithika369, line 24: How is the `schema:` tag used?

api/turing/service/ensembling_job_service.go
```diff
@@ -21,8 +21,8 @@ type EnsemblingJobFindByIDOptions struct {
...
+	Status    *models.Status
```
krithika369, line 25: Why has the `validate` tag been removed?

api/turing/api/ensembling_job_api_test.go
```diff
@@ -385,3 +385,205 @@ func TestEnsemblingJobController_CreateEnsemblingJob(t *testing.T) {
 		})
 	}
 }
+
+func TestEnsemblingJobController_GetEnsemblingJob(t *testing.T) {
```
krithika369, line 389: It would be good to add a test case with more than 1 total pages and page number > 1,

krithika369 (approved):
Left some remarks and questions. On the whole, LGTM. Thanks.

.github/workflows/e2e.yaml
```diff
@@ -20,6 +20,14 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Maximize build space
+      uses: ashwinath/maximize-build-space@master
```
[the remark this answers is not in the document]
ashwinath: Oh, because what if that person decides to do something nefarious in the later commits, like exposing keys etc. Tags can also be overriden by rewriting the Git history.

api/openapi/jobs.yaml
ashwinath: Good question, I'm not sure also, I saw @romanwozniak code having it so I decided to be consistent

api/turing/api/ensembling_job_api.go
ashwinath: Ah because in my other PR it's not always just Path options. Sometimes it's used internally by the batch ensembling process. Not all options are used in path. Path is a subset of what list does.

api/turing/service/ensembling_job_service.go
ashwinath: The schema is used by parsing the query params into here: https://github.com/gojek/turing/pull/63/files#diff-8d901405e0fd6037fb5bb3c53aade949ea58f83a31c57b302a6983eb3c8eee60R96

I saw that @romanwozniak used this technique to parse the query params into the struct.

api/turing/service/ensembling_job_service.go
ashwinath: It was removed because now it is required if you want to filter by status, which I did not intend. Anyway in my other PR I have changed this to accept a list of statuses instead.

.github/workflows/e2e.yaml
```diff
@@ -20,6 +20,14 @@ jobs:
...
+      with:
+        root-reserve-mb: 40000
+        swap-size-mb: 1024
+        remove-dotnet: 'true'
```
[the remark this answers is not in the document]
ashwinath: Oh the test will fail if this is not here, kinda sucks though

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
krithika369: Oh in that case, you could reference the specific commit directly: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-versioned-actions

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
krithika369: Right.. so I'm wondering if we eventually want to introduce some workflow changes to reduce the frequency at which the e2e tests are run (the tests themselves are only ~4 mins while the infra set up and the step to add space together make up ~12 mins). But this is something we can discuss internally later I guess.

api/turing/service/ensembling_job_service.go
krithika369: So, should `Status` be removed altogether from this PR?

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
ashwinath: ok i'll do that

api/turing/service/ensembling_job_service.go
ashwinath: Err, there's some unit tests referencing this so I don't think it's worth the effort.

ashwinath: Anyway in my other PR I will add in the querying multiple statuses, seems like a logical flow as well.

ashwinath:
Thanks for the review @krithika369

PR #64, opened 2021-06-15
Add New Relic integration.

ashwinath:
New Relic integration has been put added to wrap all handlers, but the init component has not been run, therefore it was not pushing to New Relic. This PR adds the initialisation of New Relic.

### DIFF
--- a/api/turing/cmd/main.go
+++ b/api/turing/cmd/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/rs/cors"
 
 	"github.com/gojek/mlp/pkg/authz/enforcer"
+	"github.com/gojek/mlp/pkg/instrumentation/newrelic"
 	"github.com/gojek/mlp/pkg/instrumentation/sentry"
 	"github.com/gojek/turing/api/turing/api"
 	"github.com/gojek/turing/api/turing/config"
@@ -65,6 +66,12 @@ func main() {
 	}
 	db.LogMode(false)
 	defer db.Close()
+
+	// Initialise NewRelic
+	if err := newrelic.InitNewRelic(cfg.NewRelicConfig); err != nil {
+		log.Errorf("Failed to initialize newrelic: %s", err)
+	}
+	defer newrelic.Shutdown(5 * time.Second)
 
 	// Init auth enforcer, vault client
 	var authEnforcer *enforcer.Enforcer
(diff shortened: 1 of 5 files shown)


api/turing/cmd/main.go
```diff
@@ -66,6 +67,12 @@ func main() {
 	db.LogMode(false)
 	defer db.Close()
 
+	// Initialise NewRelic
```
krithika369, line 70: Nitpick: Can this come after the Sentry initialisation so that the somewhat similar steps are placed together?

api/turing/config/config.go
```diff
@@ -72,7 +73,8 @@ type Config struct {
 	//
 	// For example:
 	// { "experiment_engine_a": {"client": "foo"}, "experiment_engine_b": {"apikey": 12} }
-	Experiment map[string]interface{}
+	Experiment     map[string]interface{}
+	NewRelicConfig newrelic.Config
```
krithika369, line 77: Nitpick: Can we move this immediately above / below the Sentry config so that the similar configs are together in the type definition?

krithika369: Can we also add a simple test case for the config?

api/turing/config/example.yaml
```diff
@@ -28,6 +28,19 @@ DeployConfig:
   MaxCPU: "4"
   MaxMemory: "8Gi"
 
+# New Relic Config
+NewRelic:
+  Enabled: true
+  AppName: turing
+  License: <your very secret license key here>
+  IgnoreStatusCodes: 
+    - 400
```
krithika369, line 37: What is the rationale for ignoring these codes?

krithika369 (approved):
Left some minor comments, other than that it LGTM. Please feel free to merge after addressing them. Thanks.

api/turing/cmd/main.go
ashwinath: Unfortunately I can't because it needs to intialise the NewRelic before the handler initialisation is done. If not this becomes the implementation of newrelic (https://github.com/gojek/mlp/blob/main/api/pkg/instrumentation/newrelic/newrelic.go#L12)

api/turing/config/example.yaml
ashwinath: This is just an example.

api/turing/config/config.go
ashwinath: yup done both.

infra/chart/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.1.1
+version: 0.1.2
```
krithika369, line 4: @ashwinath why is the chart version being bumped up? I don't think it is needed since the chart folder is not modified otherwise.

ashwinath: alright reverting.

PR #66, opened 2021-06-30
Build ensembler on master.

ashwinath:
Currently, Github actions is not publishing to master. This will allow ensemblers to build on master.

### DIFF
--- a/.github/workflows/engines-batch-ensembler.yaml
+++ b/.github/workflows/engines-batch-ensembler.yaml
@@ -60,6 +60,8 @@ jobs:
       ) || (
         github.event_name == 'push' &&
         github.event.pull_request.head.repo.full_name == github.repository
+      ) || (
+        github.ref == 'refs/heads/master'
       )
     runs-on: ubuntu-latest
     defaults:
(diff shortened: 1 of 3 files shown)


.github/workflows/engines-batch-ensembler.yaml
```diff
@@ -60,6 +60,8 @@ jobs:
       ) || (
         github.event_name == 'push' &&
         github.event.pull_request.head.repo.full_name == github.repository
+      ) || (
+        github.ref == 'refs/heads/master'
```
romanwozniak, line 64: This entire `if` condition should be reviewed and if necessary – cleaned (I think it was copied from somewhere). But in the context of this PR – LGTM

romanwozniak (approved):

.github/workflows/engines-batch-ensembler.yaml
ashwinath: yeah this master part was copied.

ashwinath:
The E2E tests were failing (strange as it was working yesterday). I'm not sure what was going on here: https://github.com/gojek/turing/runs/2950046686?check_suite_focus=true#step:4:224 but it seems like the kube-scheduler had some permissions issue and could not list/add resources. Upgrading the Kind and Kubernetes version seemed to help.

To fix it, I upgraded the Kind CLI version and the Kubernetes patch version (1.17.11 -> 1.17.17).

romanwozniak (approved):
I'm not sure if the fix is related to the upgrade of infra dependencies, but it doesn't hurt upgrading anyway, so lgtm

PR #67, opened 2021-06-30
Fix: Change building of ensembler image from master branch to main.

ashwinath:
I forgot master branch is now called main in Github. This fixes the builds for the ensembling job that are supposed to be for main branch.


romanwozniak (approved):
Damn, I haven't captured this as well :)

PR #69, opened 2021-07-01
Terminate ensembling job

ashwinath:
This PR adds the `DELETE` `/projects/{project_id}/jobs/{id}` to terminate a job.

The API accepts a request to delete the record and sets the status to `terminating`. The batch runners will periodly check if it is time to delete the associated Kubernetes resources (Job and SparkApplication) before physically deleting the record from the database.

Checks have to be done at different stages of the orchestration process, making sure that is no race conditions (I have tested this extensively) during the termination process and building process.

### DIFF
--- a/api/openapi/jobs.yaml
+++ b/api/openapi/jobs.yaml
@@ -66,7 +66,7 @@ paths:
               schema:
                 $ref: "#/components/schemas/EnsemblingJobPaginatedResults"
 
-  "/projects/{project_id}/jobs/{id}":
+  "/projects/{project_id}/jobs/{job_id}":
     get:
       tags: *tags
       operationId: "GetEnsemblingJob"
@@ -78,7 +78,7 @@ paths:
             $ref: "#/components/schemas/Id"
           required: true
         - in: path
-          name: id
+          name: job_id
           schema:
             $ref: "#/components/schemas/Id"
           required: true
@@ -95,7 +95,7 @@ paths:
     delete:
       tags: *tags
       operationId: "DeleteEnsemblingJob"
-      summary: Deletee an ongoing Ensembling Job.
+      summary: Delete an ongoing Ensembling Job.
       parameters:
         - in: path
           name: project_id
@@ -103,7 +103,7 @@ paths:
             $ref: "#/components/schemas/Id"
           required: true
         - in: path
-          name: id
+          name: job_id
           schema:
             $ref: "#/components/schemas/Id"
           required: true
--- a/api/openapi/jobs.yaml
+++ b/api/openapi/jobs.yaml
@@ -26,7 +26,7 @@ paths:
             schema:
               $ref: "#/components/schemas/EnsemblingJob"
       responses:
-        "202":
+        202:
           description: Accepted.
           content:
             application/json:
@@ -59,7 +59,7 @@ paths:
             items:
               $ref: "#/components/schemas/EnsemblerJobStatus"
       responses:
-        "200":
+        200:
           description: A JSON object, that represents paginated results response
           content:
             application/json:
@@ -83,12 +83,44 @@ paths:
             $ref: "#/components/schemas/Id"
           required: true
       responses:
-        "200":
+        200:
           description: A JSON object representing an Ensembling Job
           content:
             application/json:
               schema:
                 $ref: "#/components/schemas/EnsemblingJob"
+        404:
+          description: "Invalid ensembling job"
+
+    delete:
+      tags: *tags
+      operationId: "DeleteEnsemblingJob"
+      summary: Deletee an ongoing Ensembling Job.
+      parameters:
+        - in: path
+          name: project_id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+        - in: path
+          name: id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+      responses:
+        202:
+          description: Accepted.
+          content:
+            application/json:
+              schema:
+                type: "object"
+                properties:
+                  id:
+                    $ref: "#/components/schemas/Id"
+        400:
+          description: "Invalid ensembling job"
+        404:
+          description: "Ensembling job not found"
 
 components:
   schemas:
--- a/api/turing/api/ensembling_job_api.go
+++ b/api/turing/api/ensembling_job_api.go
@@ -109,6 +109,47 @@ func (c EnsemblingJobController) ListEnsemblingJobs(
 	return Ok(results)
 }
 
+type deleteEnsemblingJobResponse struct {
+	ID models.ID `schema:"id"`
+}
+
+// DeleteEnsemblingJob deletes and ensembling job and cancels any ongoing process.
+func (c EnsemblingJobController) DeleteEnsemblingJob(
+	_ *http.Request,
+	vars RequestVars,
+	_ interface{},
+) *Response {
+	options := &GetEnsemblingJobOptions{}
+
+	if err := c.ParseVars(options, vars); err != nil {
+		return BadRequest(
+			"failed to fetch ensembling job",
+			fmt.Sprintf("failed to parse query string: %s", err),
+		)
+	}
+
+	ensemblingJob, err := c.EnsemblingJobService.FindByID(
+		*options.ID,
+		service.EnsemblingJobFindByIDOptions{
+			ProjectID: options.ProjectID,
+		},
+	)
+	if err != nil {
+		return NotFound("ensembler not found", err.Error())
+	}
+
+	err = c.EnsemblingJobService.DeleteEnsemblingJob(ensemblingJob)
+	if err != nil {
+		return InternalServerError("unable to delete ensembling job", err.Error())
+	}
+
+	return Accepted(
+		deleteEnsemblingJobResponse{
+			ID: *options.ID,
+		},
+	)
+}
+
 // Routes returns all the HTTP routes given by the EnsemblingJobController.
 func (c EnsemblingJobController) Routes() []Route {
 	return []Route{
--- a/api/turing/api/ensembling_job_api.go
+++ b/api/turing/api/ensembling_job_api.go
@@ -135,10 +135,10 @@ func (c EnsemblingJobController) DeleteEnsemblingJob(
 		},
 	)
 	if err != nil {
-		return NotFound("ensembler not found", err.Error())
+		return NotFound("ensembling job not found", err.Error())
 	}
 
-	err = c.EnsemblingJobService.DeleteEnsemblingJob(ensemblingJob)
+	err = c.EnsemblingJobService.MarkEnsemblingJobForTermination(ensemblingJob)
 	if err != nil {
 		return InternalServerError("unable to delete ensembling job", err.Error())
 	}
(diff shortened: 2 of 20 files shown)


api/openapi/jobs.yaml
```diff
@@ -83,12 +83,44 @@ paths:
...
+      operationId: "DeleteEnsemblingJob"
+      summary: Deletee an ongoing Ensembling Job.
+      parameters:
+        - in: path
+          name: project_id
+          schema:
+            $ref: "#/components/schemas/Id"
+          required: true
+        - in: path
+          name: id
```
romanwozniak, line 106:
```suggestion
          name: job_id
```
It's better to avoid ambiguity by explicitly specifying id of what is expected here

api/turing/api/ensembling_job_api.go
```diff
@@ -109,6 +109,47 @@ func (c EnsemblingJobController) ListEnsemblingJobs(
...
+	}
+
+	ensemblingJob, err := c.EnsemblingJobService.FindByID(
+		*options.ID,
+		service.EnsemblingJobFindByIDOptions{
+			ProjectID: options.ProjectID,
+		},
+	)
+	if err != nil {
+		return NotFound("ensembler not found", err.Error())
```
romanwozniak, line 138: This message is confusing. Shouldn't it be `Job not found`?

api/turing/batch/ensembling/runner.go
```diff
@@ -96,6 +100,18 @@ func (r *ensemblingJobRunner) updateStatus() {
 		RetryCountLessThan: &r.maxRetryCount,
 	}
 	r.processEnsemblingJobs(optionsJobBuildingImage)
+
+	// JobTerminating
+	optionsJobTerminatingImage := service.EnsemblingJobListOptions{
```
romanwozniak, line 105: Just curious: do we need to maintain some sort of queue to process jobs? I.e. job that was processed (but somehow would still need to be processed again during the next run) should be appended to the end of queue and runner processes only `batchSize` jobs from the top of queue? Also, if there are more than `batchSize` jobs waiting to be processed, shouldn't we just try to process them all before runner falls asleep before the next invocation?

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -176,20 +181,17 @@ func (ib *imageBuilder) waitForJobToFinish(job *apibatchv1.Job) error {
 	for {
 		select {
 		case <-timeout:
-			log.Errorf("timeout waiting for kaniko job completion %s", job.Name)
 			return ErrTimeoutBuildingImage
 		case <-ticker.C:
 			j, err := ib.clusterController.GetJob(ib.imageConfig.BuildNamespace, job.Name)
 			if err != nil {
-				log.Errorf("unable to get job status for job %s: %v", job.Name, err)
```
romanwozniak, line 184: Why was it removed?

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -356,3 +358,26 @@ func (ib *imageBuilder) GetImageBuildingJobStatus(
 
 	return JobStatusUnknown, nil
 }
+
+func (ib *imageBuilder) DeleteImageBuildingJob(
+	projectName string,
+	modelName string,
+	versionID models.ID,
+) error {
+	kanikoJobName := ib.nameGenerator.generateBuilderJobName(
```
romanwozniak, line 367: I wonder if the job name should be stored in the DB as part of the job submission process. Otherwise it leaves a chance for some dangling jobs (scenario: job submitted -> new release of Turing is deployed, where `nameGenerator` implementation is changed -> job is terminated)

api/turing/service/ensembling_job_service.go
```diff
@@ -29,14 +29,16 @@ type EnsemblingJobListOptions struct {
...
+	Save(ensemblingJob *models.EnsemblingJob) error
+	Delete(ensemblingJob *models.EnsemblingJob) error
 	FindByID(id models.ID, options EnsemblingJobFindByIDOptions) (*models.EnsemblingJob, error)
 	List(options EnsemblingJobListOptions) (*PaginatedResults, error)
 	CreateEnsemblingJob(
 		job *models.EnsemblingJob,
 		projectID models.ID,
 		ensembler *models.PyFuncEnsembler,
 	) (*models.EnsemblingJob, error)
+	DeleteEnsemblingJob(ensemblingJob *models.EnsemblingJob) error
```
romanwozniak, line 41: I would like to see the DAO service and orchestration service to be split into two. It's confusing to have `Delete` and `DeleteEnsemblingJob` in a single service

romanwozniak: Now, after inspecting `DeleteEnsemblingJob` implementation, I think that maybe it should be renamed accordingly to what it's doing. Maybe `MarkJobForTermination`?

romanwozniak (commented):
I'm done with review, left some questions and comments

api/openapi/jobs.yaml
ashwinath: ok

api/openapi/jobs.yaml
```diff
@@ -83,12 +83,44 @@ paths:
...
             application/json:
               schema:
                 $ref: "#/components/schemas/EnsemblingJob"
+        404:
+          description: "Invalid ensembling job"
+
+    delete:
+      tags: *tags
+      operationId: "DeleteEnsemblingJob"
+      summary: Deletee an ongoing Ensembling Job.
```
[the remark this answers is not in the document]
ashwinath: ok

api/turing/api/ensembling_job_api.go
ashwinath: ok

api/turing/imagebuilder/imagebuilder.go
ashwinath: Ah this should be back there, it was noisy when i was dealing with some race condition.

api/turing/batch/ensembling/runner.go
ashwinath: I did consider building an in memory queue (I suppose this is what you mean; Merlin does this) but I think it's too much effort for not much gain. I decided on an external queue only (postgres) during the batch orchestration PR. I would argue on the following points:

1. A 10 second (configurable, we can even put it as short as 2-3 seconds) delay is not going to make a difference.
2. We still have to maintain an external state because of unforeseen downtimes.

> Also, if there are more than batchSize jobs waiting to be processed, shouldn't we just try to process them all before runner falls asleep before the next invocation?

Seems like I forgot to rename `batch_size` to `RecordsToProcessInOneIteration` but I did in the config.

Actually, there can be, at one instance, more than `RecordsToProcessInOneIteration` processing at once. If you notice [here](https://github.com/ashwinath/turing/blob/terminate-ensembling-job/api/turing/batch/runner/base.go#L18), it doesn't actually wait for the job to be complete before the next cycle starts. I made sure of this because the consideration I had was not to overload postgres and kube-apiserver. Consider this example where  `RecordsToProcessInOneIteration` = 2:

```
--------------------------------------------------------------> time
---> fire job
    -------------------------------> job 1
    ----------------------------> job 2
<----------> ---> fire job
  Delay
                 -------------------------------> job 3
                 ----------------------------> job 4
                 <-------------->
            overlapping of 4 jobs here.
```
Which is to say that at most the delay will be based on `BatchRunnerConfig.TimeInterval`.

api/turing/service/ensembling_job_service.go
ashwinath: Ya actually MarkJobForTermination is much better.

api/turing/imagebuilder/imagebuilder.go
ashwinath: I think its nice to have that, but there are some considerations:

1. What is the expected life time of the job? If the nameGenerator implementation changes, is it okay to let the termination job fail?
2. Are we expecting to change the naming convention?

api/turing/batch/ensembling/runner.go
romanwozniak: > I decided on an external queue only (postgres)

How exactly does it work? Do you order records by `updated_at` and process first these jobs, that wait in `queue` the longest?

api/turing/imagebuilder/imagebuilder.go
romanwozniak: > Are we expecting to change the naming convention?

Not "expecting", but this can happen for some reason (intentionally or accidentally) and whoever changes the nameGenerator might not even be aware of the implications. 

> is it okay to let the termination job fail?

What is `termination job`?

api/turing/batch/ensembling/runner.go
ashwinath: There is no order by (although its probably good practice to put limit and offsets with order by). In any case this is inconsequential.

The query that picks up the records for processing is this `SELECT * FROM "ensembling_jobs"  WHERE (status IN ('pending','failed_submission','failed_building')) AND (retry_count < 3) LIMIT 10 OFFSET 0`, which is written ]here]( https://github.com/gojek/turing/blob/main/api/turing/batch/ensembling/runner.go#L196-L208.)

At different points of processing the ensembling job, the status changes, which the lifecycle of the finite state machine is stated [here](https://github.com/gojek/turing/blob/main/api/turing/models/ensembling_job.go#L86-L106). This is to ensure we do not double process a record.
```go
// Status is the state of the finite machine ensembling job.
// Possible statuses:
// JobPending --▶ JobFailedSubmission
//     |
//     |
//     |
//     |
// JobBuildingImage --▶ JobFailedBuildImage
//     |
//     |
//     |
//     |
//     ▼
// JobRunning --▶ JobFailed
//     |
//     |
//     |--▶ JobTerminated
//     |
//     |
//     ▼
// JobCompleted
type Status string
```

There is no queue, it's constantly being processed or waiting for I/O. The process can be broken down as follows in pseudocode:

```
while true:
  fire and forget runner <-------- point A
  wait
  
runner():
  processJob
  updateStatus
  
processJob():
  query records
  build image
  wait for build image to be done (blocking operation)
  create secrets/svc acc/rbac etc
  create spark job (don't wait, same as kubectl apply -f sparkapplication.yaml)
  
updateStatus():
  query based on conditions() // be it terminating, imagebuilding or running
  check the status on Kubernetes
  update postgres record  
```

I'm not sure what you mean by waiting. The only waiting that is done is between firing the runners. which is explained at point A above. The actual code is [here](https://github.com/ashwinath/turing/blob/terminate-ensembling-job/api/turing/batch/runner/base.go#L15-L18)

```go
// RunBatchRunners will run all runners asynchronously.
func RunBatchRunners(interval time.Duration, runners []BatchJobRunner) {
	for {
		for _, runner := range runners {
			go runner.Run()
		}
		time.Sleep(interval)
	}
}
```

Since `go runner.Run()` is not a blocking operation, it does a `time.Sleep(interval)` almost immediately. Which allows for the regular firing of batch runners.

The "queue" that we speak of is just the database, we just keep polling the database to check for `pending`/`failed_submission`/`failed_building` records for processing.

ashwinath: The only difference I can see for the order by is either employing a stack/queue or some other picking strategy which I don't think is needed yet or any foreseeable future.

api/turing/imagebuilder/imagebuilder.go
ashwinath: Well I mean the Termination request from the API.

api/turing/batch/ensembling/runner.go
romanwozniak: Ok, thanks for the detailed explanation. I'm not too concerned about it for now.

api/turing/imagebuilder/imagebuilder.go
romanwozniak: okay

romanwozniak (approved):

PR #70, opened 2021-07-06
Bump version of Turing charts.

ashwinath:
The migrations are failing in newer installs because the DB migrations are not in sync. This will update the chart versions so that db migrations are in sync with the codebase.


krithika369 (approved):
LGTM

PR #73, opened 2021-07-08
Add a feature toggle to disable batch ensembling runners

ashwinath:
Added a toggle to turn off the batch ensembling feature.

By design, the API binary will panic if the wrong configurations are given. If the user requires the batch ensembling feature, he should be confident that it should work when it is deployed. As opposed to only finding out the problem when the feature is being used.

Small bug I found in the configuration regarding Kubernetes Tolerations: fixed by using the correct key and remove TaintKey as it is redundant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ DbConfig:
 DeployConfig:
   EnvironmentType: dev 
 EnsemblingJobConfig:
+  Enabled: true
   DefaultEnvironment: dev
   RecordsToProcessInOneIteration: 10
   MaxRetryCount: 3
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ BatchEnsemblingConfig:
     MaxRetryCount: 3
   ImageBuildingConfig:
     ImageConfig:
-      Registry: ghcr.io
+      DestinationRegistry: ghcr.io
       BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
       BuildNamespace: default
       BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ router Docker image is set to the image built and push from the previous step.
 
 ```yaml
 # config-dev.yaml
-BatchEnsemblerConfig:
+BatchEnsemblingConfig:
   Enabled: true
   JobConfig:
     DefaultEnvironment: dev
--- a/README.md
+++ b/README.md
@@ -158,44 +158,46 @@ router Docker image is set to the image built and push from the previous step.
 
 ```yaml
 # config-dev.yaml
-BatchRunnerConfig:
-  TimeInterval: 10s
+BatchEnsemblerConfig:
+  Enabled: true
+  JobConfig:
+    DefaultEnvironment: dev
+    DefaultConfigurations:
+      SparkConfigAnnotations:
+        "spark/spark.sql.execution.arrow.pyspark.enabled": "true"
+      BatchEnsemblingJobResources:
+        DriverCPURequest: "1"
+        DriverMemoryRequest: "1Gi"
+        ExecutorReplica: 2
+        ExecutorCPURequest: "1"
+        ExecutorMemoryRequest: "1Gi"
+  RunnerConfig:
+    TimeInterval: 10s
+    RecordsToProcessInOneIteration: 10
+    MaxRetryCount: 3
+  ImageBuildingConfig:
+    ImageConfig:
+      Registry: ghcr.io
+      BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
+      BuildNamespace: default
+      BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
+      DockerfileFilePath: engines/batch-ensembler/app.Dockerfile
+      BuildTimeoutDuration: 20m
+    KanikoConfig:
+      Image: gcr.io/kaniko-project/executor
+      ImageVersion: v1.5.2
+      ResourceRequestsLimits:
+        Requests:
+          CPU: "1"
+          Memory: 1Gi
+        Limits:
+          CPU: "1"
+          Memory: 1Gi
 DbConfig:
   User: turing
   Password: turing
 DeployConfig:
   EnvironmentType: dev 
-EnsemblingJobConfig:
-  Enabled: true
-  DefaultEnvironment: dev
-  RecordsToProcessInOneIteration: 10
-  MaxRetryCount: 3
-  ImageBuilderConfig:
-    Registry: ghcr.io
-    BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
-    BuildNamespace: default
-    BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
-    DockerfileFilePath: engines/batch-ensembler/app.Dockerfile
-    BuildTimeoutDuration: 20m
-  KanikoConfig:
-    Image: gcr.io/kaniko-project/executor
-    ImageVersion: v1.5.2
-    ResourceRequestsLimits:
-      Requests:
-        CPU: "1"
-        Memory: 1Gi
-      Limits:
-        CPU: "1"
-        Memory: 1Gi
-  DefaultConfigurations:
-    SparkConfigAnnotations:
-      "spark/spark.sql.execution.arrow.pyspark.enabled": "true"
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -166,6 +166,7 @@ DbConfig:
 DeployConfig:
   EnvironmentType: dev 
 EnsemblingJobConfig:
+  Enabled: true
```
romanwozniak, line 169: I brought this point during the orchestration PR review, but here it is relevant again: can we come up with a better structure of the batch ensembling related configuration? Here `Enabled` is under `EnsemblingJobConfig`, but there is also `BatchRunnerConfig` which feels to be even better for this feature toggle. Would it be better to have a single top-level config branch (for instance `BatchEnsembling`) and nest all related configuration under it:

```yaml
BatchEnsembling:
  Enabled: true

  RunnerConfig:
    TimeInterval: 10s
    # ...

  JobConfig:
    DefaultEnvironment: dev
    # ...
```

romanwozniak: Also `RecordsToProcessInOneIteration` and (maybe) `MaxRetryCount` seem to be better defined under `BatchRunnerConfig` rather than under `EnsemblingJobConfig` because they define the behaviour of the runner, not the job iteslf.

ashwinath: Yeah Ill take some time to rearrange and review the configuration

ashwinath: A consequence of the proposed configuration is that every BatchRunner has it's own time interval between runs, I'm not sure if that level of granuality is required.

I would propose either one of the following (group by runners or group by batch job). Option 1 would be similar to your suggestion but I would prefer option 2 because of the grouping by runners. It is probably neater to group by runner. But I do not have a strong opinion on this. @romanwozniak do let me know which option you prefer/other suggestions.

1. Group by batch job
```yaml
BatchEnsembling:
  Enabled: true
  
  # Job Config -> common properties
  JobConfig:
    DefaultEnvironment: dev
  
  # Runner configuration -> corresponds to runner.go
  RunnerConfig:
    TimeInterval: 10s
    RecordsToProcessInOneIteration: 10
    MaxRetryCount: 3
    
  # ImageBuilder configuration -> corresponds to the imagebuilder package
  ImageBuilderConfig:
    ImageConfig:
      Registry: ghcr.io
      BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
      BuildNamespace: default
      BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
      DockerfileFilePath: engines/batch-ensembler/app.Dockerfile
      BuildTimeoutDuration: 20m
    KanikoConfig:
      Image: gcr.io/kaniko-project/executor
      ImageVersion: v1.5.2
      ResourceRequestsLimits:
        Requests:
          CPU: "1"
          Memory: 1Gi
        Limits:
          CPU: "1"
          Memory: 1Gi
```

2. Group by runner.
```yaml
BatchRunnerConfig:
  TimeInterval: 10s

  BatchEnsembling:
    Enabled: true
    
    # Job Config -> common properties
    JobConfig:
      DefaultEnvironment: dev

    # RunnerConfig configuration -> corresponds to the runner.go
    RunnerConfig:
      RecordsToProcessInOneIteration: 10
      MaxRetryCount: 3
      
    # ImageBuilder configuration -> corresponds to imagebuilder.go
    ImageBuilderConfig:
      ImageConfig:
        Registry: ghcr.io
        BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
        BuildNamespace: default
        BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
        DockerfileFilePath: engines/batch-ensembler/app.Dockerfile
        BuildTimeoutDuration: 20m
      KanikoConfig:
        Image: gcr.io/kaniko-project/executor
        ImageVersion: v1.5.2
        ResourceRequestsLimits:
          Requests:
            CPU: "1"
            Memory: 1Gi
          Limits:
            CPU: "1"
            Memory: 1Gi

  Foo:
    Enabled: false

  Bar:
    Enabled: false
```

romanwozniak: Option 2 is interesting, but seems like a premature optimization to me. Why would we need to have multiple runners with different configuration? Also, I'd prefer to have a feature toggle as a root level flag, so it can on/off the entire batch ensembling component. With option 2 you rather enable runners, and not the feature itself

ashwinath: Actually option 2 has less abstraction as it groups up on a runner level; it assumes no similarity between runners other than they are able to turn off and on individually.

I think there is some confusion on terminology here, let me set the foundations before we miscommunicate.
Batch Runner -> Runs mutiple batch runners. (i.e. batch ensembling job is one of the many runners that can be configured.)
Batch Ensembling job -> Is a batch runner. The job that runs the batch ensember.

So in option 2 you can turn off a feature, i.e. BatchEnsembling or Foo or Bar. This is useful because:
1. If we were to add more, we can toggle individually. i.e. turn off Batch Ensembling and not Foo batch feature, not a turn off all or turn on all runners.
2. Common interface for all runners -> users don't have to relearn a new pattern for the configuration.

romanwozniak: The main reason I'm not happy with option 2 is because I can't think of any "other" Foo or Bar runners in a foreseeable future, hence that's the premature optimization that I'd like to avoid

ashwinath: Alright I think that is fair. Option 1 it is.

ashwinath:
@romanwozniak Would require another review. You can see the full config here: https://github.com/ashwinath/turing/blob/add-toggle-on-batch-ensemblers/api/turing/config/example.yaml

There was one config bug I realised last week regarding the Taint/Tolerations which I have fixed in this MR as well. There was a TaintKey config and TolerationName config but only one of them worked. I kept TolerationName and fixed it in here: https://github.com/ashwinath/turing/blob/add-toggle-on-batch-ensemblers/api/turing/config/example.yaml#L105 so that this key will be used for spark jobs.

romanwozniak:
> @romanwozniak Would require another review. You can see the full config here: https://github.com/ashwinath/turing/blob/add-toggle-on-batch-ensemblers/api/turing/config/example.yaml
> 

This example still uses `BatchRunnerConfig` as a top-level config key

README.md
```diff
@@ -158,43 +158,46 @@ router Docker image is set to the image built and push from the previous step.
...
+        ExecutorReplica: 2
+        ExecutorCPURequest: "1"
+        ExecutorMemoryRequest: "1Gi"
+  RunnerConfig:
+    TimeInterval: 10s
+    RecordsToProcessInOneIteration: 10
+    MaxRetryCount: 3
+  ImageBuildingConfig:
+    ImageConfig:
+      Registry: ghcr.io
```
romanwozniak, line 180: Would be nice to reflect in the config key name, that this is the destination registry

romanwozniak (approved):
I've added some suggestions on the config structure and naming, but overall it lgtm

README.md
```diff
@@ -158,43 +158,46 @@ router Docker image is set to the image built and push from the previous step.
 
 ```yaml
 # config-dev.yaml
-BatchRunnerConfig:
-  TimeInterval: 10s
+BatchEnsemblerConfig:
```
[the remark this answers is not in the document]
ashwinath: Yeah let me fix this. Must have had a brainfart

README.md
```diff
@@ -158,43 +158,46 @@ router Docker image is set to the image built and push from the previous step.
...
+    RecordsToProcessInOneIteration: 10
+    MaxRetryCount: 3
+  ImageBuildingConfig:
+    ImageConfig:
+      Registry: ghcr.io
+      BaseImageRef: ghcr.io/gojek/turing/batch-ensembler:latest
+      BuildNamespace: default
+      BuildContextURI: git://github.com/gojek/turing.git#refs/heads/main
+      DockerfileFilePath: engines/batch-ensembler/app.Dockerfile
+      BuildTimeoutDuration: 20m
```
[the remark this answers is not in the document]
ashwinath: The reason I split it into different levels was to differentiate the buildee (result product) and the builder (tool).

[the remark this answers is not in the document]
ashwinath: Yeah actually I think your suggestion is a good one, I'll incorporate this.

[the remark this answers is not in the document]
ashwinath: It is passed to Kaniko but its not specific to Kaniko, but rather the built ensembler image, which is something i'm not too keen on.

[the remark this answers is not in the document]
ashwinath: Oh I see, ok i think its good, I'll remove the ImageConfig layer as well.

PR #74, opened 2021-07-08
Add validation and default ensembling job values.

ashwinath:
This PR adds the following:

1. Default Spark config annotations that is configurable in the Turing API config file.
2. Add in validation for the Create ensembling job POST body.
3. Nasty side effects include the rewiring of viper configuration delimiter from `.` to `::` due to a problem in unmarshalling strings.

For example:
```go
type Foo struct {
  Bar map[string]string
}
```

with a configuration of

```yaml
Foo:
  Bar:
    "hello.one.two": "some string"
```

Will yield this structure instead:
```foo
Foo = interface{}{
  "Bar": interface{}{
    "hello": interface{}{
      "one": inteface{}{
        "two": "some string",
      },
    },
  },
}
```

To solve this, we change the delimiter to something uncommon, say a cons operator `::`.

### DIFF
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -136,7 +136,7 @@ type EnsemblingJobConfig struct {
 	// ImageBuilderConfig contains the configuration related to the built ensembler image itself.
 	ImageBuilderConfig ImageBuilderConfig `validate:"required"`
 	// DefaultConfigurations contains the default configurations applied to the ensembling job.
-	// The user is free to override/append to the configuration
+	// The user (the person who calls the API) is free to override/append the default values.
 	DefaultConfigurations DefaultEnsemblingJobConfigurations `validate:"required"`
 }
 
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -8,8 +8,9 @@ import (
 	"time"
 
 	"github.com/go-playground/validator/v10"
-	"github.com/gojek/mlp/api/pkg/instrumentation/newrelic"
-	"github.com/gojek/mlp/api/pkg/instrumentation/sentry"
+	"github.com/gojek/mlp/pkg/instrumentation/newrelic"
+	"github.com/gojek/mlp/pkg/instrumentation/sentry"
+	"github.com/gojek/turing/api/turing/models"
 	"github.com/mitchellh/mapstructure"
 
 	// Using a maintained fork of https://github.com/spf13/viper mainly so that viper.AllSettings()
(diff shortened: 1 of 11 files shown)


krithika369:
"Nasty side effects ..."

I don't think it's nasty at all - it sounds like a common use case and a legit solution. 😄

api/turing/config/config.go
```diff
@@ -134,6 +135,17 @@ type EnsemblingJobConfig struct {
 	KanikoConfig KanikoConfig `validate:"required"`
 	// ImageBuilderConfig contains the configuration related to the built ensembler image itself.
 	ImageBuilderConfig ImageBuilderConfig `validate:"required"`
+	// DefaultConfigurations contains the default configurations applied to the ensembling job.
+	// The user is free to override/append to the configuration
```
krithika369, line 139: > The user is free to override/append to the configuration

By append/override, do we mean the user-supplied config values on app startup or some other way such as configs specified during job submission? I imagine it's the latter and if so, can it be clarified with a line or two in the comments?

api/turing/models/ensembling_job.go
```diff
@@ -76,11 +76,11 @@ func (r *InfraConfig) Scan(value interface{}) error {
 
 // BatchEnsemblingJobResources contains the resources delared to run the ensembling job.
 type BatchEnsemblingJobResources struct {
-	DriverCPURequest      string `json:"driver_cpu_request,omitempty"`
```
krithika369, line 79: Why has the `omitempty` been removed?

api/turing/models/ensembling_job.go
```diff
@@ -56,8 +56,8 @@ func (r *JobConfig) Scan(value interface{}) error {
 type InfraConfig struct {
 	ArtifactURI        string                       `json:"artifact_uri"`
 	EnsemblerName      string                       `json:"ensembler_name"`
-	ServiceAccountName string                       `json:"service_account_name"`
-	Resources          *BatchEnsemblingJobResources `json:"resources"`
+	ServiceAccountName string                       `json:"service_account_name" validate:"required"`
+	Resources          *BatchEnsemblingJobResources `json:"resources" validate:"required"`
```
krithika369, line 60: Why is this required but the fields within the `BatchEnsemblingJobResources` are all optional?

api/turing/models/ensembling_job.go
```diff
@@ -12,12 +12,12 @@ import (
 // EnsemblingJob holds the information required for an ensembling job to be done asynchronously
 type EnsemblingJob struct {
 	Model
-	Name            string       `json:"name"`
-	EnsemblerID     ID           `json:"ensembler_id"`
+	Name            string       `json:"name" validate:"required"`
+	EnsemblerID     ID           `json:"ensembler_id" validate:"required"`
 	ProjectID       ID           `json:"project_id"`
 	EnvironmentName string       `json:"environment_name"`
```
krithika369, line 18: If the environment is simply going to be taken from the Ensembler, does it need to be stored here?

api/turing/models/ensembling_job.go
ashwinath: I don't think there is a reason to do that anymore since the values are going to be populated

api/turing/models/ensembling_job.go
ashwinath: This is a mistake, will remove this

api/turing/models/ensembling_job.go
ashwinath: Not quite related to this PR but I do agree with you. I think it was there last time because I thought it was a user supplied value, but during the review process this was changed to a configuration value and this was left behind.

ashwinath: Ah seems like I missed out an important bit of information. There might be a chance next time where this option might be fillable by the user. It would be easier to state the default environment for now in the database as compared to doing a data patch in the future (which might be complicated because the default environment is pulled from MLP).

krithika369 (approved):
LGTM!

PR #72, opened 2021-07-07
Turing SDK: submit batch ensembling job

romanwozniak:
### Context:
This PR extends `turing-sdk` with a functionality of configuring and submitting batch ensembling jobs. 

### Changes:
- re-generated REST client from openapi spec (so, changes to [sdk/turing/generated](https://github.com/romanwozniak/turing/tree/sdk_submit_job/sdk/turing/generated) can be safely skipped during review)
- added necessary classes related to batch ensembling under [sdk/turing/batch](https://github.com/romanwozniak/turing/tree/sdk_submit_job/sdk/turing/batch)
- added e2e example on how to submit a job at [sdk/samples/batch_job](https://github.com/romanwozniak/turing/tree/sdk_submit_job/sdk/samples/batch_job)
- some minor changes to [proto](https://github.com/romanwozniak/turing/blob/sdk_submit_job/engines/batch-ensembler/api/proto/v1/batch_ensembling_job.proto) definition and [openapi](https://github.com/romanwozniak/turing/blob/sdk_submit_job/api/openapi/jobs.yaml) spec to make them compatible with each other

### TODO
 - [x] add unit tests where missed

PR #75, opened 2021-07-13
Turing SDK: terminate ensembling job

romanwozniak:
### Context:
This PR builds up on the [previous PR](https://github.com/gojek/turing/pull/72) and adds the way to terminate a running job from the SDK

### Changes:
- Added `.terminate()` and `.refresh()` methods into `EnsemblingJob`class
- Added minor fixes into turing-api and openapi spec
- Extended [batch_job.py](https://github.com/gojek/turing/compare/main...romanwozniak:sdk_terminate_ensembling_job?expand=1#diff-24411c64b2ce258f7ecccd88cb035d695dbc0e652fba7025ddc156d26117a620) example with a demonstration on how to terminate the job
- Unit tests

### DIFF
--- a/sdk/requirements.txt
+++ b/sdk/requirements.txt
@@ -1,8 +1,9 @@
+fire
 google-auth>=1.30.0,<2.0dev
 google-cloud-storage>=1.37.0
 mlflow==1.14.1
 numpy==1.17.3
 pandas==1.2.2
 python-dateutil
+requests==2.25.1
 urllib3 >= 1.25.3
-fire
--- a/sdk/turing/_base_types.py
+++ b/sdk/turing/_base_types.py
@@ -23,6 +23,11 @@ class DataObject:
 
         return dict(attribs)
 
+    def __eq__(self, other):
+        if isinstance(self, other.__class__):
+            return self.to_dict() == other.to_dict()
+        return False
+
     def __str__(self):
         return '%s(%s)' % (
             type(self).__name__,
--- a/sdk/turing/_base_types.py
+++ b/sdk/turing/_base_types.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from turing.generated.model_utils import ModelNormal
+from turing.generated.model_utils import OpenApiModel
 
 
 class DataObject:
@@ -60,7 +60,7 @@ class ApiObject(DataObject):
         return self._updated_at
 
     @classmethod
-    def from_open_api(cls, open_api: ModelNormal):
+    def from_open_api(cls, open_api: OpenApiModel):
         """
         Factory method, for constructing ApiObject (and its sub-classes) instances
         from a relevant instance of openapi-generator generated model instance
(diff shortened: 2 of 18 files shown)


sdk/requirements.txt
```diff
@@ -1,8 +1,9 @@
+fire
 google-auth>=1.30.0,<2.0dev
 google-cloud-storage>=1.37.0
 mlflow==1.14.1
 numpy==1.17.3
 pandas==1.2.2
 python-dateutil
+requests==2.25.1
```
romanwozniak, line 8: The latest version (`2.26.0`) is not working correctly with `urllib3-mock`, so I have to pin the version explicitely

sdk/turing/batch/job.py
```diff
@@ -86,6 +87,36 @@ def status(self) -> 'EnsemblingJobStatus':
     def error(self) -> str:
         return self._error
 
+    def refresh(self):
+        """
+        Fetches latest updates of this ensembling job
+        """
+        from turing.session import active_session
```
krithika369, line 94: I probably missed this in the earlier reviews. I generally prefer having imports at the top level, so it is obvious what a module depends on. I've only seen this style of writing imports inside functions for optional dependencies which makes sense. Is there another reason why the active session is being imported in every function instead of globally ?

If it is because the `active_session` may not be set initially, I think it can be overcome in one of the following ways:
* `import turing.session` -> then use `turing.session.active_session` elsewhere
* Create a getter function in `turing.session` that can be imported and be responsible for returning the session object: `active_session().get_ensembling_job(...`

wdyt?

krithika369 (approved):

krithika369:
Couple of minor questions / comments. The rest LGTM. Thanks!

sdk/turing/_base_types.py
```diff
@@ -23,6 +23,11 @@ def to_dict(self):
 
         return dict(attribs)
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
romanwozniak: > Since this has been implemented, do we still need the tests to use obj1.to_dict() == obj2.to_dict() ?

In `ensembler_test.py` the comparison happens between a wrapper class entity (`turing.PyfuncEnsembler`) and generated class entity (`turing.generated.models.PyfuncEnsembler`) hence these objects are not equal. The other way of dealing with it would be `assert obj1 == ClassName.from_open_api(obj2)`, is it less confusing? 

> And is comparing only the public fields

Yes, I'd say that it's sufficient for now. And if we care about the private fields, then they should be made public

sdk/turing/batch/job.py
romanwozniak: iiric, local imports were added to avoid/solve cyclic imports in the codebase. But now I addressed it here: https://github.com/gojek/turing/pull/75/commits/d559dd0fcc36c25452c037c2d27b7650cf891705

PR #78, opened 2021-07-28
Bump versions of mlp/merlin dependency

romanwozniak:
#### Context 
As per title. Both dependencies were set to the initial open-source version of merlin/mlp. Update the repo to use more recent versions of these dependencies.

#### Extra:
Fixed `GET /projects` api endpoint by passing the user-email into authZ enforcer.


api/turing/api/projects_api.go
```diff
@@ -22,7 +22,7 @@ func (c ProjectsController) ListProjects(
 	}
 
 	if c.Authorizer != nil {
-		user, _ := vars.get("user")
+		user := r.Header.Get("User-Email")
```
romanwozniak, line 25: User email is populated by the auth-proxy, if it's enabled

api/turing/middleware/authorization.go
```diff
@@ -85,7 +85,7 @@ func (a *Authorizer) FilterAuthorizedProjects(
 		return nil, err
 	}
 
-	var allowedProjects []mlp.Project
+	allowedProjects := make([]mlp.Project, 0)
```
romanwozniak, line 88: Otherwise `allowedProjects` could be `nil`, if  `allowedProjectIds` is empty

ashwinath: They are actually equavilent. `make([]mlp.Project, 0)` == `var allowedProjects []mlp.Project`. Can try here: https://play.golang.org/p/BZYWfhQnXO4

romanwozniak: They aren't. Check nil slice vs empty slice in Golang, for example: https://medium.com/@habibridho/golang-nil-vs-empty-slice-87fd51c0a4d

romanwozniak: What's important, that: 
```go
fmt.Println(nilSlice == nil) // Output: true
fmt.Println(emptySlice == nil) // Output: false
```
So when you try to marshal empty slice into JSON, you'll get `[]`, but marshalling of nil slice will end up with `null`

ashwinath (approved):

api/turing/middleware/authorization.go
ashwinath: Ah interesting.

PR #80, opened 2021-07-30
UI: List ensembling jobs

romanwozniak:
### Context:
This PR adds a simple view to list ensembling jobs within a project. It is also supplemented with https://github.com/gojek/mlp/pull/30, so users can easily navigate between `Routers` and `Jobs` sections.

### Screenshots:
Screenshot 2021-07-30 at 18 48 17
Screenshot 2021-07-30 at 18 48 30

### DIFF
--- a/ui/src/job/list/ListEnsemblingJobsView.js
+++ b/ui/src/job/list/ListEnsemblingJobsView.js
@@ -1,0 +1,100 @@
+import { useTuringApi } from "../../hooks/useTuringApi";
+import {
+  EuiButton,
+  EuiPage,
+  EuiPageBody,
+  EuiPageContent,
+  EuiPageHeader,
+  EuiPageHeaderSection
+} from "@elastic/eui";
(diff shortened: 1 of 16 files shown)


ui/src/job/list/ListEnsemblingJobsView.js
```diff
@@ -0,0 +1,100 @@
...
+      });
+    }
+  }, [data, isLoaded, error]);
+
+  useEffect(() => {
+    replaceBreadcrumbs([{ text: "Jobs" }]);
+  }, []);
+
+  const onRowClick = item => {};
+  // props.navigate(`./${item.id}/details`);
```
ashwinath, line 69: Should probably be removed

romanwozniak: Nope, I just temporarily commented it out because job details page is not available yet

ashwinath (approved):

PR #81, opened 2021-08-03
UI: Add view to show ensemblers

romanwozniak:
### Context
This PR makes it possible to view a list of ensemblers in Turing UI. Plus:
* Makes it possible to view ensembling jobs of a specific ensembler
* Updates jobs api to support `ensembler_id` and `search` query parameters
* Bumps version of `@mlp/ui` library to use the newer version of `NavDrawer`

### Screenshots:
Screenshot 2021-08-03 at 23 57 48
Screenshot 2021-08-03 at 23 57 36

### DIFF
--- a/api/turing/service/ensembling_job_service.go
+++ b/api/turing/service/ensembling_job_service.go
@@ -23,7 +23,9 @@ type EnsemblingJobFindByIDOptions struct {
 type EnsemblingJobListOptions struct {
 	PaginationOptions
 	ProjectID          *models.ID      `schema:"project_id" validate:"required"`
+	EnsemblerID        *models.ID      `schema:"ensembler_id"`
 	Statuses           []models.Status `schema:"status"`
+	Search             *string         `schema:"search"`
 	RetryCountLessThan *int            `schema:"-"`
 	UpdatedAtBefore    *time.Time      `schema:"-"`
 }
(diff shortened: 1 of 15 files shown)


api/turing/service/ensembling_job_service.go
```diff
@@ -100,6 +102,14 @@ func (s *ensemblingJobService) List(options EnsemblingJobListOptions) (*Paginate
 		query = query.Where("project_id = ?", options.ProjectID)
 	}
 
+	if options.EnsemblerID != nil {
+		query = query.Where("ensembler_id = ?", options.EnsemblerID)
+	}
+
+	if options.Search != nil && len(*options.Search) > 0 {
+		query = query.Where("name like ?", fmt.Sprintf("%%%s%%", *options.Search))
```
ashwinath, line 110: This probably requires indexing of the database on the `name` field.

krithika369 (approved):

ui/src/ensembler/list/ListEnsemblersTable.js
```diff
@@ -0,0 +1,187 @@
...
+  onRowClick,
+  ...props
+}) => {
+  const searchQuery = useMemo(() => {
+    const parts = [];
+    if (!!filter.search) {
+      parts.push(filter.search);
+    }
+    if (!!filter.type) {
+      parts.push(`type:${filter.type}`);
```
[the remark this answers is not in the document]
romanwozniak: 1. The motivation is simple: if the API supports pagination, then you need to support filtering at the back-end as well, because otherwise your search will only be limited to the results shown on a single page (which is not how I imagine search should work). 
Routers API doesn't support pagination (i.e. all results are fetched at once and then pagination is only added on the client-side for the users convenience), that's why `EuiInMemoryTable` is used there with a client-side search/filtering. 

I don't think that back-end pagination matters that much for the Ensemblers View (I won't expect that user will have more than handful of ensemblers inside a single project), but I expect that it's definitely required for the Jobs View, hence I decided to standardize the implementation.

2. Do you mean some search-through functionality, where user can find different entities (router, ensembler, experiment etc,) in a single search result? I don't think that it's important for now.

I think that filtering (i.e. filter by status or by ensembler id) is more important than search, but I've added the later just in case.

api/turing/service/ensembling_job_service.go
romanwozniak: I prefer to get back to this when it cause us performance issues

PR #82, opened 2021-08-10
Migrate UI to yarn package manager

romanwozniak:
Recently, CI pipeline started failing because of the "raw" dependency that we use (`github:gojekfarm/react-lazylog`). This raw dependency should be "built" (transpiled, to be precise) from sources before it can be installed into the dependent project (in our case, into turing-ui).  `react-lazylog` uses a bunch of outdated dependencies, and one of them started causing this issue:
```js
Missing file extension "mjs" for "@mattiasbuelens/web-streams-polyfill/ponyfill" (import/extensions) at src/stream.js:8:14:
```

Without much of details, `yarn` handles the build process well. I wanted to migrate turning UI from `npm` to `yarn` anyway, so this PR does just this.

Extra: I also bumped minor versions of some of the dependencies just to keep them up to date.


ui/package.json
```diff
@@ -6,18 +6,18 @@
   "dependencies": {
     "@elastic/datemath": "5.0.3",
     "@elastic/eui": "32.3.0",
-    "@gojek/mlp-ui": "1.4.9",
-    "@reach/router": "1.3.3",
+    "@gojek/mlp-ui": "npm:@romanwozniak/mlp-ui@1.4.9",
```
romanwozniak, line 9: Will update this once https://github.com/gojek/mlp/pull/32 is merged

krithika369 (approved):
> Without much of details, yarn handles the build process well.

Can it be with much of details? :)

Curious how the resolution is handled differently for this specific failure. The [NPM package](https://www.npmjs.com/package/@mattiasbuelens/web-streams-polyfill) says it's been moved to `web-streams-polyfill@2.0.0`. Is yarn still able to fetch the deprecated package from a different registry?

ashwinath (approved):

PR #83, opened 2021-08-11
Fix Swagger UI in Helm chart.

ashwinath:
Previously, the refactoring of swagger yaml files have been split up and the process of how the yaml files were transferred to the SwaggerUI running container was not updated. As a result, SwaggerUI is not able to serve the openapi specs correctly.

This PR changes how SwaggerUI is built and deployed. A custom image will be built on top of SwaggerUI's Docker image to populate the openapi yaml files and deployed alongside with the Turing API pod.

Test for Docker build is here: https://github.com/gojek/turing/runs/3308105094?check_suite_focus=true, temporarily removed the guard for main branch only, will be reenabling that. I do not have the credentials so the pushing to GHCR will fail.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -60,6 +60,7 @@ jobs:
           - "install_merlin" install_merlin \
           - "build_turing_router_docker_image" build_turing_router_docker_image \
           - "build_turing_apiserver_docker_image" build_turing_apiserver_docker_image \
+          - "build_turing_swagger_docker_image" build_turing_swagger_docker_image \
           --and-then \
           - "install_turing" install_turing
 
--- a/.github/workflows/swagger.yaml
+++ b/.github/workflows/swagger.yaml
@@ -29,4 +29,4 @@ jobs:
         run: echo ${{ secrets.GHCR_TOKEN }} | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
 
       - name: Publish image
-        run: docker push ${{ env.DOCKER_REGISTRY }}/turing/batch-ensembler:$(make version)
+        run: docker push ${{ env.DOCKER_REGISTRY }}/turing/turing-swagger:$(make version)
--- a/.github/workflows/swagger.yaml
+++ b/.github/workflows/swagger.yaml
@@ -1,0 +1,32 @@
+name: swagger
+
+on:
+  push:
+    paths:
+    - ".github/workflows/swagger.yaml"
+    - "api/api/**"
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  publish-docker:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./api/api
+    env:
+      DOCKER_REGISTRY: ghcr.io/gojek
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Build image
+        run: make build-image
+
+      - name: Setup GitHub Container Registry
+        run: echo ${{ secrets.GHCR_TOKEN }} | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Publish image
+        run: docker push ${{ env.DOCKER_REGISTRY }}/turing/batch-ensembler:$(make version)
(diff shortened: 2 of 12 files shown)


infra/chart/values.yaml
```diff
@@ -102,8 +102,13 @@ dbMigrations:
     tag: v4.7.1
 
 swaggerUi:
-  # -- Docker tag for Swagger UI https://hub.docker.com/r/swaggerapi/swagger-ui
   image:
+    # -- Docker registry for Turing Swagger UI image. User is required to override
+    # the registry for now as there is no publicly available Turing Swagger UI image
```
romanwozniak, line 107: Can you include a GH workflow (or a step in existing turing-api workflow) and publish image to ghcr.io?

api/turing/cmd/main.go
```diff
@@ -122,10 +122,6 @@ func main() {
 	mux.Handle("/v1/internal/", http.StripPrefix("/v1/internal", health))
 	mux.Handle("/v1/", http.StripPrefix("/v1", api.NewRouter(appCtx)))
 
-	if len(cfg.SwaggerFile) > 0 {
-		// Serve Swagger Spec
-		mux.Handle("/openapi.yaml", web.FileHandler(cfg.SwaggerFile, false))
```
romanwozniak, line 127: Do we use `web.FileHandler` anywhere else? If no, shall we drop it?

romanwozniak (commented):
The CI part's missing, the rest lgtm

api/turing/cmd/main.go
ashwinath: It was used to pass the swagger.yaml file over to the Swaggerui container. So I'm dropping it.

api/api/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM swaggerapi/swagger-ui:v3.47.1
```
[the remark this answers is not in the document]
ashwinath: Was a little conflicted on this. I'll change this to the latest along with the docker-compose/makefile script

infra/chart/values.yaml
ashwinath: Turing also?

api/turing/cmd/main.go
ashwinath: I see what you mean. It's used to serve the react files as well.

infra/chart/values.yaml
ashwinath: I've done just for Swagger only.

.github/workflows/e2e.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
           - "install_merlin" install_merlin \
           - "build_turing_router_docker_image" build_turing_router_docker_image \
           - "build_turing_apiserver_docker_image" build_turing_apiserver_docker_image \
+          - "build_turing_swagger_docker_image" build_turing_swagger_docker_image \
```
romanwozniak, line 63: Do we need it here? Can we for e2e test just use `swagger/swagger-ui:latest` or smth?

.github/workflows/swagger.yaml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/checkout@v2
+
+      - name: Build image
+        run: make build-image
+
+      - name: Setup GitHub Container Registry
+        run: echo ${{ secrets.GHCR_TOKEN }} | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Publish image
+        run: docker push ${{ env.DOCKER_REGISTRY }}/turing/batch-ensembler:$(make version)
```
romanwozniak, line 32: Wrong image name

infra/chart/values.yaml
```diff
@@ -102,8 +102,13 @@ dbMigrations:
...
+    registry: docker.io/
```
romanwozniak, line 108: Now you can update the comments and values here

.github/workflows/e2e.yaml
ashwinath: It doesn't take long anyway but we should follow the full flow as much as possible.

infra/chart/values.yaml
ashwinath: done

.github/workflows/swagger.yaml
ashwinath: fixed.

romanwozniak (approved):
lgtm

PR #85, opened 2021-08-17
Fix Swagger UI base URL.

ashwinath:
The `sed` script was not updating the values properly after the refactoring of the openapi yamls. This PR fixes it.

Tested by deploying manually: 
Screenshot 2021-08-17 at 6 18 55 PM

### DIFF
--- a/infra/chart/templates/deployment.yaml
+++ b/infra/chart/templates/deployment.yaml
@@ -120,7 +120,7 @@ spec:
         args:
         - |
           echo "Update Swagger config..."
-          sed -r -i 's%^((\s*)-(\s*)url\s*:).*$$%\1 "'$${API_SERVER}'"%' $${SWAGGER_JSON}
+          sed -i "s|localhost:8080/v1|$${API_SERVER}|g" $${SWAGGER_JSON}
           echo "Running Swagger UI..."
           /usr/share/nginx/run.sh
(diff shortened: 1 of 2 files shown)


infra/chart/templates/deployment.yaml
```diff
@@ -120,7 +120,7 @@ spec:
         args:
         - |
           echo "Update Swagger config..."
-          sed -r -i 's%^((\s*)-(\s*)url\s*:).*$$%\1 "'$${API_SERVER}'"%' $${SWAGGER_JSON}
+          sed -i "s|localhost:8080/v1|$${API_SERVER}|g" $${SWAGGER_JSON}
```
romanwozniak, line 123: This will replace any occurrence of `localhost:8080/v1` which may cause some unexpected side effects (the chances are low, but not zero). Can you please fix the original `sed` statement, so only `- url: <host>` is replaced with `- url: ${API_SERVER}`? 
The previous working `sed` statement is here: https://github.com/gojek/turing/commit/cd4e94cc5c8b0d63bc01bf447002718b7fec44ff#diff-b5dc8008be0c02999a825072bd09d9c8e33efd2d58bd3f7c17b7903a09309917

ashwinath:
Closing this PR as it was an issue with our configuration input rather than the chart itself.

PR #87, opened 2021-08-23
Add dashboard URL for ensembling jobs.

ashwinath:
This PR adds the monitoring dashboard URL for the ensembling job. All ensembling jobs have the same URL so there isn't a need to persist into the database, so it is populated at runtime and sent back during the API call.

Affected APIs:
- create ensembling job
- list ensembling jobs
- get ensembling job


api/api/specs/jobs.yaml
```diff
@@ -160,6 +160,8 @@ components:
           $ref: "#/components/schemas/EnsemblerJobStatus"
         error:
           type: "string"
+        dashboard_url:
```
krithika369, line 163: This `EnsemblingJob` is also used as the request body for the create job request. But we don't expect the dashboard URL to be input (the same problem also exists with the `id` field and perhaps others in the create request).

I think we can define an explicit type to handle the create request. Eg:

```
EnsemblingJob:
      allOf:
        - $ref: '#/components/schemas/EnsemblingJobCreateRequestBody'
        - type: object
          properties:
            id:
              $ref: "#/components/schemas/Id"
            dashboard_url:
              type: string
```

wdyt ?

api/turing/service/ensembling_job_service.go
```diff
@@ -180,6 +188,7 @@ func (s *ensemblingJobService) CreateEnsemblingJob(
 	job.JobConfig.Spec.Ensembler.Uri = getEnsemblerDirectory(ensembler)
 	job.InfraConfig.ArtifactURI = ensembler.ArtifactURI
 	job.InfraConfig.EnsemblerName = ensembler.Name
+	job.DashboardURL = s.dashboardURL
```
krithika369, line 191: I would've expected that this URL is formatted with the project / namespace / job info for the specific job. Do you have an example of how and where this URL is expected to be used?

For the real time routers, we set the templated URL in the UI config which is populated according to the router being viewed. We've had some discussions in the past, for the real-time router dashboard URL to be moved to the API layer as well (see task id 634 in our backlog), so it may also be available to the SDK. So, I agree with adding the ensembling job's dashboard URL to the API. My concern is just with returning the same value for every job - that feels incomplete to me.

api/api/specs/jobs.yaml
ashwinath: Seems like this problem was solved using the `readOnly: true`, will do that instead. example: https://github.com/ashwinath/turing/blob/f2ea45af2f49114aff304be2585ba6095fb8d3bd/api/api/specs/routers.yaml#L414

krithika369: Nice. TIL. :)

api/turing/service/ensembling_job_service.go
ashwinath: This makes sense, I think this should be implemented, feels rather weird when I first did it as well. Let me add this.

ashwinath: @krithika369 I have made the changes to allow templating.

api/db-migrations/000009_ensembling_job.up.sql
```diff
@@ -22,6 +22,7 @@ CREATE TABLE IF NOT EXISTS ensembling_jobs
     retry_count      integer NOT NULL default 0,
     status           ensembling_job_status NOT NULL default 'pending',
     error            text,
+    dashboard_url    text,
```
krithika369, line 25: Do we need to persist this in the DB now? In a scenario where the dashboard URL is modified (although rare), we will have to update the data in the DB. For this reason, I think it is better to format the template string dynamically.

api/turing/config/example.yaml
```diff
@@ -14,6 +14,8 @@ AuthConfig:
 # Batch ensembler job configurations
 BatchEnsemblingConfig:
   Enabled: true
+  # Note that {{.Project}} and {{.Job}} is required.
```
krithika369, line 17: Can we add more details here on what templating vars are supported exactly (is it only these 2?) and what they will be replaced with? Like `{{.Project}}` and `{{.Job}}` for project name and job name, respectively.

api/turing/service/ensembling_job_service.go
```diff
@@ -17,7 +21,8 @@ const (
 	SparkHomeFolder = "/home/spark"
 	// EnsemblerFolder is the folder created by the Turing SDK that contains
 	// the ensembler dependencies and pickled Python files.
-	EnsemblerFolder = "ensembler"
+	EnsemblerFolder  = "ensembler"
+	jobNameMaxLength = 20
```
krithika369, line 25: Ohh why this limit? I'm afraid having this constraint is not very generalisable. Is it a limitation from the internal dashboard?

api/db-migrations/000009_ensembling_job.up.sql
ashwinath: I thought about it, it's quite troublesome because a few endpoints are affected:
1. create ensembling job
2. list ensembling job
3. get ensembling job

You might get a runtime performance hit (maybe negligible) when you call the list ensembling job method, since you have to construct the url for every job. I picked the persistence method because less code was required.

ashwinath: But if you feel strongly about this we can always change it to a runtime thing. I think the data update might be relevant but ensembling jobs are a temporary thing, the dashboard wouldn't be very useful later on, i doubt it will be worth it to even update the values.

api/turing/service/ensembling_job_service.go
ashwinath: It's a limit because of how spark truncates the pod name, I don't remember the exact value because I used trial and error, the value could be more but 20 seems safe. Are you suggesting hardcoding it?

api/turing/config/example.yaml
ashwinath: Yep ok

api/db-migrations/000009_ensembling_job.up.sql
ashwinath: actually after thinking about it, I'll change it to generate during runtime. so that its always independent of config.

api/turing/service/ensembling_job_service.go
ashwinath: I think there are certain implications that were overlooked here as discussed with @krithika369 offline.

This setting is reversed engineering from how spark generates the pod names. Generally speaking, there are a few components at play here and I will try my best to explain what is generated in terms of order.

1. Ensembling job name provided by the user is appended with the unix timestamp appended to the back. This is currently limited to 50 characters excluding the timestamp (60 included).
2. SparkApplication (CRD) is created with the name equivalent to ensembling job name which has a maximum of 60 characters.
3. Spark Driver Pod is created with the Pod name of a format roughly equivalent to this: `<Ensembling Job Name truncated to some length>-<generated hash>-driver`
4. Spark Executor Pod is created with the Pod name of a format roughly equivalent to this: `<Ensembling Job Name truncated to some length>-<generated hash>-exec-<replica number>`

So the dashboard works by using the metric provided by kube-state-metrics: `kube_pod_container_status_ready` and other relevant metric names by filtering it based on `pod=~"$job.*-driver|$job.*-exec-.*"`. According to the [kube-state-metrics documentation](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md) there seems to be no filter based on a pod annotation/label so this becomes extremely awkward and the reason why I've decided to do a string operation to extract the value out of it.

Now the problem is how are we expecting our users to name their ensembling jobs? @krithika369 probably suspects that they are going to create a name based on a fixed portion appended by some generated string; i.e. in the format `<fixed name>-<generated>` which might be a problem if we truncate to such short character sets of length 20.

My proposal is to limit the number of characters in the CreateEnsemblingJob process to 20 characters. This probably makes things a little tough. If we append 10 characters for the timestamp, users are basically left with 10 characters for their ensembling job name. A few approaches I can think of:

1. Recommend them to use a name that is short and tell them there will be a timestamp generated at the end of it, limiting them to 10 characters only. 
2. Leave the responsibility of letting them generate a name themselves.

In both circumstances, we would have to enforce a unique index on the name + mlp project or there might be collisions in naming.

Honestly speaking, I'm not sure what should be the right way to approach this problem. Maybe @romanwozniak you could weigh in on this.

krithika369: > My proposal is to limit the number of characters in the CreateEnsemblingJob process to 20 characters

After thinking about this a bit and discussing with @ashwinath offline, limiting the Ensembler name length sounds like a reasonable approach to me (for comparison, we set a 25 chart limit for [router names](https://github.com/gojek/turing/blob/main/ui/src/router/components/form/validation/schema.js#L214) which is probably the reason why we don't have to worry about any truncation of the k8s resource name later). We could put it the same constraint for ensembler names. And similar to the router version number, we can introduce the concept of a `run-id` in the jobs table that is unique per `(MLP Project ID, Ensembler ID)` so that the final name of the job (ensembler name + run ID, instead of name + timestamp) is still < 30 characters and allows enough room for the other suffixes appended by the Spark op on its resources.

> In both circumstances, we would have to enforce a unique index on the name + mlp project or there might be collisions in naming.

Makes sense to me. We do a unique constraint on (MLP Project ID, Router Name). I think the same logic would work for Ensemblers - (MLP Project ID, Ensembler Name).

ashwinath: This also affects the ensembler name as we create the ensembling job name based on the ensembler name.

api/turing/service/ensembling_job_service.go
romanwozniak: Can we also use `DashboardURL` -> `MonitoringURL` to align it with Merlin ([here](https://github.com/gojek/merlin/blob/805f33fc67ab17f23a85b1b89283b2f12d100fbf/api/models/version_endpoint.go#L38))?

api/turing/service/ensembling_job_service.go
romanwozniak: RunID approach, proposed by @krithika369 makes sense to me. So then we can limit ensembler's name length to 25 char (to make it consistent with routers) and append sequential run id. But propose to implement it in a separate PR

api/turing/api/ensembling_job_api.go
```diff
@@ -50,7 +50,7 @@ func (c EnsemblingJobController) Create(
 		return BadRequest("only pyfunc ensemblers allowed", fmt.Sprintf("ensembler type given: %T", v))
 	}
 
-	ensemblingJob, err := c.EnsemblingJobService.CreateEnsemblingJob(job, projectID, pyFuncEnsembler)
+	ensemblingJob, err := c.EnsemblingJobService.CreateEnsemblingJob(job, project, pyFuncEnsembler)
```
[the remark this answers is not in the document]
ashwinath: Hmm this makes things quite awkward. From what I understand from the docs, the implementation would look something like this:

```go
func (j *EnsemblingJob) AfterFind() error {
  proj, err := MLPServiceSingleton.GetProject(j.ProjectID) // <------- this is where it gets awkward, because we need the project name
  if err != nil {
    return err
  }
  if err := s.populateMonitoringURL(job, project); err != nil {
    return err
  }
  return nil
}
```

Since it doesn't make sense to instantiate a service inside a model, a MLPService singleton has to be used.

The other way is to store the project name inside ensembling job which probably makes the most sense. Do you know if there are assumptions that MLP project name will always remain the same? Or can a project be renamed?

api/turing/batch/ensembling/runner.go
```diff
@@ -122,7 +122,7 @@ func (r *ensemblingJobRunner) updateStatus() {
 }
 
 func (r *ensemblingJobRunner) processEnsemblingJobs(queryOptions service.EnsemblingJobListOptions) {
-	ensemblingJobsPaginated, err := r.ensemblingJobService.List(queryOptions)
+	ensemblingJobsPaginated, err := r.ensemblingJobService.List(queryOptions, nil)
```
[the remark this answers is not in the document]
ashwinath: Yeah in hindsight this seems a little awkward. But I think this is related to https://github.com/gojek/turing/pull/87#discussion_r699061831 as that 2nd argument could just be removed completely.

api/turing/service/ensembling_job_service.go
```diff
@@ -181,11 +243,14 @@ func (s *ensemblingJobService) CreateEnsemblingJob(
 	job.InfraConfig.ArtifactURI = ensembler.ArtifactURI
 	job.InfraConfig.EnsemblerName = ensembler.Name
 
+	if err := s.populateMonitoringURL(job, project); err != nil {
```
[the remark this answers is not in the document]
ashwinath: Sure.

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
romanwozniak: Currently project can't be renamed, but I wouldn't rely on this assumption going forward. Yeah, having a singleton service would be weird. Okay, ignore this comment then

api/turing/batch/ensembling/runner.go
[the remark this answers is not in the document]
romanwozniak: I commented on https://github.com/gojek/turing/pull/87#discussion_r699061831, would be nice to remove the second argument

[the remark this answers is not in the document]
ashwinath: Alright, will fetch project again inside EnsemblingJobService again.

api/turing/api/ensembling_job_api.go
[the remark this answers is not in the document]
ashwinath: Ok but I will revert this to take in projectID instead of project instead.

api/turing/batch/ensembling/runner.go
```diff
@@ -350,7 +350,7 @@ func (r *ensemblingJobRunner) terminateJob(ensemblingJob *models.EnsemblingJob,
 
 // terminateIfRequired returns true if the process should drop what it is doing.
 func (r *ensemblingJobRunner) terminateIfRequired(ensemblingJobID models.ID, mlpProject *mlp.Project) bool {
-	ensemblingJob, err := r.ensemblingJobService.FindByID(ensemblingJobID, service.EnsemblingJobFindByIDOptions{})
+	ensemblingJob, err := r.ensemblingJobService.FindByID(ensemblingJobID, service.EnsemblingJobFindByIDOptions{}, nil)
```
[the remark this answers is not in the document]
ashwinath: I have removed this

api/turing/batch/ensembling/runner_test.go
```diff
@@ -84,7 +86,7 @@ func TestRun(t *testing.T) {
 					},
 				}, nil)
 
-				svc.On("FindByID", mock.Anything, mock.Anything).Return(
+				svc.On("FindByID", mock.Anything, mock.Anything, mock.Anything).Return(
```
romanwozniak, line 89: Did you forget to update this mocks after reverting the signature of methods on `EnsemblingJobService`?

api/turing/service/ensembling_job_service.go
```diff
@@ -95,6 +113,19 @@ func (s *ensemblingJobService) FindByID(
 		return nil, err
 	}
 
+	if options.ProjectID != nil {
+		project, err := s.mlpService.GetProject(*options.ProjectID)
```
romanwozniak, line 117: Doesn't it make more sense to use `ensemblingJob.ProjectID` instead?

api/turing/service/ensembling_job_service.go
```diff
@@ -142,6 +173,21 @@ func (s *ensemblingJobService) List(options EnsemblingJobListOptions) (*Paginate
 		return nil, err
 	}
 
+	if options.ProjectID != nil {
+		for _, r := range results {
+			project, err := s.mlpService.GetProject(*options.ProjectID)
```
romanwozniak, line 178: Same here. Why do you need to use `options.ProjectID` instead `r.ProjectID`?

api/turing/service/ensembling_job_service.go
```diff
@@ -181,11 +255,21 @@ func (s *ensemblingJobService) CreateEnsemblingJob(
...
+	project, err := s.mlpService.GetProject(projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	url, err := s.generateMonitoringURL(job, project)
+	if err != nil {
+		return nil, err
+	}
+	job.MonitoringURL = url
```
romanwozniak, line 267: `MonitoringURL` is not persisted in the database, right? Is the value remain set after `Save(job)` is called?

romanwozniak:
> > I'm done with my review, left some comments
> 
> Ok I have addressed them @romanwozniak

Conceptually it LGTM, but there are some minor fixes that I commented about to be added before merging this.

api/turing/batch/ensembling/runner_test.go
ashwinath: hmm why didn't the test flag this out, let me check

ashwinath: This is not very intuitive but I think I have found out the reason why it does not give a runtime error when more arguments are supplied (not less).

Seems to have an [issue raised with the library here](https://github.com/stretchr/testify/issues/587) and it seems like it is a gotcha with the library when `mock.Anything` is used. The arity of the function is not considered when using svc.On(funcName, mock.Anything).

api/turing/service/ensembling_job_service.go
ashwinath: It's not persisted but let's just not take the chance that the value might be wiped, so I'll put it after persisting.

api/turing/service/ensembling_job_service.go
ashwinath: Ok changed it.

ashwinath:
> > > I'm done with my review, left some comments
> > 
> > 
> > Ok I have addressed them @romanwozniak
> 
> Conceptually it LGTM, but there are some minor fixes that I commented about to be added before merging this.

Updated changes.

api/turing/api/ensembling_job_api_it_test.go
```diff
@@ -384,7 +384,7 @@ func TestIntegrationEnsemblingJobController_ListEnsemblingJob(t *testing.T) {
 			),
 			ensemblingJobService: func() service.EnsemblingJobService {
 				svc := &mocks.EnsemblingJobService{}
-				svc.On("List", mock.Anything, mock.Anything).Return(
+				svc.On("List", mock.Anything, mock.Anything, mock.Anything).Return(
```
romanwozniak, line 387: Update the mock here as well?

romanwozniak (approved):
LGTM, except this comment: https://github.com/gojek/turing/pull/87#discussion_r700860843. Good to go once it's fixed.

api/turing/api/ensembling_job_api_it_test.go
ashwinath: Yeah updated the rest of this file as well. Thanks for the catch.

PR #88, opened 2021-08-27
Add API to query ensembling job pod logs

ashwinath:
This PR adds the `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs` endpoint to query ensembling job logs. There are three kinds of logs that can be pulled from this endpoint:

1. `component_type=image_builder` -> Fetches logs from the Kaniko builder for the ensembler image container.
2. `component_type=driver` -> Fetches logs from the Spark driver that is running the ensembler.
3. `component_type=executor` -> Fetches logs from the Spark executors that is running the ensembler.

This assumes that the pod is still within the Kubernetes environment. It is known that the Kubernetes Spark driver pod will be deleted once it has run its course. However, you are able to put a logging URL template in `BatchEnsemblingConfig.LoggingURLFormat` to generate the URL to the persisted logs. You could put a stackdriver URL with the following variables `{{.PodName}}` and `{{.Namespace}}` and Turing API will automatically fill these dynamic values. For example: `https://www.example.com/{{.Namespace}}/{{.PodName}}` will generate the logs URL.

There are a couple of changes from the router logs endpoint. Here's a comparison:

Router log style:
```json
[
  {
    "timestamp": "2021-08-27T06:58:57.05303975Z",
    "environment": "dev",
    "namespace": "image-builder",
    "pod_name": "batch-1630047384-3-v447b",
    "text_payload": "\u001b[36mINFO\u001b[0m[0120] Pushed image to 1 destinations               ",
  },
  {
    "timestamp": "2021-08-27T06:58:57.15303975Z",
    "environment": "dev",
    "namespace": "image-builder",
    "pod_name": "batch-1630047384-3-v447b",
    "text_payload": "\u001b[36mINFO\u001b[0m[0120] something else              ",
  },
]
```

Ensembling log style:
```json
{
  "environment": "staging",
  "namespace": "cartography",
  "logging_url": "http://www.example.com/logging?pod_name=hello&namespace=world",
  "logs": [
    {
      "timestamp": "2020-07-10T21:13:51Z",
      "pod_name": "batch-1630047384-3-v447b",
      "text_payload": "\u001b[36mINFO\u001b[0m[0120] Pushed image to 1 destinations"
    },
    {
      "timestamp": "2020-07-10T21:13:52Z",
      "pod_name": "batch-1630047384-3-v447b",
      "text_payload": "\u001b[36mINFO\u001b[0m[0120] something else"
    },
  ]
}
```

The motivation is that the Spark executor logs will disappear after the pod has done its job as Kubernetes will delete the pod. Using the router model would mean that once the logs disappear, Turing API is unable to supply the `logging_url` field. Also, factoring out `environment` and `namespace` would save a lot of network traffic and potentially lead to faster API calls although this is not a priority.

EDITS:
- Changed the way of generating openapi stub files; use a docker container instead of relying on system dependencies of openapi-generator as brew was not making it easy for me to install (and I would suspect others too).
- Remove default values and enforce that `component_type` will be required, no empty query params will be allowed (i.e. `?head_lines=&tail_lines=&component_type=router`). UI changes are expected in a separate PR that will be picked up by @romanwozniak.


romanwozniak:
> There are a couple of changes from the router logs endpoint. Here's a comparison: ...

How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod

ashwinath:
> > There are a couple of changes from the router logs endpoint. Here's a comparison: ...
> 
> How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod

It works exactly as how the current router logs work since the logic is shared. So if a call is `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs?component_type=executor` then it gets all the logs from all the pods with the same logic as turing routers.

It will look something like this, the logging URL will be common which is why it is outside the log list.
```json
{
  "environment": "staging",
  "namespace": "cartography",
  "logging_url": "http://www.example.com/logging?pod_name=hello&namespace=world",
  "logs": [
    {
      "timestamp": "2020-07-10T21:13:51Z",
      "pod_name": "pod-1",
      "text_payload": "\u001b[36mINFO\u001b[0m[0120] Pushed image to 1 destinations"
    },
    {
      "timestamp": "2020-07-10T21:13:52Z",
      "pod_name": "pod-2",
      "text_payload": "\u001b[36mINFO\u001b[0m[0120] something else"
    },
    {
      "timestamp": "2020-07-10T21:13:53Z",
      "pod_name": "pod-3",
      "text_payload": "\u001b[36mINFO\u001b[0m[0120] something else from pod 3"
    },
  ]
}
```

ashwinath:
@romanwozniak I have refactored `PodLogService` to be a generic service. Since there are quite a bit of changes (some are quite ugly as there are many calls to `EnsemblingJobService` to generate diff parts). I have also tested the full flow with our internal system and it generates the correct monitoring and logging url.

Regenerating the OpenAPI stub files caused some minor errors which I missed in the previous PR https://github.com/gojek/turing/pull/87.

romanwozniak:
> > > There are a couple of changes from the router logs endpoint. Here's a comparison: ...
> > 
> > 
> > How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod
> 
> It works exactly as how the current router logs work since the logic is shared. So if a call is `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs?component_type=executor` then it gets all the logs from all the pods with the same logic as turing routers.
> 
> It will look something like this, the logging URL will be common which is why it is outside the log list.
>  ...

sorry @ashwinath, I somehow missed this comment. So my concern and the reason of why I left that comment is that in this case `loggingURLTemplate` should support multiple pods simultaneously i.e. if `component_type` is `executor` and there are 3 replicas of executor (pod1, pod2, pod3), then logging URL should be something like `http://www.example.com/logging?pod_name=pod1&pod_name=pod2&pod_name=pod3&namespace=world"`. Or alternatively, it should support regex or something

api/turing/api/pod_log_api.go
```diff
@@ -48,78 +68,217 @@ func (c PodLogController) ListPodLogs(_ *http.Request, vars RequestVars, _ inter
...
-		servicebuilder.ComponentTypes.Enricher,
-		servicebuilder.ComponentTypes.Ensembler:
-	case "":
-		componentType = "router"
-	default:
-		return BadRequest(fmt.Sprintf("Invalid component type '%s'", componentType),
-			"must be one of router, enricher or ensembler")
+	componentType, err := getComponentType(servicebuilder.ComponentTypes.Router, vars, routerAllowedComponents)
+	if err != nil {
+		return BadRequest(err.Error(), "must be one of router, enricher or ensembler")
```
romanwozniak, line 73: Can this `"must be one of router, enricher or ensembler"` message be generated based on `routerAllowedComponents`, instead of a hard-coded string?

romanwozniak: Also, a better idea would be to automate this by specifying the schema of query variables together with validation tags and then use `ParseVars` to parse and validate them. 

Example: https://github.com/gojek/turing/blob/9a728b1132aa2d7f7b1a4af093d103fcf5ba5afe/api/turing/api/ensembling_job_api.go#L97

api/turing/api/pod_log_api.go
```diff
@@ -48,78 +68,217 @@ func (c PodLogController) ListPodLogs(_ *http.Request, vars RequestVars, _ inter
...
+		DefaultContainer: cluster.KnativeUserContainerName,
+		Environment:      router.EnvironmentName,
+		LabelSelectors: []service.LabelSelector{
+			{
+				Key:   cluster.KnativeServiceLabelKey,
+				Value: servicebuilder.GetComponentName(routerVersion, componentType),
+			},
+		},
+	}
+	varLogError := c.parsePodLogRequest(&request, vars)
```
romanwozniak, line 87: Same here, please use validator and `ParseVars`

api/turing/api/pod_log_api.go
```diff
@@ -48,78 +68,217 @@ func (c PodLogController) ListPodLogs(_ *http.Request, vars RequestVars, _ inter
...
+		return componentType, nil
+	}
+
+	return "", fmt.Errorf("Invalid component type '%s'", componentType)
+}
+
+type ListEnsemblingPodLogsOptions struct {
+	ProjectID *models.ID `schema:"project_id" validate:"required"`
+	ID        *models.ID `schema:"job_id" validate:"required"`
+}
```
romanwozniak, line 195: This can be extended with other query variables (like `component_type`, `since_time` etc)

api/turing/api/pod_log_api.go
```diff
@@ -48,78 +68,217 @@ func (c PodLogController) ListPodLogs(_ *http.Request, vars RequestVars, _ inter
...
+	if err != nil {
+		return InternalServerError("Failed to list logs", err.Error())
+	}
+
+	loggingURL, err := c.EnsemblingJobService.FormatLoggingURL(ensemblerName, namespace, componentType)
+	if err != nil {
+		return InternalServerError("Failed to format monitoring URL", err.Error())
+	}
+
+	logs := service.ConvertPodLogsToV2(
```
romanwozniak, line 169: Maybe add some inline comments here about this logic?

api/turing/cluster/labeller/labeller.go
```diff
@@ -3,18 +3,18 @@ package labeller
 import "fmt"
 
 const (
-	// orchestratorValue is the value of the orchestrator (which is Turing)
+	// OrchestratorValue is the value of the orchestrator (which is Turing)
 	orchestratorValue = "turing"
-	// environmentLabel refers to the environment the Kubernetes object is in
+	// EnvironmentLabel refers to the environment the Kubernetes object is in
```
romanwozniak, line 8: Isn't it expected that comment matches the variable name?

romanwozniak (commented):
I didn't find anything particularly "ugly" :) But I left some suggestions on how to improve the validation of request vars

api/turing/cluster/labeller/labeller.go
ashwinath: oops, I made it public then back to private and forgot to change it. My editor nags me when it's public and not exported with a comment but doesn't do the converse action.

api/turing/api/pod_log_api.go
ashwinath: alright

api/api/specs/common.yaml
```diff
@@ -10,7 +10,6 @@ components:
     Id:
       type: "integer"
       format: "int32"
-      readOnly: true
```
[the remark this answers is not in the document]
ashwinath: Because Id is used by mlp project ID as well, which is not read only. It breaks a lot of the APIs when I tested it.

README.md
```diff
@@ -160,7 +160,6 @@ router Docker image is set to the image built and push from the previous step.
 # config-dev.yaml
 BatchEnsemblingConfig:
   Enabled: true
-  MonitoringURLTemplate: https://grafana.example.com/graph/d/foo/dashboard?var-project={{.Project}}&var-job={{.Job}}
```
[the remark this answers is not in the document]
ashwinath: Hmm. My thought was this is an optional variable so I ommited it from the config-dev.yaml but left it in config/example.yaml with comments.

api/turing/api/pod_log_api.go
ashwinath: I hesitated on this while implementing the logic because, 

> > > > There are a couple of changes from the router logs endpoint. Here's a comparison: ...
> > > 
> > > 
> > > How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod
> > 
> > 
> > It works exactly as how the current router logs work since the logic is shared. So if a call is `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs?component_type=executor` then it gets all the logs from all the pods with the same logic as turing routers.
> > It will look something like this, the logging URL will be common which is why it is outside the log list.
> > ...
> 
> sorry @ashwinath, I somehow missed this comment. So my concern and the reason of why I left that comment is that in this case `loggingURLTemplate` should support multiple pods simultaneously i.e. if `component_type` is `executor` and there are 3 replicas of executor (pod1, pod2, pod3), then logging URL should be something like `http://www.example.com/logging?pod_name=pod1&pod_name=pod2&pod_name=pod3&namespace=world"`. Or alternatively, it should support regex or something

What is the use case for this? I find this to be very oddly specific and not sure how this can be used in the UI. All podnames are too similar. I don't see the value of specifying the pod name. Let me give you an example of how the pod names are like:

Builder: only 1 pod so there is no differentiation `component_type=image_builder` is good enough.
Driver: only 1 driver so there is no differentiation `component_type=driver` is good enough.

Executor: may have multiple but `component_type=executor` will filter all of them out. So what is left is to differentiate between the pods. From what I remember, all executor pod have names like this `<ensembler_name>-<replicaset-hash>-exec-<replica number indexed from 0>`. Personally I don't see any value differentiating pod 1 from 2 and 3.

ashwinath:
I hesitated on this while implementing the logic because, 

> > > > There are a couple of changes from the router logs endpoint. Here's a comparison: ...
> > > 
> > > 
> > > How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod
> > 
> > 
> > It works exactly as how the current router logs work since the logic is shared. So if a call is `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs?component_type=executor` then it gets all the logs from all the pods with the same logic as turing routers.
> > It will look something like this, the logging URL will be common which is why it is outside the log list.
> > ...
> 
> sorry @ashwinath, I somehow missed this comment. So my concern and the reason of why I left that comment is that in this case `loggingURLTemplate` should support multiple pods simultaneously i.e. if `component_type` is `executor` and there are 3 replicas of executor (pod1, pod2, pod3), then logging URL should be something like `http://www.example.com/logging?pod_name=pod1&pod_name=pod2&pod_name=pod3&namespace=world"`. Or alternatively, it should support regex or something

What is the use case for this? I find this to be very oddly specific and not sure how this can be used in the UI. All podnames are too similar. I don't see the value of specifying the pod name. Let me give you an example of how the pod names are like:

Builder: only 1 pod so there is no differentiation `component_type=image_builder` is good enough.
Driver: only 1 driver so there is no differentiation `component_type=driver` is good enough.

Executor: may have multiple but `component_type=executor` will filter all of them out. So what is left is to differentiate between the pods. From what I remember, all executor pod have names like this `<ensembler_name>-<replicaset-hash>-exec-<replica number indexed from 0>`. Personally I don't see any value differentiating pod 1 from 2 and 3.

api/turing/api/pod_log_api.go
ashwinath: I don't think this can be done with the current playground validator. I can't put a default value but i need it to be `oneof=router enricher ensembler`. But if it's empty it should be router. I'm not sure if the mux schema can do this or if validator can do this.

api/turing/api/pod_log_api.go
ashwinath: I don't think this can be done with default variables, let's continue the discussion here: https://github.com/gojek/turing/pull/88#discussion_r703165527

api/turing/api/pod_log_api.go
ashwinath: Do you remember the reason why David wrote it like this in the first place? Instead of using playground validator.

ashwinath:
> > > > > There are a couple of changes from the router logs endpoint. Here's a comparison: ...
> > > > 
> > > > 
> > > > How this `logging_url` would work for multiple executor pods? I.e. If I have a job with 5 executor replica, then I'd expect to have my log entries to be grouped by the `component_type` (i.e `executor`) and then `logging_url` should be common for each executor pod
> > > 
> > > 
> > > It works exactly as how the current router logs work since the logic is shared. So if a call is `GET ​/projects​/{project_id}​/jobs​/{job_id}​/logs?component_type=executor` then it gets all the logs from all the pods with the same logic as turing routers.
> > > It will look something like this, the logging URL will be common which is why it is outside the log list.
> > > ...
> > 
> > 
> > sorry @ashwinath, I somehow missed this comment. So my concern and the reason of why I left that comment is that in this case `loggingURLTemplate` should support multiple pods simultaneously i.e. if `component_type` is `executor` and there are 3 replicas of executor (pod1, pod2, pod3), then logging URL should be something like `http://www.example.com/logging?pod_name=pod1&pod_name=pod2&pod_name=pod3&namespace=world"`. Or alternatively, it should support regex or something
> 
> What is the use case for this? I find this to be very oddly specific and not sure how this can be used in the UI. All podnames are too similar. I don't see the value of specifying the pod name. Let me give you an example of how the pod names are like:
> 
> Builder: only 1 pod so there is no differentiation `component_type=image_builder` is good enough.
> Driver: only 1 driver so there is no differentiation `component_type=driver` is good enough.
> 
> Executor: may have multiple but `component_type=executor` will filter all of them out. So what is left is to differentiate between the pods. From what I remember, all executor pod have names like this `<ensembler_name>-<replicaset-hash>-exec-<replica number indexed from 0>`. Personally I don't see any value differentiating pod 1 from 2 and 3.

Ok i think the confusion was why is there only one logging URL if there are multiple pods. The answer can be found here: https://github.com/gojek/turing/pull/88/files#diff-363acf5d843a16aeaa0efa8fc1283edc32bbd24622f06dcd5eb61f438eb894b4R36-R40 where we fit the pod pattern into a regex and that is sent inside the url. But whether `.*` wildcards are supposed in the URL is not something we want to think about.

api/turing/api/pod_log_api.go
romanwozniak: Yeah, the `default` tag was only added recently and is still not merged into the main branch: https://github.com/gorilla/schema/pull/183, but as discussed separately, maybe we don't really need default values at all.

> Do you remember the reason why David wrote it like this in the first place? 
Can't remember reviewing it :)

ashwinath: As discussed in private, we will remove default values and enforce that `component_type` will be required, no empty query params will be allowed (i.e. `?head_lines=&tail_lines=&component_type=router`). UI changes are expected in a separate PR that will be picked up by @romanwozniak.

ashwinath: I have overhauled the parser/validator.

api/turing/api/pod_log_api.go
```diff
@@ -189,9 +188,18 @@ func getComponentType(defaultComponentType string, vars RequestVars, allowedType
 	return "", fmt.Errorf("Invalid component type '%s'", componentType)
 }
 
-type ListEnsemblingPodLogsOptions struct {
-	ProjectID *models.ID `schema:"project_id" validate:"required"`
+type podLogOptions struct {
 	ID        *models.ID `schema:"job_id" validate:"required"`
```
romanwozniak: Shall this be moved to `listEnsemblingPodLogsOptions` instead? I can imagine that it needs to be `schema:"router_id"` for router's pod logs options

romanwozniak: NVM, it was already changed in the next commit

api/turing/api/pod_log_api.go
```diff
@@ -176,91 +163,26 @@ func (c PodLogController) ListEnsemblingJobPodLogs(_ *http.Request, vars Request
...
-}
-
-type ListEnsemblingPodLogsOptions struct {
+type podLogOptions struct {
 	ProjectID *models.ID `schema:"project_id" validate:"required"`
-	ID        *models.ID `schema:"job_id" validate:"required"`
+	Previous  bool       `schema:"previous"`
+	HeadLines *int64     `schema:"head_lines" validate:"omitempty,gte=0"`
+	TailLines *int64     `schema:"tail_lines" validate:"omitempty,gte=0"`
+	SinceTime *time.Time `schema:"since_time"`
```
romanwozniak, line 171: What is the default time format? Is it `RFC3339`? Can it also recognize millis (example: `2006-01-02T15:04:05.244Z`)?

romanwozniak (approved):
LGTM!

api/turing/api/pod_log_api.go
ashwinath: Yes it can, the reason why I did not put the validator is because it will fail the parsing stage, which is done before the validation.

PR #91, opened 2021-09-06
UI: Show config diff before updating the router

romanwozniak:
### Context:
To make the process of updating routers less error prone, we would like to show to users a diff between the old and new router's configuration. 

This PR introduces some refactoring to existing components and adds into the router update form this extra step, that renders a text-based diff of the configuration, before making it possible to deploy the changes.

### Demo:
https://user-images.githubusercontent.com/1886194/132236057-6cfb9061-92ae-4454-b1a9-811134c13943.mp4


ui/src/router/edit/EditRouterView.js
```diff
@@ -1,37 +1,100 @@
...
 import { UpdateRouterForm } from "../components/form/UpdateRouterForm";
 import { ExperimentEngineContextProvider } from "../../providers/experiments/ExperimentEngineContextProvider";
+import { VersionComparisonView } from "./components/VersionComparisonView";
+import { useTuringApi } from "../../hooks/useTuringApi";
+import { DeploymentSummary } from "../components/form/components/DeploymentSummary";
+import { ConfirmationModal } from "../../components/confirmation_modal/ConfirmationModal";
+
+const EditRouterView = ({ projectId, currentRouter, ...props }) => {
+  const { data: updatedRouter } = useContext(FormContext);
+  const [showDiffView, toggleDiffView] = useToggle();
```
krithika369, line 17: Can the `StepsWizardHorizontal` not be reused for this (hiding the `EuiStepsHorizontal` if needed)?

krithika369 (commented):
Do you think it would be good to also add this yaml display to the create router wizard (nothing to compare, but a summary of the current configs in yaml)?

ui/src/router/edit/EditRouterView.js
romanwozniak: That's what I initially started exploring, but then dropped, because it will break the layout and there is no a simple way to customize it:

_Current:_
Screenshot 2021-09-07 at 08 43 39

_With `StepsWizardHorizontal`:_
Screenshot 2021-09-07 at 08 45 32

romanwozniak:
> Do you think it would be good to also add this yaml display to the create router wizard (nothing to compare, but a summary of the current configs in yaml)?

We use YAML only because it's the easiest way to represent the diff between two configs. If there is no comparison required, then I'm not sure if YAML is the best way to represent the summary of deployment

krithika369 (approved):
LGTM!

ui/src/router/edit/EditRouterView.js
krithika369: I see.. Ok, let's keep it the way it is now.

PR #92, opened 2021-09-07
Update OpenAPI library version

krithika369:
This PR updates the `getkin/kin-openapi` library version used by the Turing API to the latest at `v0.75.0`. This is required because the internal Turing repo's dependencies use a newer version of the same library which causes conflicts.

### DIFF
--- a/api/turing/middleware/openapi_validation.go
+++ b/api/turing/middleware/openapi_validation.go
@@ -8,12 +8,14 @@ import (
 
 	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers"
+	legacyrouter "github.com/getkin/kin-openapi/routers/legacy"
 )
 
 // OpenAPIValidation middleware validates HTTP requests against OpenAPI spec.
 type OpenAPIValidation struct {
-	router  *openapi3filter.Router
 	options *openapi3filter.Options
+	router  routers.Router
 }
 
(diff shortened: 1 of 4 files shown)


api/turing/middleware/openapi_validation.go
```diff
@@ -30,34 +32,36 @@ func NewOpenAPIValidation(
...
-	opts := &openapi3filter.Options{}
 
-	if options.IgnoreAuthentication {
-		// when IgnoreAuthentication is true, the authentication function always succeed i.e returns nil error.
-		opts.AuthenticationFunc = func(_ context.Context, _ *openapi3filter.AuthenticationInput) error {
-			return nil
-		}
+	// Create router
+	var router routers.Router
+	if router, err = legacyrouter.NewRouter(swagger); err != nil {
```
romanwozniak, line 55: Why do we still need to use `legacyrouter`?

romanwozniak (approved):
LGTM

api/turing/middleware/openapi_validation.go
krithika369: No strong reason. The alternative is `gorilla/mux` router, but in comparison, the legacy router supposedly captures more granular errors: https://pkg.go.dev/github.com/getkin/kin-openapi@v0.75.0/routers/legacy

On the other hand, the `gorilla/mux` router can match variables in the paths with extensions like `{id}.json`. But we don't really use it.

Since the `legacyrouter` is closer to what we were using earlier, I picked it. It's a one line change to opt for the other, if we need to later.

PR #93, opened 2021-09-08
Add RunID Generation for ensembling jobs.

ashwinath:
As requested from this [thread](https://github.com/gojek/turing/pull/87#discussion_r694403174) we are supplying a `run_id` to give ensembling jobs more sane and shorter names.

### DIFF
--- a/api/api/specs/ensemblers.yaml
+++ b/api/api/specs/ensemblers.yaml
@@ -149,7 +149,7 @@ components:
         name:
           type: "string"
           minLength: 3
-          maxLength: 15
+          maxLength: 20
         created_at:
           type: "string"
           format: "date-time"
--- a/api/turing/service/ensembling_job_service.go
+++ b/api/turing/service/ensembling_job_service.go
@@ -225,11 +225,6 @@ func (s *ensemblingJobService) List(options EnsemblingJobListOptions) (*Paginate
 	return paginatedResults, nil
 }
 
-func generateDefaultJobName(ensemblerName string) string {
-	t := time.Now().Unix()
-	return fmt.Sprintf("%s-%d", ensemblerName, t)
-}
-
 func getEnsemblerDirectory(ensembler *models.PyFuncEnsembler) string {
 	// Ensembler URI will be a local directory
 	// Dockerfile will build copy the artifact into the local directory.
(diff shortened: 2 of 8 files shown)


api/turing/service/ensembling_job_service.go
```diff
@@ -257,10 +251,6 @@ func (s *ensemblingJobService) generateMonitoringURL(job *models.EnsemblingJob,
 	}
 
 	name := job.Name
```
romanwozniak, line 253: Is this separate variable `name` still required?

api/api/specs/ensemblers.yaml
```diff
@@ -149,7 +149,7 @@ components:
         name:
           type: "string"
           minLength: 3
-          maxLength: 15
+          maxLength: 20
```
romanwozniak, line 152: Again, how did we come up with this number? Does it need to be that strict, or `25` is fine too?

romanwozniak (approved):
LGTM

api/api/specs/ensemblers.yaml
ashwinath: I think 25 might be risky, I tested with 20 in our internal dashboard. In the description linked thread I explained how the pod names are generated. But regarding the length of the hashes generated by spark, I'm not too sure nor did I really specifically look into the implementation. I just did a trial and error and it looked good enough.

romanwozniak: Okay

api/turing/service/ensembling_job_service.go
ashwinath: yes this is still needed to generate, but the caveat is that it should be called after db.Save() to trigger the name generation.

ashwinath: MonitoringURL is still generated at runtime and not persisted into the database.

romanwozniak: Sorry, I meant do you still need `name := job.Name` here, or you can just use `job.Name` whenever you need that `name` variable (i.e. L256)

ashwinath: Oh I see, just asthetic reasons. Yeah I just inlined it.

PR #94, opened 2021-09-08
UI: Add Ensembling Jobs Logs Viewer

romanwozniak:
### Context: 
This PR integrates with https://github.com/gojek/turing/pull/88 and makes it possible to view ensembling job execution logs in the Turing console.

### Details:
- This PR introduces [`useTuringPollingApiEmitter`](https://github.com/gojek/turing/compare/main...romanwozniak:ui_ensembling_job_logs?expand=1#diff-25f4de81b9a9df887ff18799f3e95b41364e476b68c2adcd6cd9cfd0958a90f1) hook, that makes it possible to poll Turing API endpoint with a configured interval and subscribe to data from it via provided event emitter. 

- Existing component for viewing router logs was refactored to utilize `useTuringPollingApiEmitter` and generalized to support both Router Logs API and Jobs Logs API (v2)

### Screenshot:
Screenshot 2021-09-09 at 00 07 21


ui/src/config.js
```diff
@@ -43,32 +43,40 @@ export const appConfig = {
...
   tables: {
     defaultTextSize: "s",
     defaultIconSize: "s",
-    dateFormat: "YYYY-MM-DDTHH:mm.SSZ"
-  }
+    dateFormat: "YYYY-MM-DDTHH:mm.SSZ",
+  },
+  podLogs: {
+    // Interval (in ms) between API calls to Logs API
+    pollInterval: 10000,
```
krithika369, line 58: Is 10s too long? I believe this was 5s previously.

ui/src/components/pod_logs_viewer/hooks/useLogsEmitter.js
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    apiEmitter.on("data", (entries) => {
+      const lastTimestamp = extractTimestamp(entries);
+
+      if (!!lastTimestamp) {
+        setQuery((q) => ({
+          ...q,
+          since_time: lastTimestamp,
+          head_lines: configOptions.batchSize,
```
krithika369, line 43: CMIIW, but per our [Swagger docs](https://github.com/gojek/turing/blob/main/api/api/specs/logs.yaml#L54) for the Pod logs API, `If both tail_lines and head_lines are set, tail_lines will be ignored`. Soo it seems to me like the `head_lines` should not be set if the user chose an option besides "from the container start".

krithika369 (approved):
The refactoring looks great. Left some questions / comments. Besides these, it LGTM.

ui/src/components/collapsible_badge_group/CollapsibleBadgeGroup.js
```diff
@@ -30,7 +30,11 @@ export const CollapsibleBadgeGroup = ({
       );
 
       if (!isExpanded && items.length > minItemsCount) {
-        children.push(<EuiLink {...getToggleProps()}>Show All</EuiLink>);
+        children.push(
+          <EuiLink key={minItemsCount + 1} {...getToggleProps()}>
```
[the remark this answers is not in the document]
romanwozniak: That's a minor bug from https://github.com/gojek/turing/pull/89. Each element in the array should have a unique id to help React identify which items have changed (added/removed/reordered). The `key` was added to each `EuiBadge` (L:24), but was missed from this `EuiLink`.

ui/src/config.js
romanwozniak: `pollInterval` is only used when no new log entries are received in the current request. Otherwise, a new request with updated `since_time` query parameter is fired immediately. So this has very little effect on the pods (routers/jobs) with frequently updated logs data, but reduces the number of API calls if there is no updates (i.e. router is scaled down to 0 replicas)

ui/src/components/pod_logs_viewer/hooks/useLogsEmitter.js
romanwozniak: This is expected. When user chose for example "Last 100 records", then first API call is
```
?taill_ines=100
``` 
Then, we fetch a timestamp of the latest log entry from the first call AND (just in case) limit the max number of records (i.e. `head_lines`) that will be fetched in the next (and subsequent) API calls:
```
?head_lines=500&since_time=2021-09-09:12:00:00.234Z&tail_lines=ignored
```

PR #95, opened 2021-09-09
[BUG] Fix config diff view for the scenarios where experiment engine configuration has been updated

romanwozniak:
### Context:
This PR fixes the config diff view, that is shown to users on router update confirmation dialog. It was previously identified, that in some scenarios, the YAML representation of the config contains the data, that shouldn't be there. Such as:
1. When experiment engine client's passkey wasn't updated, it shouldn't be highlighted on a diff view
2. When the router, that previously had no experiment engine configured is updated with a new experiment engine configuration, the diff should show non encrypted client's passkey.
3. Experiment variables section shouldn't contain helper data fields, such as `[idx, selected]`

### Implementation Notes:
For the bug #3, it's preferred to not include such helper information into the API payload data object at all (and instead keep it in a separate state). However, this requires a complete refactoring of the current implementation of the experiment engine form components, hence I decided to just strip these unused keys from the data object in a post-processing step.


ui/src/utils/object.js
```diff
@@ -1 +1,19 @@
-export const isEmpty = obj => !obj || !Object.keys(obj).length;
+export const isEmpty = (obj) => !obj || !Object.keys(obj).length;
+
+// Creates a copy of original object and removes specific keys from it
+export const stripKeys = (obj, keys, deep = true) =>
```
ashwinath, line 4: Not gonna lie, took me some time to understand this function. Not sure about the conventions here but I find that deeply nested ternary operators are extremely hard to read.

ashwinath: IMO splitting into if else branches would have been easier to process, something like this:

```javascript
const stipKeysV2 = (obj, keys, deep = true) => {
    if (obj !== Object(obj)) return obj

    if (Array.isArray(obj)) {
        if (!deep) return obj
        return obj.map((item) => stripKeys(item, keys, deep))

    return Object.keys(obj)
      .filter((k) => !keys.includes(k))
      .reduce(
        (acc, x) =>
          Object.assign(acc, {
            [x]: deep ? stripKeys(obj[x], keys, deep) : obj[x],
          }),
        {}
      );
}
```

romanwozniak: Done

ashwinath (approved):

PR #96, opened 2021-09-20
Tag latest ensembler

ashwinath:
Right now it does not have the latest tag with each new release. This PR adds the latest tag so that we can always pull the latest for testing purposes.

### DIFF
--- a/.github/workflows/engines-batch-ensembler.yaml
+++ b/.github/workflows/engines-batch-ensembler.yaml
@@ -82,4 +82,4 @@ jobs:
         run: echo ${{ secrets.GHCR_TOKEN }} | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
 
       - name: Publish image
-        run: docker push ${{ env.DOCKER_REGISTRY }}/turing/batch-ensembler:$(make version)
+        run: make push-image
--- a/engines/batch-ensembler/Dockerfile
+++ b/engines/batch-ensembler/Dockerfile
@@ -31,7 +31,7 @@ RUN chmod 644 -R /etc/metrics/conf/*
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
 
 # Install wget and other libraries required by Miniconda3
-RUN apt-get update -q && \
+RUN apt-get update --allow-releaseinfo-change -q && \
     apt-get install -q -y \
         bzip2 \
         ca-certificates \
--- a/engines/batch-ensembler/Makefile
+++ b/engines/batch-ensembler/Makefile
@@ -32,8 +32,14 @@ build-image:
 	mkdir -p temp-deps
 	cp -r ../../sdk temp-deps/
 	@docker build . \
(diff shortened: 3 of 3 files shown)


romanwozniak:
Using `latest` tag is an anti-pattern for most of scenarios. It kills the reproducibility of the deployment as well as introduces some hard-to-discover side effects (i.e. if deployed with kubernetes and `imagePullPolicy` is set to `IfNotPresent`, then the resulting deployment might be not what you expect it to be).

ashwinath:
> Using `latest` tag is an anti-pattern for most of scenarios. It kills the reproducibility of the deployment as well as introduces some hard-to-discover side effects (i.e. if deployed with kubernetes and `imagePullPolicy` is set to `IfNotPresent`, then the resulting deployment might be not what you expect it to be).

I agree with you on this, if you set the `latest` tag in your deployments the developer ought to be slapped. But I dont think that's the use case for this latest tag. It's good for stuff like integration tests, or if someone wants to play around with the container without having to lookup the commit hash they need for the latest part.

But if you feel strongly about not supporting the latest tags, I'll just close the MR.

romanwozniak:
> > Using `latest` tag is an anti-pattern for most of scenarios. It kills the reproducibility of the deployment as well as introduces some hard-to-discover side effects (i.e. if deployed with kubernetes and `imagePullPolicy` is set to `IfNotPresent`, then the resulting deployment might be not what you expect it to be).
> 
> I agree with you on this, if you set the `latest` tag in your deployments the developer ought to be slapped. But I dont think that's the use case for this latest tag. It's good for stuff like integration tests, or if someone wants to play around with the container without having to lookup the commit hash they need for the latest part.
> 
> But if you feel strongly about not supporting the latest tags, I'll just close the MR.

I'm not completely agains `latest` tag, but:
a) It should only be added to stable releases, not dev builds
b) It should not be used in e2e tests for the reproducibility reasons I described above

This MR doesn't seem to be complaint with these criteria

ashwinath:
> 
> 
> > > Using `latest` tag is an anti-pattern for most of scenarios. It kills the reproducibility of the deployment as well as introduces some hard-to-discover side effects (i.e. if deployed with kubernetes and `imagePullPolicy` is set to `IfNotPresent`, then the resulting deployment might be not what you expect it to be).
> > 
> > 
> > I agree with you on this, if you set the `latest` tag in your deployments the developer ought to be slapped. But I dont think that's the use case for this latest tag. It's good for stuff like integration tests, or if someone wants to play around with the container without having to lookup the commit hash they need for the latest part.
> > But if you feel strongly about not supporting the latest tags, I'll just close the MR.
> 
> I'm not completely agains `latest` tag, but:
> a) It should only be added to stable releases, not dev builds
> b) It should not be used in e2e tests for the reproducibility reasons I described above
> 
> This MR doesn't seem to be complaint with these criteria

Actually on second thoughts, it's not good on integration tests.

Okay it makes sense, it should be on tags. Closing this PR.

PR #97, opened 2021-09-21
[FIX]: Update helm chart releaser action to publish charts to `turing-ml` org

romanwozniak:
## Context:
GitHub Pages is a low-effort and generally great way to host Helm charts registry of the project. We were using it before, until `gojek` org configured its own organization-level pages with a custom domain (`www.gojek.io`). After that was done, our existing GH Pages URL (https://gojek.github.io/turing) is forcefully redirected to https://www.gojek.io/turing, which expectedly throws 404 error. 

The reason why it happens is described in this GitHub Issue: https://github.com/isaacs/github/issues/547, which remains unresolved since 2016 (and unlikely to be resolved anytime soon). 

## Solution: 
We can't use this project's GH Pages because of the reason described above, but we still can leverage on GH Pages as Helm chart registry if we publish charts to another repository outside of `gojek` org and that's exactly what https://github.com/stefanprodan/helm-gh-pages helps to achieve.

I've created https://github.com/turing-ml organization and https://github.com/turing-ml/charts repository under it. Workflow under `gojek/turing` project is updated to publish the chart into https://github.com/turing-ml/charts (`gh-pages` branch), so users will be able to install turning with:
```
helm repo add turing https://turing-ml.github.io/charts

helm install turing turing/turing
```


krithika369 (approved):
LGTM!

PR #98, opened 2021-09-22
Generate router monitoring URL on API side rather than client side.

ashwinath:
Router Monitoring URL is now generated on the API side. UI has also been changed to use the values from the Turing API.


api/turing/api/router_versions_api.go
```diff
@@ -24,11 +24,31 @@ func (c RouterVersionsController) ListRouterVersions(
...
+
 	// List router versions
 	routerVersions, err := c.RouterVersionsService.ListRouterVersions(router.ID)
 	if err != nil {
 		return InternalServerError("unable to retrieve router versions", err.Error())
 	}
+
+	// Append monitoring URL to router versions
+	for _, routerVersion := range routerVersions {
+		monitoringURL, err := c.RouterVersionsService.GenerateMonitoringURL(
```
romanwozniak, line 40: Is there any particular reason to have this done at the controller level, instead of a service level (similarly how it's done for ensembling jobs)?

romanwozniak: Also, kinda nitpicking, but you can avoid creating a new `monitoringURL` variable for each item in the slice:
```golang
var err error
for _, routerVersion := range routerVersions {
  routerVersion.MonitoringURL, err = c.RouterVersionsService.GenerateMonitoringURL(...)
  if err != nil {
    return InternalServerError("unable to generate monitoringURL for router version", err.Error()). 
  }
}
```

api/turing/api/routers_api.go
```diff
@@ -35,6 +35,20 @@ func (c RoutersController) ListRouters(
 		return InternalServerError("unable to list routers", err.Error())
 	}
 
+	// Append monitoring URL to router
+	for _, router := range routers {
+		monitoringURL, err := c.RouterVersionsService.GenerateMonitoringURL(
```
romanwozniak, line 40: Same comment about whether service is a better place to implement this than controller

api/turing/config/config.go
```diff
@@ -268,6 +268,8 @@ type RouterDefaults struct {
 	// {"experiment_engine_a": {"endpoint": "http://engine-a.com", "timeout": "500ms"},
 	//  "experiment_engine_b": {"endpoint": "http://engine-b.com", "timeout": "250ms"} }
 	Experiment map[string]interface{}
+
+	MonitoringURLFormat *string
```
romanwozniak, line 272: Is it possible to use the same variable for both jobs and routers?

ashwinath: Yeah it's under a different struct.

api/turing/api/router_versions_api.go
ashwinath: Yeah, actually I feel that it's kinda weird to just abstract the for loop out into a service. It looks long to you because of the linter (I have to split the 1 function call into 6 lines). Your example below shows how simple the line is already, it's already in a service. If I were to do it in the service, all it does is a for loop, add an additional dependency, router versions service into the router service and process this again. For now the services looks simple but when things get more complicated, we may run into a dependency cycle in the future. i.e. Service A depends on Service B and Service B depends on Service C. What if Service C requires a dependency on service A or it's just mentally difficult to process all the dependencies.

I would rather push for a mediator pattern that is decoupled from the database calls than to push this to the service.

ashwinath: But I also understand why you would want it there:
```go
	if router, errResp = c.getRouterFromRequestVars(vars); errResp != nil {
		return errResp
	}
	if routerVersion, errResp = c.getRouterVersionFromRequestVars(vars); errResp != nil {
		return errResp
	}
```

This should have handled everything. Anyway, I'll do it but this might be a problem next time.

ashwinath: Done.

api/turing/api/routers_api.go
ashwinath: done

ui/src/config.js
```diff
@@ -72,10 +72,6 @@ export const resultLoggingConfig = {
   protoUrl: process.env.REACT_APP_RESULT_LOG_PROTO_URL,
 };
 
-export const monitoringConfig = {
-  dashboardUrl: process.env.REACT_APP_ROUTER_MONITORING_URL || "",
```
romanwozniak, line 76: You should also remove it from `ui/.env.development`

romanwozniak (approved):

PR #99, opened 2021-09-23
Add environment variables for batch ensembling

ashwinath:
This PR adds environment variable configuration via the SDK. The API also supports environment variables to be propagated to the Spark driver and executors.

I regenerated the OpenAPI stubs and it seems that I have forgotten to regenerate the stubs from this PR: https://github.com/gojek/turing/pull/93


api/api/specs/jobs.yaml
```diff
@@ -419,6 +419,10 @@ components:
           type: string
         resources:
           $ref: "#/components/schemas/EnsemblingResources"
+        environment_variables:
+          type: object
+          additionalProperties:
+            type: string
```
romanwozniak, line 425: Can we make it consistent (in terms of property name and schema definition) with `routers.yaml#/components/schemas/EnvVar`? 
As regarding schema definition, I'm not sure which one I prefer – `object` with `additionalProperties` or `array` with `EnvVar` items, so I'm open to discussion. Probably, the best would be to stick to `array[EnvVar]` (so it doesn't affect schema of routers) as a type persisted in the database (it's also consistent with how k8s deals with env vars), but allow user to set job's variables as dictionary via turing-sdk.

api/turing/cluster/spark.go
```diff
@@ -128,6 +129,20 @@ func createSparkRequest(request *CreateSparkRequest) (*apisparkv1beta2.SparkAppl
 	}, nil
 }
 
+func addEnvVars(request *CreateSparkRequest) []apicorev1.EnvVar {
```
romanwozniak, line 132: This is a confusing naming of this function (add env vars to what or where?). Also it's hiding the fact, that default variables are added to user-defined variables here. I'd suggest to refactor it to be explicit.

```golang
func getEnvVarFromRequest(request *CreateSparkRequest) []apicorev1.EnvVar {
	var envVars []apicorev1.EnvVar
	for key, value := range request.EnvironmentVariables {
		envVars = append(envVars, apicorev1.EnvVar{key, value})
	}

	return envVars
}

// And then, combine user-defined vars with default variables where you need them

Env: append(defaultEnvVars, ...getEnvVarFromRequest(request))
```

sdk/turing/generated/model/ensembler.py
```diff
@@ -62,7 +62,7 @@ class Ensembler(ModelComposed):
 
     validations = {
         ('name',): {
-            'max_length': 15,
+            'max_length': 20,
```
romanwozniak, line 65: Is it re-generated from openapi-specs or a manual fix? :)

romanwozniak (commented):
Done with review and left some suggestions. 
Can you also make sure that UI renders env variables as well? All components and layout is already done as part of my previous PRs, but you might need to change the variable name here: https://github.com/gojek/turing/blob/2c673b24428589f84d71d710f7ad9777f6ac9dbd/ui/src/jobs/details/config/configuration_section/ConfigurationConfigSection.js#L18

sdk/turing/generated/model/ensembler.py
ashwinath: This is regenerated, did not manually modify this.

api/api/specs/jobs.yaml
ashwinath: Ok, did not realise this was used somewhere else as well.

sdk/turing/generated/model/ensembler.py
romanwozniak: Got it. Maybe we need to add it into our CI to check if the generated clients are in sync with the specs. (That's out of scope of this PR)

api/turing/cluster/spark.go
ashwinath: ok

ashwinath:
> Done with review and left some suggestions.
> Can you also make sure that UI renders env variables as well? All components and layout is already done as part of my previous PRs, but you might need to change the variable name here:
> 
> https://github.com/gojek/turing/blob/2c673b24428589f84d71d710f7ad9777f6ac9dbd/ui/src/jobs/details/config/configuration_section/ConfigurationConfigSection.js#L18

Looks good to me:

Screenshot 2021-09-27 at 5 25 08 PM

api/api/specs/jobs.yaml
```diff
@@ -419,6 +419,13 @@ components:
           type: string
         resources:
           $ref: "#/components/schemas/EnsemblingResources"
+        env:
+          type: array
+          items:
+            $ref: "#/components/schemas/EnvVar"
```
romanwozniak, line 425: You should use `common.yaml#/components/schemas/EnvVar` instead, no need to create a local copy of it. 
The reason why it was done before for `Id` schema is because external refs were not resolved properly from within `path` parameter section (i.e.https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/api/specs/jobs.yaml#L19).

api/api/specs/routers.yaml
```diff
@@ -793,14 +793,7 @@ components:
           type: "string"
 
     EnvVar:
```
romanwozniak, line 795: Same, don't need to create a local copy of `EnvVar` schema

ashwinath: done

api/api/specs/jobs.yaml
ashwinath: done

romanwozniak (commented):
One more question, why do we define this type manually, instead of using the one generated from the openapi specs: https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L66?

I likely missed it in the previous review rounds, but now it came to my attention because you are manually adding `EnvVars` field into it, instead of re-generating it from the specs

ashwinath:
> One more question, why do we define this type manually, instead of using the one generated from the openapi specs:
> 
> https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L66
> ?
> 
> I likely missed it in the previous review rounds, but now it came to my attention because you are manually adding `EnvVars` field into it, instead of re-generating it from the specs

InfraConfig has always been hand written. JobConfig is generated. I'm not sure if it should be inside the JobConfig since the value is going to be discarded by the ensembler. I put it into InfraConfig since it is easier to reference the value rather than navigating through the clunky JobConfig. Also I would suppose environment variables has always been an infra thing.

romanwozniak:
> > One more question, why do we define this type manually, instead of using the one generated from the openapi specs:
> > https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L66
> > 
> > ?
> > I likely missed it in the previous review rounds, but now it came to my attention because you are manually adding `EnvVars` field into it, instead of re-generating it from the specs
> 
> InfraConfig has always been hand written. JobConfig is generated. I'm not sure if it should be inside the JobConfig since the value is going to be discarded by the ensembler. I put it into InfraConfig since it is easier to reference the value rather than navigating through the clunky JobConfig. Also I would suppose environment variables has always been an infra thing.

No, the question is not about where `EnvVars` should be (I'm ok with having it under `InfraConfig`). What I'm trying to understand why 
> InfraConfig has always been hand written

ashwinath:
> 
> 
> > > One more question, why do we define this type manually, instead of using the one generated from the openapi specs:
> > > https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L66
> > > 
> > > ?
> > > I likely missed it in the previous review rounds, but now it came to my attention because you are manually adding `EnvVars` field into it, instead of re-generating it from the specs
> > 
> > 
> > InfraConfig has always been hand written. JobConfig is generated. I'm not sure if it should be inside the JobConfig since the value is going to be discarded by the ensembler. I put it into InfraConfig since it is easier to reference the value rather than navigating through the clunky JobConfig. Also I would suppose environment variables has always been an infra thing.
> 
> No, the question is not about where `EnvVars` should be (I'm ok with having it under `InfraConfig`). What I'm trying to understand why
> 
> > InfraConfig has always been hand written

Something's amiss here, am I out of the loop about something? Has the models package been autogenerated or something?

romanwozniak:
> There's some stuff in InfraConfig that is not in the JSON payload, like EnsemblerName.

Ok, this is the answer I was looking for :) However I can see `ensembler_name` defined in the [`EnsemblerInfraConfig `](https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/api/specs/jobs.yaml#L411) schema.

I was wondering why the [`JobConfig`](https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L49) is just an alias to `openapi.EnsemblerConfig`, but [`InfraConfig`](https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/api/turing/models/ensembling_job.go#L66) is defined completely from the scratch. 

If it's the case that `InfraConfig` has some extra fields, compared to openapi schema, then it can still be composed from the base type + extra fields, similar to this:
```golang
type InfraConfig struct {
    openapi.EnsemblerInfraConfig
    EnsemblerName string `json:"ensembler_name"`
} 
```

romanwozniak (approved):

ashwinath:
> ```go
> openapi.EnsemblerInfraConfig
> ```

Wait, then what about validation tags? Some fields like InfraConfig.ServiceAccountName has a required tag but some do not.

ashwinath:
@romanwozniak Could you help to take a look at the last commit one last time? I refactored to use the openapi generated code but it's quite a number of changes involved.

api/turing/models/ensembling_job.go
```diff
@@ -64,15 +64,13 @@ func (c *JobConfig) Scan(value interface{}) error {
...
-	ServiceAccountName string                       `json:"service_account_name" validate:"required"`
-	EnvVars            EnvVars                      `json:"env"`
-	Resources          *openapi.EnsemblingResources `json:"resources"`
+	openapi.EnsemblerInfraConfig
+	ServiceAccountName string `json:"service_account_name" validate:"required"`
 }
 
 // Value returns json value, implement driver.Valuer interface
 func (r *InfraConfig) Value() (driver.Value, error) {
+	// Some weird logic because gorm doesn't know how to marshal embedded structs
```
romanwozniak, line 73: Nothing weird here.

api/turing/models/ensembling_job.go
```diff
@@ -85,6 +83,31 @@ func (r *InfraConfig) Scan(value interface{}) error {
...
+		toSerialize["resources"] = r.Resources.Get()
+	}
+
+	if r.Env != nil {
+		toSerialize["env"] = r.Env
+	}
+
+	return json.Marshal(toSerialize)
+}
+
```
romanwozniak, line 110: This method ruins the entire idea (which is DRY and keeping a single source of truth) of using the generated code. 
Would this work?

```go
func (r *InfraConfig) MarshalJSON() ([]byte, error) {
         r.EnsemblerInfraConfig.SetServiceAccountName(r.ServiceAccountName)

	return json.Marshal(r.EnsemblerInfraConfig)
}
```

api/turing/models/ensembling_job.go
ashwinath: I forgot to erase this.

api/turing/models/ensembling_job.go
ashwinath: I have to test this, there's some weird mechanic when you do `assert.Equal(t, expected, actual)` after querying from database. I'm very sure it will break all the tests because when you receive the payload from the API, you are setting the outer struct's ServiceAccountName. Because with this code you are setting the inner struct's ServiceAccountName while having the outer struct's ServiceAccountName. This actually sets 2 sources of truth.

ashwinath: Yeah this wont work:
```
    ensembling_job_service_test.go:234:
                Error Trace:    ensembling_job_service_test.go:234
                                                        database.go:92
                                                        ensembling_job_service_test.go:203
                Error:          Not equal:
                                expected: &models.InfraConfig{EnsemblerInfraConfig:openapi.EnsemblerInfraConfig{ArtifactUri:(*string)(0xc0005125d0), EnsemblerName:(*string)(0x495ab50), ServiceAccountName:(*string)(0xc0005136f0), Resources:openapi.NullableEnsemblingResources{value:(*o
penapi.EnsemblingResources)(0xc0008724b0), isSet:true}, Env:(*[]openapi.EnvVar)(0xc00049e800)}, ServiceAccountName:"test-service-account-1"}
                                actual  : &models.InfraConfig{EnsemblerInfraConfig:openapi.EnsemblerInfraConfig{ArtifactUri:(*string)(0xc00007d5b0), EnsemblerName:(*string)(0xc00007d5c0), ServiceAccountName:(*string)(nil), Resources:openapi.NullableEnsemblingResources{value:(*openapi
.EnsemblingResources)(0xc00059aed0), isSet:true}, Env:(*[]openapi.EnvVar)(0xc000696d40)}, ServiceAccountName:"test-service-account-1"}
```

PR #101, opened 2021-09-28
Add ability to load custom Experiment Engine UI

krithika369:
This PR adds the ability to load a remote UI component exposed by an experiment engine, at the `/experiments` path of the Turing app. This is made possible by the following 3 configurations (added to `ui/.env.development`):
* `REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_APPNAME`
* `REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_URL`
* `REMOTE_COMPONENTS_PROXY_PATHS`

The first 2 introduce the concept of a 'default' experiment engine, tied to the Turing app. Although Turing support multiple experiment engines when configuring routers, the `/experiments` endpoint is meant to configure the experiments themselves and this will only be possible for one experiment engine that is recognised as the default. In the future, these 2 UI env vars will be removed and the values will instead be obtained through the Turing API.

The 3rd env var will be used only in the local environment, to proxy API requests required by the remote component so that CORS issues can be bypassed.

The remote component is created and consumed with the use of [Module Federation](https://webpack.js.org/concepts/module-federation/). Some information on the main changes are described below.

### Modifications
* `ui/package.json` - Upgrade to CRA5 alpha build, use [craco](https://github.com/gsoft-inc/craco) to override the Webpack configs
* `ui/craco.config.js` - Add Module Federation settings
* `ui/src/setupProxy.js` - Add proxy settings for API calls made by the remote component
* `ui/src/utils/remoteComponent.js`, `ui/src/hooks/useDynamicScript.js` - Make it possible to load the remote module dynamically, as opposed to specifying the experiment engine remote in the Craco configs (ref: [module-federation-examples](https://github.com/module-federation/module-federation-examples/blob/master/dynamic-system-host)).
* `ui/src/experiment/ExperimentsRouter.js` - Dynamically load the experiment engine's landing page or show an appropriate message if it is not configured or if the load fails.
Screenshot 2021-09-28 at 12 00 54 PM
* `ui/src/App.js` - Register the `/experiments` route.
* `ui/src/bootstrap.js` - Moved the contents of `ui/src/index.js` here to let it be loaded asynchronously.

The other changes in this PR are automatic formatting fixes.

### References
* [Create React App 5 Alpha](https://github.com/facebook/create-react-app/discussions/11278)
* [Module Federation support with CRA5](https://github.com/facebook/create-react-app/issues/9510)


ui/src/setupProxy.js
```diff
@@ -1,28 +1,51 @@
-const proxy = require("http-proxy-middleware");
+const { createProxyMiddleware } = require("http-proxy-middleware");
```
krithika369, line 1: `http-proxy-middleware` dependency version has been upgraded in the new CRA.

ui/.env.development
```diff
@@ -14,4 +14,13 @@ REACT_APP_USER_DOCS_URL=[{"href":"https://github.com/gojek/turing","label":"Turi
...
+# Default Experiment Engine Remote UI settings
+REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_APPNAME=
+REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_URL=
+
 REACT_APP_OAUTH_CLIENT_ID=*** set clientId ***
+
+# JSON string map, key being the (unroutable) path used by the remote app for API calls,
+# and value being the URL to proxy to. Used only in local development mode, to bypass CORS.
+# More info: https://create-react-app.dev/docs/proxying-api-requests-in-development/
+REMOTE_COMPONENTS_PROXY_PATHS=
```
krithika369, line 26: Used by `setupProxy.js` for local development. `REACT_APP_` prefix is not required here.

ui/.env.development
```diff
@@ -14,4 +14,13 @@ REACT_APP_USER_DOCS_URL=[{"href":"https://github.com/gojek/turing","label":"Turi
 # Link to the .proto file used by the Kafka Result Logger when using Protobuf serialization
 REACT_APP_RESULT_LOG_PROTO_URL=
 
+# Default Experiment Engine Remote UI settings
+REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_APPNAME=
+REACT_APP_DEFAULT_EXPERIMENT_ENGINE_REMOTE_URL=
```
romanwozniak, line 19: Since these two variables needs to be defined simultaneously (either both or none), they can be combined into a single env variable defined as JSON object, similar to `REACT_APP_USER_DOCS_URL`.

ui/src/bootstrap.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
romanwozniak, line 1: Is it necessarily to rename `index.js`?

ui/src/experiment/ExperimentsRouter.js
```diff
@@ -0,0 +1,55 @@
...
+import { PageTitle } from "../components/page/PageTitle";
+
+import useDynamicScript from "../hooks/useDynamicScript";
+import loadComponent from "../utils/remoteComponent";
+import { defaultExperimentEngine } from "../config";
+
+export const ExperimentsRouter = ({ projectId }) => {
+  // Retrieve script from host dynamically
+  const { ready, failed } = useDynamicScript({
+    url: defaultExperimentEngine.url,
```
romanwozniak, line 19: What if  `defaultExperimentEngine` is `undefined`?

ui/src/experiment/ExperimentsRouter.js
```diff
@@ -0,0 +1,55 @@
...
+                : "Failed to load Experiment Exgine"}
+            </EuiText>
+          </EuiPageContent>
+        </EuiPageBody>
+      </EuiPage>
+    );
+  }
+
+  // Load component from remote host
+  const Component = React.lazy(
```
romanwozniak, line 46: can it be called `ExperimentsLandingPage` then?

ui/src/experiment/ExperimentsRouter.js
```diff
@@ -0,0 +1,55 @@
...
+    loadComponent(defaultExperimentEngine.appName, "./ExperimentsLandingPage")
+  );
+
+  return (
+    <React.Suspense fallback="Loading Experiments">
```
romanwozniak, line 51: What's the difference between this `fallback="Loading Experiments"` and `"Loading Experiment Exgine ..."` at L:36? In what scenarios each of them will be shown?

ui/src/experiment/ExperimentsRouter.js
```diff
@@ -0,0 +1,55 @@
...
+      <Component projectId={projectId} />
+    </React.Suspense>
+  );
```
romanwozniak, line 54:
```suggestion
  // Load component from remote host
  const ExperimentsLandingPage = React.lazy(
    loadComponent(defaultExperimentEngine.appName, "./ExperimentsLandingPage")
  );
  
  return ready 
    ? (
      <React.Suspense fallback="Loading Experiments">
        <ExperimentsLandingPage projectId={projectId} />
      </React.Suspense>
    ) : (
      <EuiPage>
        <EuiPageBody>
          <EuiPageHeader>
            <EuiPageHeaderSection>
              <PageTitle title="Experiments" />
            </EuiPageHeaderSection>
          </EuiPageHeader>
          <EuiPageContent>
            <EuiText size="s" color="subdued">
              { failed 
                ? (!!defaultExperimentEngine.url 
                     ? "Failed to load Experiment Exgine" 
                     : "No default Experiment Engine configured"
                ) : "Loading Experiment Exgine ..."
              }              
            </EuiText>
          </EuiPageContent>
        </EuiPageBody>
      </EuiPage>
    );
```

romanwozniak: But more specifically, I wonder if this component should be broke down into two:
1) When `defaultExperimentEngine.url` is not configured at all then we don't need to try to load something from remote host and just render `"No default Experiment Engine configured"` msg
2) When `defaultExperimentEngine.url` is configured, then `useDynamicScript` and `loadComponent` are called and proper remote component is rendered.

romanwozniak (commented):
Also, your changes introducing many formatting changes, which makes the review process less pleasant. We use `prettier` for having a shared formatting configuration (it's defined [here](https://github.com/gojek/turing/blob/992d465ceb24a9f7e84a7c3346bed982a58862c6/ui/package.json#L43)), and formatting is triggered with husky's pre-commit hook. 

Seems like you are using a newer version of `prettier` than I do and `--jsx-bracket-same-line` option is [deprecated since v2.40](https://prettier.io/docs/en/options.html#deprecated-jsx-brackets), hence the discrepancy. Can you please replace `--jsx-bracket-same-line` in `package.json` with `--bracket-same-line` and reformat your branch with:
```bash
yarn prettier "src/**/*.{js,jsx,ts,tsx,json,css,scss,md}" --bracket-same-line --write
```

ui/.gitignore
```diff
@@ -15,3 +15,6 @@ yarn-error.log*
 
 .env.development.local
 /.eslintcache
+
+# linting
+.husky/
```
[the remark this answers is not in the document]
krithika369: I realise now there is a `.gitignore` inside this generated folder, so it's unnecessary to mention it here - removed.

ui/src/bootstrap.js
krithika369: Yes, this is required when Module Federation is being used, so `bootstrap.js` is loaded asynchronously which allows the host app to wait for the remote components.

ui/src/experiment/ExperimentsRouter.js
krithika369: `defaultExperimentEngine` will at least be `{appName: undefined, url: undefined}`. Anyway, refactored this file based on your other comments, so this is no longer relevant.

ui/src/experiment/ExperimentsRouter.js
krithika369: `Loading Experiment Engine` happens when the `remoteEntry.js` file is being retrieved and appended to the html head.  This part is now required because we are not setting the remote module's configurations in `craco.config.js` at build time anymore.

`fallback="Loading Experiments"` or later for the router form,  may be `fallback="Loading Experiment Configuration"` is a placeholder while React lazy-loads the component (using `loadComponent` in the preceding lines). It's hardly noticeable but still required - without the `fallback` property, the rendering will fail in the first attempt.

Refactored the file to use `<FallbackView/>` for all scenarios where the remote component is not successfully loaded.

ui/src/experiment/ExperimentsRouter.js
krithika369: Thanks for the suggestion. Refactored this file.

krithika369:
@romanwozniak thanks for the review. I've addressed your comments. Could you have another look?

romanwozniak (approved):

PR #102, opened 2021-09-28
Make it possible to configure Turing UI dynamically after static build is done

romanwozniak:
## Problem:
Currently, Turing UI configuration consists of a mix of hard-coded values (defined at [`config.js`](https://github.com/gojek/turing/blob/main/ui/src/config.js)) and a set of `REACT_APP_*` environment variables, that are supplied during the build time. This approach implies that we need to re-build the entire application just to change some configuration keys. It also forces us to release separate builds of Turing for different environments (i.e. dev/staging/production) because the configuration of different environments is naturally expected to be different. 

## Solution:
This PR introduces necessary changes to support overriding Turing UI settings at the run time (after the application is bundled as an optimized static release). This is achieved by including `app.config.js` into `index.html` and merging the runtime config (defined at `app.config.js`) with build-time config (defined at [`config.js`](https://github.com/gojek/turing/blob/main/ui/src/config.js)).
So, if the user wants to override Sentry DSN in Turing's static build, she can update `app.config.js` file in `dist` directory with a desired value like this:

```js
var config = {
    sentryConfig: {
        dsn: "my-new-dsn"
    }
};
```

## Changes:
The main change in the codebase is that config objects (i.e. `appConfig`, `sentryConfig` etc) should no longer be imported directly into components where they are used, but instead, required configuration should be retrieved with a custom `useConfig` React hook. This hook merges user-provided configuration with the configuration defined during the build.

### DIFF
--- a/ui/src/config.js
+++ b/ui/src/config.js
@@ -1,3 +1,4 @@
+import React from "react";
 import objectAssignDeep from "object-assign-deep";
 
 /*
@@ -87,8 +88,15 @@ const buildTimeConfig = {
   alertConfig,
 };
 
-export const useConfig = () => {
-  const runTimeConfig = window.config;
+const ConfigContext = React.createContext({});
 
-  return objectAssignDeep({}, buildTimeConfig, runTimeConfig);
+export const ConfigProvider = ({ children }) => {
+  const runTimeConfig = window.config;
+  const config = objectAssignDeep({}, buildTimeConfig, runTimeConfig);
+
+  return (
+    <ConfigContext.Provider value={config}>{children}</ConfigContext.Provider>
+  );
 };
+
+export const useConfig = () => React.useContext(ConfigContext);
--- a/ui/src/config.js
+++ b/ui/src/config.js
@@ -1,3 +1,5 @@
+import objectAssignDeep from "object-assign-deep";
+
 /*
  * In development environment, we set turingApiUrl, merlinApiUrl and mlpApiUrl
  * to unreachable paths so that the requests will be made to the given API
@@ -7,8 +9,8 @@
  * to the UI if the API is served from the same host. When the API's origin differs from
  * that of the UI, appropriate CORS policies are expected to be in place on the API server.
  */
-export const apiConfig = {
-  apiTimeout: process.env.REACT_APP_API_TIMEOUT || 5000,
+const apiConfig = {
+  apiTimeout: process.env.REACT_APP_API_TIMEOUT,
   useMockData: process.env.REACT_APP_USE_MOCK_DATA || false,
   turingApiUrl:
     process.env.NODE_ENV === "development"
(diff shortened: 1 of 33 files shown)


krithika369:
> So, if the user wants to override Sentry DSN in Turing's static build

Is that a good idea? It seems to me that this method makes it easy to expose sensitive information.

For our internal use case, perhaps the specific example of Sentry DSN can be set at build time because it's consistent across the different environments. But `REACT_APP_OAUTH_CLIENT_ID` varies and I'm not sure `app.config.js` is the place for it. How would they be handled?

Recently, you had also brought up fetching UI configs from the API. How does that work alongside this approach?

romanwozniak:
> > So, if the user wants to override Sentry DSN in Turing's static build
> 
> Is that a good idea? It seems to me that this method makes it easy to expose sensitive information.

It's not any different from the current approach. All the data from `REACT_APP_*` variables is embedded in the build (you can look for it in bundle js files) and is publicly accessible. There is no safe way of securely storing secrets on the client applications that are running on a browser.

> For our internal use case, perhaps the specific example of Sentry DSN can be set at build time because it's consistent across the different environments. But `REACT_APP_OAUTH_CLIENT_ID` varies and I'm not sure `app.config.js` is the place for it. How would they be handled?

Sentry DSN is not a secret (it's just a sentry project identifier). Neither `oauthClientId` is. 

> Recently, you had also brought up fetching UI configs from the API. How does that work alongside this approach?

Fetching configs from API can still be implemented (then `useConfig` hooks needs to be updated to also incorporate configuration retrieved from the API), but now I'm less eager to introduce `/settings` API, because the primary motivation for it was the ability to configure the application at run-time (which is already addressed by this PR).

ui/src/config.js
```diff
@@ -63,16 +61,34 @@ export const appConfig = {
...
+  apiConfig,
+  authConfig,
+  appConfig,
+  sentryConfig,
+  resultLoggingConfig,
+  alertConfig,
+};
+
+export const useConfig = () => {
+  const runTimeConfig = window.config;
```
krithika369, line 91: What is this sorcery? Where are we saving the dynamic config from `app.config.js`  to the window?

krithika369 (approved):
LGTM!

TIL many things.

ui/src/config.js
romanwozniak: `config` is a global variable defined in `app.config.js` and any global variables can be accessed as properties of `window` object. ([See More](https://developer.mozilla.org/en-US/docs/Glossary/Global_object#window_object_in_the_browser))

ui/src/config.js
```diff
@@ -63,16 +61,34 @@ export const appConfig = {
...
+const buildTimeConfig = {
+  apiConfig,
+  authConfig,
+  appConfig,
+  sentryConfig,
+  resultLoggingConfig,
+  alertConfig,
+};
+
+export const useConfig = () => {
```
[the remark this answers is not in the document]
romanwozniak: Thanks for suggestion, this makes sense. I addressed it here: https://github.com/gojek/turing/pull/102/commits/8c4325fee7cd93349bccf3be299ce9d00e434a65

PR #107, opened 2021-10-12
Fix run id version db migration script.

ashwinath:
There was a mistake assuming that `run_id` could be patched in the db migrations script. This PR fixes and puts the `run_id` schema change in a new file. Tested inside a private kubernetes cluster.


infra/charts/turing/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.1.7
+version: 0.1.8
```
romanwozniak, line 4: Why do you need to change the helm chart version?

ashwinath: the migration scripts are packaged with the helm chart, so if its not bumped, there might be some issues with caching and stuff.

PR #110, opened 2021-10-13
Allow Turing's Kubernetes controller to use in cluster configuration as well as setting up of RBAC in Helm charts.

ashwinath:
As part of the efforts to get users to deploy Turing easily, we have removed the need for user generated CA certificates to authenticate with the Kubernetes master. This simplifies the process as we remove vault as a dependency, as well as manual efforts to generate the CA certificate for a new user to install this. Eventually, we are aiming to simplify this process even further by including a helm chart to provision the necessary infrastructure to allow Turing to run.

For GCP, the Kubernetes cluster must be destroyed and recreated if you did not generate a client certificate, which might be troublesome for the user who is deploying Turing.

This PR adds the following:

- RBAC for in cluster Kubernetes controller, Turing can now be used within the cluster without the need for authentication details.
- Helm chart updates to toggle between in cluster configuration and user CA certificates.
- Turing API configuration changes to allow between credentials from Vault and in cluster certificates.

This has been tested using an existing Kubernetes cluster with all the infrastructure set up (Spark, Istio, Knative). The next few PRs will include the installation of Spark controller, Istio and Knative.


infra/charts/turing/values.yaml
```diff
@@ -99,6 +99,11 @@ turing:
   # # Example
   # - name: varlog
   #   mountPath: /var/log
+  clusterConfig:
+    # -- Configuration to use in cluster configuration or authenticate kubernetes with vault configuration
+    useInClusterConfig: false
+    # -- Service account name for turing's kubernetes controller
+    serviceAccountName: turing
```
romanwozniak, line 106: Don't have strong opinion on this, but why do we hardcode it, instead of using `{{ .Release.Namespace }}-svc-acc`? This can be added into `_helpers.tpl` as:
```
{{- define "turing.service-account" -}}
...
{{- end -}}
```

infra/charts/turing/README.md
```diff
@@ -14,22 +20,27 @@ Turing: ML Experimentation System
...
 | turing.livenessProbe.path | string | `"/v1/internal/live"` | HTTP path for liveness check |
 | turing.readinessProbe.path | string | `"/v1/internal/ready"` | HTTP path for readiness check |
-| turing.resources | object | `{}` | Resources requests and limits for Turing API. This should be set  according to your cluster capacity and service level objectives. Reference: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ |
+| turing.resources | object | `{}` | Resources requests and limits for Turing API. This should be set according to your cluster capacity and service level objectives. Reference: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/ |
 | turing.service.externalPort | int | `8080` | Turing API Kubernetes service port number |
 | turing.service.internalPort | int | `8080` | Turing API container port number |
+| turing.uiConfig | object | `{}` | Turing UI configuration. Refer to https://github.com/gojek/turing/blob/main/ui/public/app.config.js |
```
romanwozniak, line 43: Thanks, I missed this

api/turing/config/config_test.go
```diff
@@ -693,13 +707,6 @@ func TestConfigValidate(t *testing.T) {
 			},
 			wantErr: true,
 		},
-		"missing vault address": {
-			validConfigUpdate: func(validConfig Config) Config {
-				validConfig.VaultConfig.Address = ""
-				return validConfig
-			},
-			wantErr: true,
-		},
```
romanwozniak, line 702: Would it be better to update/add a test case for cluster config instead just removing existing scenario?

api/turing/config/config.go
```diff
@@ -285,10 +285,18 @@ type AuthorizationConfig struct {
...
+	// and has the relevant credentials to handle all cluster operations.
+	InClusterConfig bool
+	// VaultConfig is required if InClusterConfig is false.
+	VaultConfig *VaultConfig
+}
+
 // VaultConfig captures the config for connecting to the Vault server
 type VaultConfig struct {
-	Address string `validate:"required"`
-	Token   string `validate:"required"`
```
romanwozniak, line 291: Maybe it still is `required`, but `VaultConfig` itself is not `required` anymore. I.e. `VaultConfig` (line 293) can be `nil`, but if it's not nil, then `Address`/`Token` are `required`

infra/charts/turing/values.yaml
ashwinath: sounds good actually, I never knew this existed.

api/turing/config/config_test.go
ashwinath: I added more test cases with the `required_if` tag.

PR #112, opened 2021-10-27
Install optional infrastructure required to run Turing.

ashwinath:
This PR adds the installation of all the infrastructure required by Turing. It installs the following software onto the Kubernetes cluster:

1. [Spark Operator](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator)
2. [Istio](https://istio.io/)
3. [Knative Serving](https://knative.dev/docs/serving/)

For the spark operator, it installs the charts along with the `turingcluster` charts as if it was a dependency, nothing special. As for the version chosen, the latest works so I have decided to go with the latest version available.

However for the other 2, Istio and Knative serving, there are no charts publically available. So it uses the cluster-init container to install everything. For Istio, it downloads the charts using the recommended bash script recommended [here](https://istio.io/latest/docs/setup/getting-started/#download). For Knative, it installs the yaml files provided by the [official souce](https://knative.dev/docs/admin/install/serving/install-serving-with-yaml/).

Cleanup might be complicated for the end user, but I have included a pre-delete helm hook to clean up all resources used by this install chart if the user wants to delete this set of infrastructure. With the Helm pre delete hooks, the administrator would just have to issue the command `helm delete <release name> --timeout 15m` and it will be good to go. Deleting Knative will take some time and it is required that the timeout is set to an explicitly long duration. From my tests it seems like 15 minutes is a good timeout duration.


infra/cluster-init/init.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "    KNATIVE_ISTIO_VERSION             Knative Istio version, default: 0.18.1."
+}
+
+function install_istio {
+    echo "Installing Istio."
+
+    local istio_version=${ISTIO_VERSION:-1.9.9}
+    curl -L https://istio.io/downloadIstio | ISTIO_VERSION=${istio_version} TARGET_ARCH=x86_64 sh -
+    kubectl create namespace istio-system --dry-run=client -o yaml | kubectl apply -f -
+    helm install istio-base "istio-${istio_version}/manifests/charts/base" \
```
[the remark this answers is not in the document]
ashwinath: I'm caving in and have swapped it to istioctl. Reason being you can't really set global values across all the charts easily because there is no "uber" istio chart.

ashwinath:
Done with the E2E tests and Readme updates. We should probably split local development and production environment setups once we are done with the full Turing installation.

.github/workflows/turing.yaml
```diff
@@ -12,9 +12,11 @@ on:
       - "docs/**"
       - "engines/batch-ensembler/**"
       - "sdk/**"
+      - "infra/cluster-init/**"
```
romanwozniak, line 15: Since `cluster-init/**` is involved in this workflow (you are building its docker image for the e2e test), I think that it should not be excluded here.

romanwozniak: Also, it's expected that `path-ignore` at L11 is the same as `path-ignore` at L25 (same criteria for `push` and `pull_request`)

.github/workflows/turing.yaml
```diff
@@ -403,6 +405,47 @@ jobs:
...
+            --set image.registry=${{ env.LOCAL_REGISTRY }}/ \
+            --set image.repository=${{ github.repository }}/cluster-init \
+            --set image.tag=${{ env.CLUSTER_INIT_VERSION }} \
+            --install \
+            --wait
+
+          # wait for install infra job to finish
+          kubectl logs -n infrastructure -f $(kubectl get po -n infrastructure | grep -v 'NAME' | grep -v 'spark' | head -n 1 | awk '{print $1}')
+          kubectl get po -A
+          kubectl get svc -A
```
romanwozniak, line 428: Is it for debug only ? Can be removed now?

.github/workflows/turing.yaml
```diff
@@ -476,7 +521,8 @@ jobs:
             --set turing.image.repository=${{ github.repository }}/turing-api \
             --set turing.image.tag=${{ env.TURING_API_VERSION }} \
             --set turing.config.RouterDefaults.Image=${{ env.LOCAL_REGISTRY }}/${{ github.repository }}/turing-router:${{ env.TURING_ROUTER_VERSION }} \
-            --values infra/e2e/turing.helm-values.yaml
+            --values infra/e2e/turing.helm-values.yaml \
+            --wait
```
romanwozniak, line 525: Since you made everything synchronous  (with `--wait` added to helm install commands), there is very little point in `wait-for-deployment` step. Not sure if we want to keep it

.github/workflows/turing.yaml
```diff
@@ -543,6 +589,12 @@ jobs:
           KUBECONFIG_USE_LOCAL: true
         run: go test -v -parallel=2 ./e2e/... -tags=e2e -run TestEndToEnd
 
+      # Invoke helm hooks on delete
+      - name: Tear down infrastructure job
+        run: |
+          helm delete -n infrastructure turing-init --timeout 15m
+          kubectl get po -A
```
romanwozniak, line 596: If it's a code, that's expected to be read and maintained by multiple people, it's advised to use the long form of flags for better readability, i.e 
```
kubectl get pods --all-namespaces
```

README.md
```diff
@@ -48,9 +48,12 @@ in these getting started guide.
 > you will need to deploy the infrastructure using an approach that ensures
 > the services are scalable and reliable. For instance, using cloud provider
 > services such as Google Kubernetes Engine (GKE) for provisioning the Kubernetes cluster.
+> See "Installing prerequisites on Kubernetes using the init chart" section for a recommended production installation guide.
 >
 > The following guides are tested on Linux and MacOS.
 
+For a Kubernetes based setup, see
```
romanwozniak, line 55: Incomplete statement?

README.md
```diff
@@ -333,6 +336,32 @@ web browser.
...
+To install the required components on your Kubernetes cluster, issue the following command:
+
+```bash
+helm upgrade -n infrastructure turing-init infra/charts/turing-init \
+    --set image.registry="<image registry>/" \
+    --set image.repository="<repository>" \
+    --set image.tag="<tag version>" \
+    --install \
+    --wait
+```
```
romanwozniak, line 356: Can we expand this section a bit and include the success criteria. I.e. what resources users expect to see in the cluster after provisioning is successful.

romanwozniak (approved):
Please check the comments I left, but it LGTM

.github/workflows/turing.yaml
ashwinath: I think it's good to have information whenever things go wrong, so that it won't be a black box.

README.md
ashwinath: ah, good catch, i edited the top line but forgot to remove this.

.github/workflows/turing.yaml
ashwinath: Yeah I'll remove this. It's hard to tell which step it failed at if you do the wait for all to complete.

.github/workflows/turing.yaml
ashwinath: Oh I see. Removing `cluster-init/**` from `paths-ignore`

.github/workflows/turing.yaml
ashwinath: ok, done

README.md
ashwinath: Ya ok

PR #116, opened 2021-11-01
Display router timeout if available

krithika369:
This PR addresses edge cases from https://github.com/gojek/turing/pull/115

* Show the router timeout value only if available (new routers would not have a `config` property)
* Show "Not available" in place of router timeout if it is currently updating (rationale: when the router is updating, it is not clear if this value belongs to the new config or old config).

### DIFF
--- a/ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
+++ b/ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
@@ -1,13 +1,15 @@
 import React from "react";
-import { EuiBadge, EuiText } from "@elastic/eui";
+import { EuiBadge, EuiTextColor } from "@elastic/eui";
 import { DateFromNow } from "@gojek/mlp-ui";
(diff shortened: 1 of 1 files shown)


ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
```diff
@@ -24,7 +26,11 @@ export const RouterDetailsPageHeader = ({ router }) => {
     },
     {
       title: "Timeout",
-      description: <EuiText size="s">{router.config.timeout}</EuiText>,
+      description: (
+        <EuiTextColor size="s" color={isRouterUpdating ? "subdued" : "default"}>
```
krithika369, line 30: Using `EuiTextColor` directly instead of `<EuiText size="s"><EuiTextColor color="subdued">...` because with the latter, there is a vertical small alignment issue that I cannot easily identify the fix for - it seems the Timeout's value is placed a few pixels higher than all others in the header section. Eg:

Screenshot 2021-11-01 at 9 28 32 AM

terryyylim (approved):

PR #103, opened 2021-10-06
Serve API docs from Turing API server

romanwozniak:
### Context:
This PR introduces [swagger-ui-generator](https://github.com/gojek/turing/compare/main...romanwozniak:swagger_gen_dist?expand=1#diff-8deb06e38b91d01d9817aeca9816a99a753c59caab600482a1332d9cd1680d65) script together with necessary refactoring to make it possible to serve API docs from turing-api server.

### Changes:
* Added [swagger-ui-generator](https://github.com/romanwozniak/turing/tree/swagger_gen_dist/scripts/swagger-ui-generator) that downloads specified version of static `swagger-ui` distribution and configures it to be used with a local spec file
* Added [Makefile](https://github.com/gojek/turing/compare/main...romanwozniak:swagger_gen_dist?expand=1#diff-86b84087c9937b69f4fc51b0bce8b8b128d604af74850a2190abd84b604c5ccc) to simplify resolving and merging multiple openapi specs into a single file and downloading swagger ui to be used with that bundled spec file.
* Refactored `web.ServeReactApp` into `server.ServeSinglePageApplication` to re-use it for serving both Turing-UI and Swagger-UI
* Refactored `cmd/main.go` to define all the http routes of turing API with `gorilla/mux` (it was previously mix of `gorilla/mux` routers and `http` package routes)
* Updated helm chart to remove swagger-ui container from the deployment

PR #117, opened 2021-11-02
Handle API path prefix in Authorization Middleware, allow templated Experiment Engine Homepage URL

krithika369:
This PR makes some minor modifications.
1. In an earlier [PR](https://github.com/gojek/turing/pull/103), the set up of the HTTP handler was refactored. The usage of the [Subrouter](https://pkg.go.dev/github.com/gorilla/mux#Route.Subrouter) with [PathPrefix](https://pkg.go.dev/github.com/gorilla/mux#Router.PathPrefix), however, broke the expectation within the Authorization Middleware, causing the resources to be formatted as `v1:projects:{id}:*` as opposed to `projects:{id}:*`. This MR propagates the API handler's path prefix to the Authorization middleware upon initialisation, so it can be ignored at the time of validation.
2. The UI now allows a templated string to be supplied by the experiment engine, for its `home_page_url` property so that the template `{{projectId}}` in the string would be replaced by the current MLP project ID. This is useful for the internal Experiment Engine used in Gojek.

### DIFF
--- a/api/turing/api/appcontext.go
+++ b/api/turing/api/appcontext.go
@@ -43,6 +43,7 @@ type AppContext struct {
 func NewAppContext(
 	db *gorm.DB,
 	cfg *config.Config,
+	apiPathPrefix string,
 	authEnforcer *enforcer.Enforcer,
 	vaultClient vault.VaultClient,
 ) (*AppContext, error) {
@@ -50,7 +51,7 @@ func NewAppContext(
 	var authorizer *middleware.Authorizer
 	var err error
 	if authEnforcer != nil {
-		authorizer, err = middleware.NewAuthorizer(*authEnforcer)
+		authorizer, err = middleware.NewAuthorizer(*authEnforcer, apiPathPrefix)
 		if err != nil {
 			return nil, errors.Wrapf(err, "Failed initializing Authorizer")
 		}
--- a/api/turing/api/appcontext.go
+++ b/api/turing/api/appcontext.go
@@ -3,7 +3,6 @@ package api
 import (
 	"fmt"
 
-	"github.com/gojek/mlp/api/pkg/authz/enforcer"
 	"github.com/gojek/mlp/api/pkg/vault"
 	batchensembling "github.com/gojek/turing/api/turing/batch/ensembling"
 	batchrunner "github.com/gojek/turing/api/turing/batch/runner"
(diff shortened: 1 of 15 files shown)


LeonLnj (approved):
LGTM

api/turing/api/appcontext.go
```diff
@@ -43,14 +43,15 @@ type AppContext struct {
 func NewAppContext(
 	db *gorm.DB,
 	cfg *config.Config,
+	apiPathPrefix string,
 	authEnforcer *enforcer.Enforcer,
```
romanwozniak, line 47: Can we instead of passing `apiPathPrefix` here, replace `authEnforcer` with `authorizer`? I.e. authorizer should initialized (if configured) outside of the app context. It doesn't make sense to initialize `authEnforcer` outside, but `authorizer` – inside of the app context, as they are used together.

ui/src/router/components/configuration/components/config.js
```diff
@@ -1,7 +1,16 @@
-const getExperimentUrl = (engineProps, experiment) => {
+import template from "lodash/template";
+import templateSettings from "lodash/templateSettings";
+
+export const getHomePageUrl = (engineProps, projectId) => {
+  // If {{projectId}} present in the URL, substitute with actual project id.
+  templateSettings.interpolate = /{{([\s\S]+?)}}/g;
```
romanwozniak, line 6: What does this regex mean?

ui/src/router/versions/details/RouterVersionDetailsView.js
```diff
@@ -119,7 +119,11 @@ export const RouterVersionDetailsView = ({
 
             <Router primary={false}>
               <Redirect from="/" to="details" noThrow />
-              <RouterVersionConfigView path="details" config={version} />
+              <RouterVersionConfigView
+                path="details"
+                projectId={projectId}
```
romanwozniak, line 124: You likely don't need this `projectId` here, as `@reach/router` will pass it for you

romanwozniak (approved):
Left some minor questions/suggestions, but it LGTM

ui/src/router/components/configuration/components/config.js
krithika369: I copied this regex from what was previously added for the client-side formatting of the router Monitoring URL. It's meant to identify variables (which may have whitespaces in between) surrounded by double curly braces `{{ }}`.

In this case, my intention was to match `{{projectId}}` in the URL. As of now, it is the only variable in this scenario.

krithika369:
Thanks, both, for the quick review! 🙏🏽

PR #118, opened 2021-11-08
Implement Experiment Manager Types

krithika369:
This PR introduces the concept of Standard vs Custom Experiment Managers. Accordingly, the `ExperimentManager` interface has been refactored and 2 new interfaces introduced - `StandardExperimentManager` and `CustomExperimentManager `. Their differences are detailed in the [Experiment Engine Developer Guide](https://github.com/krithika369/turing/blob/experiment_engine_interface/engines/experiment/docs/developer_guide.md).

### Experiment Engine
The main change is the separation of the `ExperimentManager` interface into standard and custom.
* The standard managers are expected to implement a larger set of APIs to take advantage of the inbuilt experiment UI in Turing. So that the standard managers don't have to implement methods that they don't intend to use, a `BaseStandardExperimentManager` type has been provided with default implementations for all the optional methods. Most of the concepts and the data types (like `TuringExperimentConfig`) previously defined for Experiment Engines in Turing are now only applicable to the standard managers.
* The custom managers are expected to expose a custom UI and therefore, their API contracts are loosely defined (eg: `interface{}` in place of `TuringExperimentConfig`). 

Modifications:
* `engines/experiment/docs` - Updated user docs
* `engines/experiment/manager/manager.go`, `engines/experiment/manager/types.go` - New interfaces and data structure definitions related to the experiment manager types
* `engines/experiment/manager/base_standard_manager.go` - Previously `base_manager.go`, implementing all (and only) the optional methods in the `StandardExperimentManager` interface.
* `engines/experiment/manager/adapter.go` - Public convenience methods for accessing the specific methods in the experiment manager, based on its type. Handles the casting and coercing of values.

### API
* `api/api/specs/experiment-engines.yaml` - Update the experiment config and experiment engine config to reflect standard vs custom configs
* `api/turing/api/request/request.go`, `api/turing/api/deployment_controller.go` - When creating / updating / deploying routers with experiments, do special handling of passkey in standard configs only (there is no visibility to Turing about what info custom configs hold)
* `api/turing/models/experiment_engine.go` - Retrieve experiment config data in DB as `interface{}`. Will be cast to `TuringExperimentConfig` for standard experiment engines, where required.
* `api/turing/service/experiment_service.go` - Add utility methods `IsStandardExperimentManager` and `GetStandardExperimentConfig`. All experiment manager methods will now be called using the adapter functions defined in the experiments package.

### UI
* Adopt `standard_experiment_manager_config` everywhere in the standard experiment UI
* `ui/src/router/components/form/components/experiment_config/ExperimentConfigPanel.js` - Load custom experiment config form UI if the Experiment Manager is custom
* `ui/src/router/components/configuration/components/ExperimentConfigSection.js` - Load custom experiment info UI if the Experiment Manager is custom
* Refactor experiment engine workflows so that the engine props may be retrieved from the `ExperimentEngineContextProvider` where needed (eg: in `yup` validation context or for Version comparison view), rather than relying on the saved props. This helps retrieve the Experiment Manager type (standard vs custom) on the fly and conditionally process the experiment config.
* `ui/src/providers/experiments/ExperimentEngineContextProvider.js` - Parse custom experiment engine validation schema. Refactor experiment schema validation logic in `ui/src/router/components/form/validation/schema.js`.
* `ui/src/hooks/useRemoteComponent.js` - Add hook to memoize and return remote component (especially in forms, we do not want the component to be refetched and re-rendered on every change).

Failure to load the remote component is handled appropriately:
Screenshot 2021-11-09 at 8 34 16 AM


api/turing/api/deployment_controller.go
```diff
@@ -188,10 +188,16 @@ func (c RouterDeploymentController) deployRouterVersion(
 		}
 	}
 
-	if routerVersion.ExperimentEngine.Type != models.ExperimentEngineTypeNop {
+	expSvc := c.BaseController.AppContext.ExperimentsService
+	if expSvc.IsStandardExperimentManager(string(routerVersion.ExperimentEngine.Type)) {
```
krithika369, line 192: Parsing passkey is only applicable to Standard Experiment Engines' configs.

api/turing/validation/validator.go
```diff
@@ -101,22 +100,14 @@ func newExperimentConfigValidator(expSvc service.ExperimentsService) func(valida
 		case string(models.ExperimentEngineTypeNop):
 			return
 		case string(models.ExperimentEngineTypeLitmus), string(models.ExperimentEngineTypeXp):
```
krithika369, line 102: We still have some tech debt in this repo, explicitly mentioning the internal exp engines in many places (including unit tests). But the concerning part is this validation which will prevent creation of experiments for other new experiment engine plugins. Will add this cleanup to the backlog.

.github/workflows/turing.yaml
```diff
@@ -236,8 +236,7 @@ jobs:
         with:
           path: api
           key: |
-            ${{ runner.os }}-vendor-${{ hashFiles('api/go.mod') }}
```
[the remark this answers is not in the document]
krithika369: Also, I'm having trouble identifying what the hash corresponds to. I tried stripping down the [hash logic](https://github.com/actions/runner/blob/main/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts) here but the hex doesn't match:

https://codesandbox.io/s/recursing-raman-emy7f?file=/index.js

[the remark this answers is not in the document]
ashwinath: I don't understand how the caching works. It builds in the container right? Does it pull the vendor outside the container then builds it in the container?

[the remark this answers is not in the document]
ashwinath: try this? https://github.community/t/hashfiles-with-multiple-patterns/17168/3 hash the router and experiment folders as well

engines/experiment/docs/concepts.md
```diff
@@ -4,26 +4,16 @@
...
-* **Client Info** - If the experiment engine supports basic auth / authorization, this specifies the credentials used to access the APIs, to generate the treatment.
-* **Experiment(s)** - One or more experiments that must be run for every request.
-* **Variables** - Parameters for running the experiment(s), to be retrieved from the request header / payload.
-
-Turing users will configure these properties through the Turing app, when they set up their routers. More details may be found under [Configure Experiment Engine](../../../docs/markdowns/creating-router/configure_rules_engine.md).
+When creating a router, Turing users may choose to configure an experiment on an available experiment engine (the [Developer Guide](./developer_guide.md) has info on adding experiment engines to Turing). When the router is deployed, with every incoming request to the router, a call will be made to the experiment engine using the following information.
```
ashwinath, line 15: Nitpick:

```suggestion
When creating a router, Turing users may choose to configure an experiment on an available experiment engine (the [Developer Guide](./developer_guide.md) has info on adding experiment engines to Turing). When the router is deployed, with every incoming request to the router, a call will be made to the experiment engine with the following information:
```

engines/experiment/docs/developer_guide.md
```diff
@@ -20,17 +31,59 @@ The primary data types associated with the management of experiments may be visu
...
+##### API
+The experiment managers are required to implement the methods in the `CustomExperimentManager` interface. This is a lean interface with methods critical to the interaction between the Turing app and the router.
 
-### Turing UI
+##### UI
+The custom experiment manager should provide the following UI components for edit and view experiment engine configuration. The remote UI information should be set in `CustomExperimentManagerConfig` in the Experiment Engine properties.
 
-The responses from the experiment manager APIs will be used to populate the UI components for Client, Experiment and Variables configuration. For example, the client selection panel will only be displayed if `client_selection_enabled` is set to true in the Experiment Engine properties.
\ No newline at end of file
+```javascript
```
ashwinath, line 57: typescript?

engines/experiment/manager/types.go
```diff
@@ -23,6 +24,33 @@ type Engine struct {
...
 
+type RemoteUI struct {
+	// Name is the name of the remote app declared in the Module Federation plugin
+	Name string `json:"name"`
+	// URL is the Host + Remote Entry file at which the remote UI can be found
+	URL string `json:"url"`
+}
+
+type CustomExperimentManagerConfig struct {
+	// RemoteUI specifies the information for the custom experiment engine UI to be
```
terryyylim, line 35: Is a UI implementation compulsory for `CustomExperimentManager`, or do we just expect the screenshot you've attached in the PR description?

.github/workflows/turing.yaml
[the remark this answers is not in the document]
romanwozniak: I think all you need to do is as suggested by @ashwinath, to modify`key` to include hashes of internal `api`'s dependencies (engines/experiment and engines/router). Something like
`${{ runner.os }}-vendor-${{ hashFiles('api/go.mod') }}-${{ hashFiles('engines/experiment/go.mod') }}-${{ hashFiles('engines/router/go.mod') }}`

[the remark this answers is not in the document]
romanwozniak: As regarding why do we need `restore-keys`, the logic of it is well described in the docs (https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key), but to put it into context, let me illustrate this by this example:
1. You have your application with v1 of dependencies. After successfully running the pipeline for the first time, it will create a cache for your dependencies under the `ubuntu-vendor-<v1-god.mod-hash>`.
2. Then, you updated your `go.mod` to add/update/delete new dependencies, so your workflows will they to look for the cache ander `ubuntu-vendor-<v2-god.mod-hash>` which doesn't exist yet.
3. If you have your `restore-keys` set to `ubuntu-vendor-`, then it partially matches the key of `ubuntu-vendor-<v1-god.mod-hash>`, hence it will use this cache to speed-up your build.
4. Since `ubuntu-vendor-<v1-god.mod-hash>` does not 100% match your dependencies set, you still need to run `go mod vendor` to fetch/update missing dependencies

It all works well with properly versioned public dependencies, however, there could be some quirks if we use `replace` to point experiment/router dependencies into the private repo. I don't know how exactly it works, so some investigation might be required. But maybe if you update versions of experiment/router (it's currently set to `v0.0.0`) to something else, then `go mod vendor` will try to re-fetch it and will update it in the previously fetched cache from the v1 cache archive ()

api/turing/api/deployment_controller.go
```diff
@@ -188,14 +188,23 @@ func (c RouterDeploymentController) deployRouterVersion(
 		}
 	}
 
+	expSvc := c.BaseController.AppContext.ExperimentsService
 	if routerVersion.ExperimentEngine.Type != models.ExperimentEngineTypeNop {
-		// If passkey has been set, decrypt it
-		if routerVersion.ExperimentEngine.Config.Client.Passkey != "" {
-			experimentPasskey, err = c.CryptoService.Decrypt(routerVersion.ExperimentEngine.Config.Client.Passkey)
+		if expSvc.IsStandardExperimentManager(string(routerVersion.ExperimentEngine.Type)) {
```
romanwozniak, line 193: Why do we need to cast it to string here?

api/turing/api/deployment_controller.go
```diff
@@ -188,14 +188,23 @@ func (c RouterDeploymentController) deployRouterVersion(
...
+				experimentPasskey, err = c.CryptoService.Decrypt(expConfig.Client.Passkey)
+				if err != nil {
+					return "", c.updateRouterVersionStatusToFailed(err, routerVersion)
+				}
+			}
 		}
+
 		// Get the deployable Router Config for the experiment
 		experimentConfig, err = c.ExperimentsService.GetExperimentRunnerConfig(
 			string(routerVersion.ExperimentEngine.Type),
```
romanwozniak, line 210: and here

engines/experiment/manager/adapter_test.go
```diff
@@ -0,0 +1,257 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/gojek/turing/engines/experiment/manager"
+	"github.com/gojek/turing/engines/experiment/manager/mocks"
+)
+
+func TestStandardExperimentMethods(t *testing.T) {
```
romanwozniak, line 16: Is there any particular reason why do you merge test cases for all the methods in a single test function? Can we have each function tested separately for easier debugging when something fails?

engines/experiment/manager/adapter_test.go
```diff
@@ -0,0 +1,257 @@
...
+	}`), &testRawConfig)
+	assert.NoError(t, err)
+	testSuccessResponse := json.RawMessage([]byte(`{}`))
+
+	// Set up mock experiment managers
+	// Standard Exp Manager
+	stdExpMgr := &mocks.StandardExperimentManager{}
+	stdExpMgr.On("GetEngineInfo").Return(manager.Engine{Type: manager.StandardExperimentManagerType})
+	stdExpMgr.On("GetExperimentRunnerConfig", testStdExpConfig).Return(testSuccessResponse, nil)
+	stdExpMgr.On("GetExperimentRunnerConfig", mock.Anything).Return(nil, errors.New("Unexpected Parameters"))
```
romanwozniak, line 106: I'm not too sure if it's feasible, but I would prefer to have this (as well as `testRawConfig` and maybe something else I'm missing) moved into fixtures for DRY and separation of test cases from stubs

romanwozniak (approved):
Thanks for the detailed PR description and sorry for the delay on my end in reviewing it. Left some minor comments, but generally it LGTM.

P.S. Could you please split UI changes from the API changes into a separate PR next time. This will for sure make it easier to review :)

.github/workflows/turing.yaml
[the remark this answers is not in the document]
krithika369: >I  think all you need to do is as suggested by @ashwinath, to modify key to include hashes of internal api's dependencies (engines/experiment and engines/router). Something like
${{ runner.os }}-vendor-${{ hashFiles('api/go.mod') }}-${{ hashFiles('engines/experiment/go.mod') }}-${{ hashFiles('engines/router/go.mod') }}

I'm afraid the `go.mod` is not enough. A change in logic in the `engines/experiment` module will not affect its `go.mod` file (or `go.sum`, for that matter) - adding them to the hash pattern will not detect that something changed in this scenario and this actually the case in this PR. But I think @ashwinath's idea of "hash the router and experiment **folders** " may do the trick, but (from what I understand) not if we use `restore-keys` (below).

Thanks for the detailed explanation on `restore-keys` - it is much clearer to me. In step #4, it is expected that the outdated dependencies are re-updated. So, regardless of what cached vendors have been restored, it is supposed to correct itself in the `Build Docker image` step since `make build-image` also does a `go mod vendor`. But, as you've pointed out, this fails to update the local dependency - this is strange to me, I've never encountered this in my local development. This is why I'm not sure if the `build-api` step's cache pattern needs to be changed at all:
* Using the deps' `go.mod` is not exhaustive
* Hashing the deps' folders as Ashwin suggested may be good but it will create a new cache key almost always which can easily hit the max cache size supported by Github.

I will perhaps try to exclude `github.com/gojek/turing` path from the cache

engines/experiment/docs/concepts.md
krithika369: Oh I intentionally put using because "with" makes it look like that info is passed down in the API call. Rather, it may be implicitly made use of in some cases.

engines/experiment/manager/types.go
krithika369: The screenshot in the PR description is for the standard experiment manager. It will only work for standard ones.

> Is a UI implementation compulsory for CustomExperimentManager

Yes. More info here: `engines/experiment/docs/developer_guide.md`

krithika369:
> Thanks for the detailed PR description and sorry for the delay on my end in reviewing it. Left some minor comments, but generally it LGTM.
> 
> P.S. Could you please split UI changes from the API changes into a separate PR next time. This will for sure make it easier to review :)

@roman, Apologies for the painfully long MR. It was my original intention to keep the UI changes minimal in this MR but I ended up refactoring some of the code as I began to test and discover issues. Will be mindful of this in the future.

@gojek/turing-dev, I did make a few more changes in the last few days, mostly refactoring. The main addition is [this](https://github.com/gojek/turing/blob/582b14e61118954a6d3e6c56d24d5aed71717c52/engines/experiment/manager/types.go#L38-L40) - the custom experiment manager is now allowed to expose a serialised `yup` validation schema, so it may be used by Turing to validate the router form. I attempted to keep the validation entirely in the remote component (by passing down `isTouched` from the parent form, to toggle the validation in the child component) but with that approach, the parent (Turing) will have to learn if there was an error in the child and where. So, decided to do it this way which also feels cleaner to me.

.github/workflows/turing.yaml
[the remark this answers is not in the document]
krithika369: Eh wait: https://github.com/gojek/turing/runs/4208336794?check_suite_focus=true#step:4:844

The problem is that the `vendor` data is actually updated but the API code is older. I think it's because we are caching `api` instead of `api/vendor`: https://github.com/gojek/turing/blob/main/.github/workflows/turing.yaml#L238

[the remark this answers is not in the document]
krithika369: Ok, that was it.

api/turing/api/deployment_controller.go
krithika369: The experiment service methods have always worked with strings. Eg: https://github.com/gojek/turing/blob/main/api/turing/service/experiment_service.go#L22

But in the RouterVersion model, this value is of type [models.ExperimentEngineType](https://github.com/gojek/turing/blob/main/api/turing/models/experiment_engine.go#L14). There is some inconsistency with how the value is processed but I am also considering removing this custom type altogether later on, as I feel there is little use for it (or it could be defined in the `experiment/engines` package, with the possible values being dynamically generated based on what engines are registered).

api/turing/api/deployment_controller.go
krithika369: Same: https://github.com/gojek/turing/pull/118#discussion_r749078096

engines/experiment/manager/adapter_test.go
krithika369: I chose to do this because they all have the same set up function (for `GetEngineInfo`).

Separated them out now.

Later, we could integrate: https://github.com/stretchr/testify, which may also help address your other comment: https://github.com/gojek/turing/pull/118#discussion_r748848904

But I will keep this for a future MR.

.github/workflows/turing.yaml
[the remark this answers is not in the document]
ashwinath: > I think all you need to do is as suggested by @ashwinath, to modify`key` to include hashes of internal `api`'s dependencies (engines/experiment and engines/router). Something like `${{ runner.os }}-vendor-${{ hashFiles('api/go.mod') }}-${{ hashFiles('engines/experiment/go.mod') }}-${{ hashFiles('engines/router/go.mod') }}`

I missed this, I think go.mod isn't enough, a digest should be generated based on all the source code as well. I don't think it is an extremely slow process but we can test it out.

PR #119, opened 2021-11-10
Fix validation of toggling batch ensembling

ashwinath:
Issue:
Setting `BatchEnsemblingConfig.Enabled = false` prevents the application from starting up because it will fail the validation. Also, when we disable batch ensembling, there will some nil pointer errors if we try to instantiate EnsemblingJobService, to prevent this, we disable the endpoints for batch ensembling if we disable batch ensembling altogether.

There is one issue with the validator and I have put it as a comment in the code.

Changes made:
1. Updated the validator to use the `required_if` tag.
2. Disable batch ensembling endpoints when the feature is turned off via the config.


api/turing/config/config.go
```diff
@@ -94,10 +94,14 @@ func (c *Config) Validate() error {
...
-	Enabled             bool                `validate:"required"`
-	JobConfig           JobConfig           `validate:"required"`
-	RunnerConfig        RunnerConfig        `validate:"required"`
-	ImageBuildingConfig ImageBuildingConfig `validate:"required"`
+	Enabled bool
+	// For JobConfig, RunnerConfig and ImageBuildingConfig, it is too difficult to enable custom logic
+	// where if it is enabled but all 3 are blank, then throw an error.
+	// This seems like the best effort for validating if enabled is set to false.
+	// Best case is to just detect if the user tried to fill in but filled some components wrongly.
+	JobConfig           *JobConfig           `validate:"omitempty,required_if=Enabled false"`
```
krithika369, line 102: Shouldn't these be `validate:"required,required_if=Enabled true"`, i.e., each of these 3 values are required if `BatchEnsemblingConfig.Enabled=true` ? 

`validate:"omitempty,required_if=Enabled false"` does nothing (i.e., any value is acceptable when `Enabled=true` and empty value will be ignored when `Enabled=false`).

> ... where if it is enabled but all 3 are blank, then throw an error

Not sure I follow the intention. Does this mean it is ok for one or more of the 3 values to be empty but not all 3? I'd assume that if the batch ensembling is enabled, then all 3 should be provided (or that each config is considered optional independently because they have some system defaults). I don't get the part about "all 3 are blank".

api/turing/api/appcontext.go
```diff
@@ -83,26 +83,27 @@ func NewAppContext(
...
 	// Initialise Batch components
 	// Since there is only the default environment, we will not create multiple batch runners.
 	var batchJobRunners []batchrunner.BatchJobRunner
 
+	var ensemblingJobService service.EnsemblingJobService
 	if cfg.BatchEnsemblingConfig.Enabled {
+		// Initialise Ensembling Job Service
+		ensemblingJobService = service.NewEnsemblingJobService(
+			db,
+			cfg.BatchEnsemblingConfig.JobConfig.DefaultEnvironment,
```
krithika369, line 95: The way the validator is written now and with the usage of pointers, `JobConfig` could be `nil` when Enabled=true.

ashwinath: Ok I added some checks for JobConfig, RunnerConfig and ImageBuildingConfig

api/turing/config/config.go
ashwinath: > Shouldn't these be validate:"required,required_if=Enabled true", i.e., each of these 3 values are required if BatchEnsemblingConfig.Enabled=true ?

Unfortunately this doesn't work, seems like a limitation of the library. So what I want is:
Enabled: true then validate JobConfig + RunnerConfig + ImageBuildingConfig
Enabled: false then don't need to validate JobConfig + RunnerConfig + ImageBuildingConfig, because you shoudn't need to set anything if you disabled the feature.

Why I put omitempty is because if let's say the user did intend to fill up JobConfig, RunnerConfig and ImageBuildingConfig but did it partially, it would throw an validation error.

So on this way of setting things, the logic doesnt work as intended in the following scenarios:
```go
c := &BatchEnsemblingConfig {
  // Case 1: batch ensembling disabled but user tried to fill the settings for batch ensembling
  Enabled: false,
  // 1/2 out of 3 of the following are wrong,
  JobConfig: &JobConfig{/* with/without correct settings */},
  RunnerConfig: &RunnerConfig{/* with/without correct settings */},
  ImageBuildingConfig: &ImageBuildingConfig{/* with/without correct settings */},
}

// Case 2: batch ensembling enabled but user tried to fill the settings for batch ensembling (1/2/3 out of 3 of the permutations are wrongly filled in)
// This case will be checked by the playground/validator
c := &BatchEnsemblingConfig {
  Enabled: true,
  // 1/2 out of 3 of the following are wrong
  JobConfig: &JobConfig{/* with/without correct settings */},
  RunnerConfig: &RunnerConfig{/* with/without correct settings */},
  ImageBuildingConfig: &ImageBuildingConfig{/* with/without correct settings */},
}

c := &BatchEnsemblingConfig {
  // Case 3: batch ensembling enabled but user left all configuration nil
  // This case will be checked by the nil pointer checked with this comment: https://github.com/gojek/turing/pull/119#discussion_r747952017
  Enabled: true,
  JobConfig: nil,
  RunnerConfig: nil,
  ImageBuildingConfig: nil,
}
```

Anyway I'll be removing the comments as I think they are confusing.

ashwinath: wait, i just realised required_if=Enabled true and not false. let me try again.

krithika369: I see.. What about: validate:"-,required_if=Enabled false"

https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Skip_Field

api/turing/config/config.go
```diff
@@ -94,10 +94,15 @@ func (c *Config) Validate() error {
 
 // BatchEnsemblingConfig captures the config related to the running of batch runners
 type BatchEnsemblingConfig struct {
-	Enabled             bool                `validate:"required"`
-	JobConfig           JobConfig           `validate:"required"`
-	RunnerConfig        RunnerConfig        `validate:"required"`
-	ImageBuildingConfig ImageBuildingConfig `validate:"required"`
+	// Due to the limitations of the go-playground/validator,
```
krithika369, line 97: Picking up from [this](https://github.com/gojek/turing/pull/119#discussion_r747988031) comment, I think that all 3 validate tags can be changed to `validate:"required_if=Enabled true"`. As you've highlighted, `Enabled=false` will still evaluate any configs that are set and I believe the `StructFiltered` should be the way around it.

I feel it's not critical to fix, so if it will take some time to address, we could add a TODO here to summarize the plan for the fix and move on for now.

api/turing/config/config.go
ashwinath: I tried using the method the base_manager used, it doesn't seem to work as intended as well. It works the same way as `validate:"required_if=Enabled true"`.

```go
func (c *Config) Validate() error {
	validate, err := newConfigValidator()
	if err != nil {
		return err
	}
	// Check general flow first
	if err := validate.Struct(c); err != nil {
		fmt.Println("here") // Here it hits when Enabled=false, but other settings filled halfway
		return err
	}

	// Extra rules go here
	return c.validateBatchEnsemblingConfig(validate)
}

func (c *Config) validateBatchEnsemblingConfig(validate *validator.Validate) error {
	if !c.BatchEnsemblingConfig.Enabled {
		return nil
	}

	return validate.StructFiltered(c.BatchEnsemblingConfig, func(ns []byte) bool {
		// Determine the fields for validation
		validateFields := []string{
			"BatchEnsemblingConfig.JobConfig",
			"BatchEnsemblingConfig.RunnerConfig",
			"BatchEnsemblingConfig.ImageBuildingConfig",
		}
		// If the field's fully qualified name starts with the name of any of the chosen fields,
		// do not filter it from validation (return false will pick up the field for validation).
		for _, field := range validateFields {
			if strings.HasPrefix(string(ns), field) {
				return false
			}
		}
		return true
	})
}
```

This works the same as 
```go
type BatchEnsemblingConfig struct {
	Enabled             bool
	JobConfig           *JobConfig           `validate:"required_if=Enabled True"`
	RunnerConfig        *RunnerConfig        `validate:"required_if=Enabled True"`
	ImageBuildingConfig *ImageBuildingConfig `validate:"required_if=Enabled True"`
}
```

Going further beyond means removing all struct tags and writing our own custom validation, which I'm not sure if it's worth pursuing.

But a slight improvement with this correct required_if tag is that it guarantees that JobConfig, RunnerConfig and ImageBuildingConfig to not be nil when Enabled is true, so actually the nil check is not required anymore. Nonetheless I'll leave the nil check in place.

api/turing/config/config.go
ashwinath: Oh ya i changed this already.

api/turing/config/config.go
krithika369: Curious: In [this comment](https://github.com/gojek/turing/pull/119#discussion_r748048389):

>I tried using the method the base_manager used, it doesn't seem to work as intended as well. It works the same way as validate:"required_if=Enabled true"

Did you have your BatchEnsemblingConfig set to:
```
type BatchEnsemblingConfig struct {
	Enabled             bool
	JobConfig           *JobConfig           `validate:"-"`
	RunnerConfig        *RunnerConfig        `validate:"-"`
	ImageBuildingConfig *ImageBuildingConfig `validate:"-"`
...
}
```

The "-" asks the validator to skip validation on the nested struct which seems to happen by default.

ashwinath: I just tried this, the behaviour is also unintended. It will not try to validate if enabled is set to True.

ashwinath: I think the current state has the best behaviour for now (being the safest). The only unintended behaviour as of this state is enabled = false + other configs (JobConfig, RunnerConfig, ImageBuildingConfig) partially set. Which I think is acceptable because it won't be used if Enabled is false anyway but it forces the user to either fix it or remove them.

What is important is that Enabled = true and all 3 configs (JobConfig, RunnerConfig, ImageBuildingConfig) are set.

krithika369: I see. Okay.. Probably not very useful spending more time on this to tweak it. The current behavior sounds reasonable.

krithika369 (approved):
LGTM

api/turing/config/config.go
ashwinath: Yah I wrote in the test cases, so its pretty fast to run the configuration you asked. 😄

PR #121, opened 2021-11-14
Publish Turing's chart with optional MLP dependencies

romanwozniak:
### Context:
Turing depends on core MLP API (`projects` and `secrets` API) and Merlin API (`environments` API). Each of these components, as well as Turing, has its own Helm chart, however, these charts are missing some of the default values which makes the deployment much harder as it requires from users/administrators deep knowledge about the configuration of each of the components. With this PR we are moving forward and making it possible to install Turing with a single `helm install` command.

#### MLP/Merlin added as Turing's subcharts:
This could be changed in the future, but for now, we added MLP/Merlin charts as sub-charts of the Turing chart. The reason for this is that:
a) There are no published charts of Merlin at the moment 
b) It was required to update both MLP/Merlin charts to resolve conflicts over cluster resources (ConfigMaps/Deployments/Secrets etc.) created with the same name. 
It's expected, that at some point these changes will be propagated into the corresponding MLP/Merlin repositories and charts will be published into the charts registry, so we can remove these sub-charts and add the dependencies directly.

#### End-to-End test suite:
Our e2e CI job is now updated to 
1. Use the new chart to deploy Turing with all its dependencies into the k3d-k8s cluster
2. Run e2e separately for in-cluster credentials deployment (Turing API uses credentials mounted to its pods to deploy routers into the same k8s cluster where the API is deployed) and remote cluster deployments (Turing API fetches remote cluster credentials from Vaults KV secrets engine).

#### Other Changes:
- Added support of the `LogLevel` config into Turing API, to make it possible to enable `DEBUG` logs during the deployment
- Added Gateway and VirtualService to deploy mlp/merlin/turing under a single host:
  - `/api/turing` -> Turing API
  - `/api/merlin` -> Merlin API
  - `/api` -> MLP API
  - `/turing` -> Turing UI
  - `/` -> MLP UI

### Next Steps:
This will be addressed in a separate PR:
- [ ] Update documents section with detailed information on configurable chart values
- [ ] Update Turing's installation guide with updated information how to deploy Turing into k8s cluster


infra/charts/turing/subcharts/merlin/values.yaml
```diff
@@ -0,0 +1,173 @@
...
+    dashboardRepository: data-science/slo-specs
+    dashboardBranch: master
+    alertRepository: lens/artillery/datascience
+    alertBranch: master
+  warden:
+    apiHost: ""
+
+monitoring:
+  enabled: false
+  # baseURL: ""
```
terryyylim, line 79: I think it's okay to uncomment these 2 lines.

infra/charts/turing/subcharts/mlp/db-migrations/2_applications_config.up.sql
```diff
@@ -0,0 +1,29 @@
...
+                {
+                    "name": "Ensemblers",
+                    "href": "/ensemblers"
+                },
+                {
+                    "name": "Ensembling Jobs",
+                    "href": "/jobs"
+                }
+            ]
+        }',
```
terryyylim, line 26: Should we include the following in this db-migration or in a separate one as introduced in this [PR](https://github.com/gojek/mlp/pull/39)?
```
{
       "name": "Experiments",
       "href": "/experiments"
}
```

api/turing/service/mlp_service.go
```diff
@@ -3,17 +3,16 @@ package service
...
 	"context"
 	"fmt"
-	"log"
 	"net/http"
 	"time"
 
 	"github.com/antihax/optional"
 	merlin "github.com/gojek/merlin/client"
 	mlp "github.com/gojek/mlp/api/client"
+	"github.com/gojek/turing/api/turing/log"
```
terryyylim, line 12: nit: `turing` related imports in a separate block?

api/turing/it/database/database.go
```diff
@@ -1,17 +1,19 @@
...
 
 package database
 
 import (
 	"database/sql"
 	"fmt"
+	"testing"
+	"time"
+
+	"github.com/gojek/turing/api/turing/log"
```
terryyylim, line 12: nit: `turing` related imports in a separate block?

infra/charts/turing/templates/deployment.yaml
```diff
@@ -110,7 +110,7 @@ spec:
           secretName: {{ template "turing.fullname" .}}-api-config
       - name: db-migrations
         configMap:
-          name: {{ .Release.Name }}-db-migrations
+          name: {{ template "turing.fullname" .}}-db-migrations
```
terryyylim, line 113: nit `. }}` (space after fullstop)

.github/workflows/turing.yaml
```diff
@@ -342,15 +342,22 @@ jobs:
...
       TURING_API_VERSION: ${{ needs.build-api.outputs.api-version }}
       TURING_ROUTER_VERSION: ${{ needs.build-router.outputs.router-version }}
     needs:
       - build-api
       - build-router
       - build-cluster-init
+    strategy:
+      fail-fast: false
+      matrix:
+        useInClusterConfig: [false, true]
```
ashwinath, line 354: Screenshot 2021-11-15 at 9 57 47 AM

I'm not sure if it's possible but can we have nicer names other than true, false here?

api/e2e/test/00_all_e2e_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
```
ashwinath, line 1: What's the reason for this build tag change?

.github/workflows/turing.yaml
romanwozniak: This is how it works by design, but I have some workaround in mind, let me try it.

.github/workflows/turing.yaml
```diff
@@ -491,44 +479,45 @@ jobs:
...
+        with:
+          workloads: >-
+            deployment/mockserver,
+            deployment/turing-mlp,
+            deployment/turing-merlin,
+            deployment/turing
+          timeout: 300 #seconds
+          max-restarts: -1
 
       - if: steps.wait-for-deployment.outcome == 'failure' && always()
```
krithika369, line 518: What's this `always()` here ?

api/e2e/test/00_all_e2e_test.go
romanwozniak: Ah, I'm using go 1.17 on my local machine and go 1.17 has introduced transition from `+build ...` tags to `go:build ...` format. ([Release Notes](https://golang.org/doc/go1.17)). 
Both tags are currently accepted so I'll just keep it here as we will eventually have to upgrade anyway.

infra/charts/turing/.helmignore
```diff
@@ -0,0 +1 @@
+../.helmignore
```
krithika369, line 1: What does this do? Does this ignore the patterns defined inside `../.helmignore` ?

infra/charts/turing/subcharts/mlp/db-migrations/2_applications_config.up.sql
romanwozniak: I removed it intentionally as Turing Experiments aren't open-sourced yet. 
But moving forward, I think that populating this data (applications and applications config) by DB migrations is not a great idea (it was introduced by me so I'm fully responsible for it :D) and it should be replaced with a config file passed to MLP API at the deployment time.

infra/charts/turing/requirements.yaml
```diff
@@ -1,4 +1,14 @@
...
+  - name: mlp
+    version: 1.0.0
+    repository: file://subcharts/mlp
+    tags:
+      - mlp
+  - name: merlin
+    version: 0.7.0
+    repository: file://subcharts/merlin
+    tags:
+      - mlp
```
krithika369, line 14: Why is `vault` not a direct, conditional dependency for Turing? I thought it should be the same as MLP/Merlin.

infra/charts/turing-init/values.yaml
```diff
@@ -7,7 +7,7 @@ image:
   # -- Docker registry for Turing cluster init
   registry: ghcr.io/
   # -- Docker image repository for Turing cluster init
-  repository: gojek/turing-cluster-init
+  repository: gojek/turing/cluster-init
```
ashwinath, line 10: oops! thanks

.github/workflows/turing.yaml
romanwozniak: Better now? 
Screenshot 2021-11-15 at 10 17 03 AM

ashwinath:
Were the Merlin and MLP subcharts written from scratch? Is there a reason why we can't use the published charts from https://turing-ml.github.io/charts/ ?

romanwozniak:
> Were the Merlin and MLP subcharts written from scratch? Is there a reason why we can't use the published charts from https://turing-ml.github.io/charts/ ?

This is answered in PR's description.

infra/charts/turing/subcharts/merlin/values.yaml
```diff
@@ -0,0 +1,173 @@
...
+  enabled: false
+  dsn: ""
+
+newrelic:
+  enabled: false
+  appname: "merlin-api-dev"
+  licenseSecretName: "newrelic-license-secret"
+
+authorization:
+  enabled: true
```
krithika369, line 124: Nit: The default value is `true` here but `false` for MLP/Turing ? Should we be consistent?

ashwinath:
> > Were the Merlin and MLP subcharts written from scratch? Is there a reason why we can't use the published charts from https://turing-ml.github.io/charts/ ?
> 
> This is answered in PR's description.

> a) There are no published charts of Merlin at the moment

We are basing this on tags/0.9.0 or chart version 0.7.0 right? There is one manually published.

> b) It was required to update both MLP/Merlin charts to resolve conflicts over cluster resources (ConfigMaps/Deployments/Secrets etc.) created with the same name.

Anyway, I was just wondering if you could have used the fullnameOverride to bypass the configmap name collision: https://github.com/gojek/merlin/blob/v0.9.0/charts/merlin/templates/_helpers.tpl#L10

But since the effort has been made already and time is a huge constraint, don't think this should be done now.

romanwozniak:
> We are basing this on tags/0.9.0 or chart version 0.7.0 right? There is one manually published.

Correct, it's based on merlin 0.9.1 (chart 0.7.0) and mlp 1.14.6 (chart 1.0.0), but with a fair amount of modifications around the config structure.

> Anyway, I was just wondering if you could have used the fullnameOverride to bypass the configmap name collision: https://github.com/gojek/merlin/blob/v0.9.0/charts/merlin/templates/_helpers.tpl#L10

In order to have it fixing the problem, you have to use `template "merlin.fullname" .` everywhere, which is not the case in case of existing Merlin chart (for example: https://github.com/gojek/merlin/blob/fddb5edc91c02b8948f039c3eebf2a9ea73243b8/charts/merlin/templates/merlin-encryption-secret.yaml#L4). 
This PR adds all the necessary changes to actually make it possible to use `nameOverride` when it's required. (Example: https://github.com/romanwozniak/turing/blob/efbf6ec0f966c52e9ab8688033a739e8ed316a10/infra/charts/turing/values.yaml#L182).

krithika369 (approved):
Added some questions for clarification. LGTM.

api/turing/service/mlp_service.go
romanwozniak: It's usually grouped by  standard library imports vs external dependencies imports, not by turing and non-turing deps

infra/charts/turing/subcharts/merlin/templates/deployment.yaml
```diff
@@ -0,0 +1,216 @@
...
+            - |
+              curl --fail -v \
+                -H "X-Vault-Token: root" \
+                -d '{"master_ip": "kubernetes:443", "certs":"", "client_certificate":"","client_key":""}' \
+                "{{ template "merlin.vault.address" $ }}/v1/secret/{{ $cluster.cluster }}"
+        {{- end }}
+      {{- end }}
+      containers:
+      - name: merlin
+        image: "{{ .Values.image.registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
ashwinath, line 50: I wonder if this pattern should be used, we are guilty of being inconsistent here. See [this](https://github.com/gojek/turing/blob/main/infra/charts/turing/templates/deployment.yaml#L60). Sometimes we need the forward slash and sometimes we dont need it. I wonder if one day we should standardise this and do what other helm repositories do; [just repository and tag](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L40). I don't think it should be addressed in this PR but maybe a comment indicating the forward slash requirement/non requirement.

.github/workflows/turing.yaml
romanwozniak: `always()` allows running this step even if the previous step has failed. There is an implicit `success()` condition on each step (so step is not being executed if any of the previous steps has failed) hence it should be explicitly overridden here. I think it can also be replaced with `failure()`, but it doesn't change anything as we are also looking for another specific condition to happen (failure of the `wait-for-deployment` step).

infra/charts/turing/.helmignore
romanwozniak: Yup, works similarly to `.gitignore` and `.dockerignore` – it excludes provided paths from being packaged with the chart

ashwinath:
> > We are basing this on tags/0.9.0 or chart version 0.7.0 right? There is one manually published.
> 
> Correct, it's based on merlin 0.9.1 (chart 0.7.0) and mlp 1.14.6 (chart 1.0.0), but with a fair amount of modifications around the config structure.
> 
> > Anyway, I was just wondering if you could have used the fullnameOverride to bypass the configmap name collision: https://github.com/gojek/merlin/blob/v0.9.0/charts/merlin/templates/_helpers.tpl#L10
> 
> In order to have it fix the problem, you have to use `template "merlin.fullname" .` everywhere, which is not the case for the existing Merlin chart (for example https://github.com/gojek/merlin/blob/fddb5edc91c02b8948f039c3eebf2a9ea73243b8/charts/merlin/templates/merlin-encryption-secret.yaml#L4). This PR adds all the necessary changes to actually make it possible to use `nameOverride` when it's required. (Example: https://github.com/romanwozniak/turing/blob/efbf6ec0f966c52e9ab8688033a739e8ed316a10/infra/charts/turing/values.yaml#L182).

Ah I did not realise that it isn't using the fullnameOverride everywhere, we should be conscious about this. The other thing is also we have to consider truncating all names to 63 characters next time as well.

infra/charts/turing/requirements.yaml
romanwozniak: But `vault` is also Merlin's dependency (it's even more Merlin's dependency compared to Turing, because Turing can run without Vault (with the in-cluster credentials provided), but Merlin - can't). So for now Vault is Merlin's dep, but we can consider making it a root-level dependency for MLP once we have that chart updated

infra/charts/turing/subcharts/merlin/values.yaml
romanwozniak: Yes, this makes sense. Thanks for catching

infra/charts/turing/templates/cluster-role-binding.yaml
```diff
@@ -2,13 +2,13 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "turing.service-account" . }}
```
ashwinath, line 5: I'm curious, did you encounter the problem with trying to get the hyphenated version of variable names to work or did you do this out of consistency? I ran into this problem in the past (https://github.com/helm/helm/issues/2192)

infra/charts/turing/subcharts/merlin/templates/deployment.yaml
romanwozniak: This _IS_ an attempt to standardize it. Among three (mlp/merlin/turing), only Turing expected to have a trailing slash as part of the registry name. So the link you included (See [this](https://github.com/gojek/turing/blob/main/infra/charts/turing/templates/deployment.yaml#L60)) was actually updated in this PR to address the inconsistency.

infra/charts/turing/templates/virtualservice-mlp.yaml
```diff
@@ -0,0 +1,88 @@
...
+      rewrite:
+        uri: "/"
+      route:
+        - destination:
+            host: {{ template "turing.fullname" . }}
+            port:
+              number: 8080
+    - match:
+        - uri:
+            prefix: "/api/merlin/"
```
ashwinath, line 42: Is this necessary? `/api/merlin` is a subset of `/api/merlin/`

infra/charts/turing/subcharts/merlin/templates/deployment.yaml
ashwinath: Ok, I'm still in the middle of reviewing this PR so I have not reached this part yet. But ok glad this was addressed.

.github/workflows/turing.yaml
ashwinath: Yah, saw this change. Nice.

infra/charts/turing/values.yaml
```diff
@@ -108,6 +135,9 @@ dbMigrations:
     # -- Docker tag for golang-migrate Docker image https://hub.docker.com/r/migrate/migrate
     tag: v4.7.1
 
+sentry:
+  dsn: ""
```
ashwinath, line 139: Maybe a comment to say what this is and regenerate the helm-docs?

romanwozniak:
test

ashwinath (approved):
Mostly pretty good, I can't imagine this being an easy task.

infra/charts/turing-init/values.yaml
ashwinath: Actually if we were to do this, can we delete the old images, so that it's not confusing to the user on which image to use.

infra/charts/turing/values.yaml
romanwozniak: As it's said in PR description, docs will be addressed in a separate PR

infra/charts/turing/templates/virtualservice-mlp.yaml
romanwozniak: There is an issue with istio https://github.com/istio/istio/issues/8076#issuecomment-427057691 on this.

infra/charts/turing/.helmignore
krithika369: Oh my question was actually about why the content of this file is `../.helmignore`. But I realise now that it's actually symlinked to `../.helmignore`. Makes sense.

infra/charts/turing-init/values.yaml
romanwozniak: There are no old images, the CI pipeline publishes it under `gojek/turing/cluster-init`, it was just that helm chart contains wrong values for it.

infra/charts/turing/templates/virtualservice-mlp.yaml
ashwinath: This is hilarous, could you add a comment inside with this link to show why it was done this way?

PR #122, opened 2021-11-15
Add module resolution for react-beautiful-dnd

terryyylim:
Since our internal Experiment-Engine (EE) UI requires the use of Drag-And-Drop component from `eui` library which has dependency on `react-beautiful-dnd`, an issue as described in this [PR](https://github.com/gojekfarm/react-beautiful-dnd/pull/1) was found and needs to be resolved, by using the fix introduced (via module resolution).

Module resolution is required on Turing because it uses the same `eui` version as the EE (`32.3.0`). In the scenario where the EE uses a different minor/patch version, the change introduced in this PR will not be required since the dependency is not a singleton. Validation has been done and this behaviour is consistent with our conclusion in previous PR: #114 .

Regardless, this module resolution fix will be useful if Turing-UI starts using Drag-And-Drop component for 2 different lists.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -39,7 +39,8 @@
     "react-scripts": "5.0.0-next.37"
   },
   "resolutions": {
-    "react-scripts/**/postcss-normalize": "10.0.1"
+    "react-scripts/**/postcss-normalize": "10.0.1",
+    "**/react-beautiful-dnd": "git+https://github.com/gojekfarm/react-beautiful-dnd.git#master"
   },
   "lint-staged": {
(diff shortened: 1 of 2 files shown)


krithika369 (approved):
LGTM. For ease of understanding the reason for this change in the future, could you elaborate more on the mechanics of module federation?

It is my understanding that if we change the experiment engine's Eui dependency to a different version than Turing (say, a different minor version), this change in Turing is not required (since this dependency is not a singleton). Could you also test this and update the PR description? In doing so, we can also validate that the behavior is still consistent with whatever we concluded in the previous PR: https://github.com/gojek/turing/pull/114.

PR #124, opened 2021-11-16
Add templating of turing ingress for networking api v1beta1 and v1.

ashwinath:
Some clusters may be on a lower version of Kubernetes (< 1.19) and do not allow for the Ingress object under the `networking.k8s.io/v1` API but allow for `networking.k8s.io/v1beta1` API. This PR allows for either configuration.


infra/charts/turing/templates/ingress.yaml
```diff
@@ -1,11 +1,10 @@
 {{- if .Values.turing.ingress.enabled }}
-apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: {{ template "turing.fullname" . }}
   namespace: {{ .Release.Namespace }}
   annotations:
-    kubernetes.io/ingress.class: {{ .Values.turing.ingress.class }}
+    kubernetes.io/ingress.class: "{{ .Values.turing.ingress.class }}"
```
ashwinath, line 7: IIRC empty string is valid as well but not nil object.

ashwinath:
I tested the template generation of the ingress with `turing.ingress.useV1Beta1 = true` in a Kubernetes 1.18 environment.

infra/charts/turing/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.2.0
+version: 0.2.1
```
romanwozniak, line 4: This should be `0.2.2` now

romanwozniak (approved):
The same thing should probably be done for mlp/merlin ingresses as well, to keep it consistent. Besides this, LGTM

infra/charts/turing/Chart.yaml
ashwinath: ok I'll bump it up.

PR #125, opened 2021-11-16
Fix building cluster-init container on tags.

ashwinath:
As of now, the pattern for building on release tags are not there, so it will not publish a tagged version. We also set the appVersion to 1.0.0 like turing charts and also set the default image to 1.0.0.

We should also retag version 1.0.0 since we missed this out. I'm also unsure if the `make version` works on tags, could you help to check if this will work?


ashwinath:
Tagging as latest and pushing it manually for now

PR #126, opened 2021-11-17
Batch predict endpoint

leonlnj:
### Overview
This PR will add a new endpoint, `batch_predict` to Turing router. 
Currently the individual request are handled async. Individual request will behave the same as current state (no retries). 
Individual request failures will not fail the entire request and be lodged as part of the response.

Current batch endpoint expects "batch_request" with an array of requests and will return a response with "batch_result" with the array of response with fields `code`, `error`, `data`. 

Sample request
image

Sample response
image
image

### Modification
`api/e2e/test/01_create_router_test.go` - Add batch_endpoint testing to e2e test
`api/turing/cluster/servicebuilder/router.go` - Change string to []string, to add batch_endpoint prefix to default
`api/turing/cluster/virtual_service.go` - Loop through prefixes to add multiple matches if provided

`engines/router/missionctl/handlers/http_handler.go` - Refactored `ServeHTTP` to extract common method for batch_predict. Primarily `getPrediction()`, `enableTracingSpan()`, `measureRequestDuration()`.

`engines/router/missionctl/handlers/batch_http_handler.go` - New handler for the batch endpoint. Embedding `httpHandler`   and providing implementation for `ServeHTTP` using common method refactored in `http_handler.go`. 

`engines/router/missionctl/turingctx/context.go` - Added new method to return original context as parent and create a new turingreqid with input suffix

### DIFF
--- a/api/turing/cluster/servicebuilder/router.go
+++ b/api/turing/cluster/servicebuilder/router.go
@@ -72,7 +72,7 @@ const (
 const (
 	defaultIstioGateway   = "istio-ingressgateway.istio-system.svc.cluster.local"
 	defaultGateway        = "knative-ingress-gateway.knative-serving"
-	defaultMatchURIPrefix = "/v1/predict"
+	defaultMatchURIPrefix = "/v1/predict,/v1/batch_predict"
 )
 
 // NewRouterService creates a new cluster Service object with the required config
@@ -158,7 +158,7 @@ func (sb *clusterSvcBuilder) NewRouterEndpoint(
 		Endpoint:        host,
 		DestinationHost: defaultIstioGateway,
 		HostRewrite:     veURL.Hostname(),
-		MatchURIPrefix:  defaultMatchURIPrefix,
+		MatchURIPrefix:  strings.Split(defaultMatchURIPrefix, ","),
 	}, nil
 }
 
--- a/api/turing/cluster/servicebuilder/router.go
+++ b/api/turing/cluster/servicebuilder/router.go
@@ -70,10 +70,11 @@ const (
 
 // Router endpoint constants
 const (
-	defaultIstioGateway   = "istio-ingressgateway.istio-system.svc.cluster.local"
-	defaultGateway        = "knative-ingress-gateway.knative-serving"
-	defaultMatchURIPrefix = "/v1/predict,/v1/batch_predict"
+	defaultIstioGateway = "istio-ingressgateway.istio-system.svc.cluster.local"
+	defaultGateway      = "knative-ingress-gateway.knative-serving"
 )
+
+var defaultMatchURIPrefixes = []string{"/v1/predict", "/v1/batch_predict"}
 
 // NewRouterService creates a new cluster Service object with the required config
 // for the Turing router to be deployed.
@@ -151,14 +152,14 @@ func (sb *clusterSvcBuilder) NewRouterEndpoint(
 	host := strings.Replace(veURL.Hostname(), routerName, routerEndpointName, 1)
 
 	return &cluster.VirtualService{
-		Name:            routerEndpointName,
-		Namespace:       project.Name,
-		Labels:          labels,
-		Gateway:         defaultGateway,
-		Endpoint:        host,
-		DestinationHost: defaultIstioGateway,
-		HostRewrite:     veURL.Hostname(),
-		MatchURIPrefix:  strings.Split(defaultMatchURIPrefix, ","),
+		Name:             routerEndpointName,
+		Namespace:        project.Name,
+		Labels:           labels,
+		Gateway:          defaultGateway,
+		Endpoint:         host,
+		DestinationHost:  defaultIstioGateway,
+		HostRewrite:      veURL.Hostname(),
+		MatchURIPrefixes: defaultMatchURIPrefixes,
 	}, nil
 }
 
--- a/api/turing/cluster/virtual_service.go
+++ b/api/turing/cluster/virtual_service.go
@@ -7,14 +7,14 @@ import (
 )
 
 type VirtualService struct {
-	Name            string            `json:"name"`
-	Namespace       string            `json:"namespace"`
-	Labels          map[string]string `json:"labels"`
-	Gateway         string            `json:"gateway"`
-	Endpoint        string            `json:"endpoint"`
-	DestinationHost string            `json:"destination_host"`
-	HostRewrite     string            `json:"host_rewrite"`
-	MatchURIPrefix  []string          `json:"match_uri_prefix"`
+	Name             string            `json:"name"`
+	Namespace        string            `json:"namespace"`
+	Labels           map[string]string `json:"labels"`
+	Gateway          string            `json:"gateway"`
+	Endpoint         string            `json:"endpoint"`
+	DestinationHost  string            `json:"destination_host"`
+	HostRewrite      string            `json:"host_rewrite"`
+	MatchURIPrefixes []string          `json:"match_uri_prefix"`
 }
 
 func (cfg VirtualService) BuildVirtualService() *v1alpha3.VirtualService {
@@ -29,8 +29,8 @@ func (cfg VirtualService) BuildVirtualService() *v1alpha3.VirtualService {
 		},
 		Weight: 100,
 	}
-	httpMatches := make([]*networking.HTTPMatchRequest, len(cfg.MatchURIPrefix))
-	for index, prefix := range cfg.MatchURIPrefix {
+	httpMatches := make([]*networking.HTTPMatchRequest, len(cfg.MatchURIPrefixes))
+	for index, prefix := range cfg.MatchURIPrefixes {
 		uri := networking.HTTPMatchRequest{
 			Uri: &networking.StringMatch{
 				MatchType: &networking.StringMatch_Prefix{
--- a/api/turing/cluster/virtual_service.go
+++ b/api/turing/cluster/virtual_service.go
@@ -14,7 +14,7 @@ type VirtualService struct {
 	Endpoint        string            `json:"endpoint"`
 	DestinationHost string            `json:"destination_host"`
 	HostRewrite     string            `json:"host_rewrite"`
-	MatchURIPrefix  string            `json:"match_uri_prefix"`
+	MatchURIPrefix  []string          `json:"match_uri_prefix"`
 }
 
 func (cfg VirtualService) BuildVirtualService() *v1alpha3.VirtualService {
@@ -29,6 +29,17 @@ func (cfg VirtualService) BuildVirtualService() *v1alpha3.VirtualService {
 		},
 		Weight: 100,
 	}
+	httpMatches := make([]*networking.HTTPMatchRequest, len(cfg.MatchURIPrefix))
+	for index, prefix := range cfg.MatchURIPrefix {
+		uri := networking.HTTPMatchRequest{
+			Uri: &networking.StringMatch{
+				MatchType: &networking.StringMatch_Prefix{
+					Prefix: prefix,
+				},
+			},
+		}
+		httpMatches[index] = &uri
+	}
 
 	return &v1alpha3.VirtualService{
 		ObjectMeta: metav1.ObjectMeta{
@@ -41,15 +52,7 @@ func (cfg VirtualService) BuildVirtualService() *v1alpha3.VirtualService {
 			Gateways: []string{cfg.Gateway},
 			Http: []*networking.HTTPRoute{
 				{
-					Match: []*networking.HTTPMatchRequest{
-						{
-							Uri: &networking.StringMatch{
-								MatchType: &networking.StringMatch_Prefix{
-									Prefix: cfg.MatchURIPrefix,
-								},
-							},
-						},
-					},
+					Match: httpMatches,
 					Route: []*networking.HTTPRouteDestination{
 						httpRouteDest,
 					},
--- a/engines/router/missionctl/handlers/batch_http_handler.go
+++ b/engines/router/missionctl/handlers/batch_http_handler.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"sync"
 
 	"github.com/gojek/turing/engines/router/missionctl"
@@ -76,8 +77,9 @@ func (h *batchHTTPHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) 
 
 		go func(index int, jsonRequestBody json.RawMessage) {
 			defer waitGroup.Done()
+			requestContext := turingctx.NewTuringContextWithSuffix(ctx, strconv.Itoa(index))
 			var batchResponse batchResponse
-			resp, httpErr := h.getPrediction(ctx, req, ctxLogger, jsonRequestBody)
+			resp, httpErr := h.getPrediction(requestContext, req, ctxLogger, jsonRequestBody)
 			if httpErr != nil {
 				batchResponse.StatusCode = httpErr.Code
 				batchResponse.ErrorMsg = httpErr.Message
--- a/engines/router/missionctl/handlers/batch_http_handler.go
+++ b/engines/router/missionctl/handlers/batch_http_handler.go
@@ -1,26 +1,37 @@
 package handlers
 
 import (
-	"github.com/gojek/turing/engines/router/missionctl"
+	"encoding/json"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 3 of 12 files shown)


romanwozniak:
Thanks for this MR @leonlnj, before I have some more feedback on it, would it be possible to extend our e2e test suite to also cover batch predict flow?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	// Create context from the request context
+	ctx := turingctx.NewTuringContext(req.Context())
+	// Create context logger
+	ctxLogger := log.WithContext(ctx)
+	defer func() {
+		_ = ctxLogger.Sync()
+	}()
+
+	// Get the unique turing request id from the context
+	turingReqID, err := turingctx.GetRequestID(ctx)
```
romanwozniak, line 47: Are there any side effects if we have the same turing request ID for multiple predictions? Would it mess with result logging? @krithika369 can you think of any?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Read the request body
+	requestBody, err := ioutil.ReadAll(req.Body)
+	if err != nil {
+		h.error(ctx, rw, errors.NewHTTPError(err))
+		return
+	}
+
+	//Valid request
+	if valid := json.Valid(requestBody); !valid {
```
romanwozniak, line 64: Is it necessary to explicitly check if the payload is a valid JSON or would it be sufficient to just capture a possible error from `json.Unmarshall` method below?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+	var batchRequests map[string][]interface{}
+	err = json.Unmarshal(requestBody, &batchRequests)
+	if err != nil {
+		h.error(ctx, rw, errors.NewHTTPError(errors.Newf(errors.BadInput,
+			`Invalid json request`)))
+		return
+	}
+
+	//Validate request
+	if _, ok := batchRequests["batch_request"]; !ok {
```
romanwozniak, line 80: 1. Is it necessary to have individual requests to be nested under `batch_request`? Can the batch predict payload be just `[{"key": "request_1"}, {"key": "request_2"}, ...]`?
2. If there is a reason to keep the current structure (i.e. `batch_request` property), then can we define it as Go type so we don't use a hard-coded `"batch_request"` string for accessing the data?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	//Handle request asynchronously
+	var batchResponses = make([]batchResponse, len(batchRequests["batch_request"]))
+	var waitGroup sync.WaitGroup
+	for index, value := range batchRequests["batch_request"] {
+		waitGroup.Add(1)
+
+		go func(index int, jsonRequestBody interface{}) {
+			defer waitGroup.Done()
+			batchRequestBody, _ := json.Marshal(jsonRequestBody)
```
romanwozniak, line 94: That's very inefficient to unmarshal data first to marshal it again later. Think about the possible batch size in a magnitude of a thousand(s) predict requests. 

Maybe you can use `[]json.RawMessage` (instead of `[]interface{}`) to split the batch into individual requests?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+			var batchResponse batchResponse
+			resp, httpErr := h.getPrediction(ctx, req, ctxLogger, batchRequestBody)
+			if httpErr != nil {
+				batchResponse.StatusCode = httpErr.Code
+				batchResponse.ErrorMsg = httpErr.Message
+				batchResponses[index] = batchResponse
+				return
+			}
+			err := json.Unmarshal(resp.Body(), &batchResponse.Data)
```
romanwozniak, line 103: Along the same lines, maybe `batchResponse.Data` can be set to either `json.RawMessage` or even `[]byte`

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+			if err != nil {
+				batchResponse.StatusCode = http.StatusInternalServerError
+				batchResponse.ErrorMsg = "Unable to marshall response into json"
+			}
+			batchResponse.StatusCode = http.StatusOK
+			batchResponses[index] = batchResponse
+		}(index, value)
+	}
+	waitGroup.Wait()
```
romanwozniak, line 112: How do we handle a timeout? Do we take into account router's timeout setting for the batch predict endpoint?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Write the json response to the writer
+	rw.Header().Set("Content-Type", "application/json")
```
romanwozniak, line 115: Maybe we can just mirror back the original request's Content-Type:
```go
rw.Header().Set("Content-Type", resp.Header().Get("Content-Type"))
```

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Get the unique turing request id from the context
+	turingReqID, err := turingctx.GetRequestID(ctx)
+	if err != nil {
+		ctxLogger.Errorf("Could not retrieve Turing Request ID from context: %v",
+			err.Error())
+	}
+	ctxLogger.Debugf("Received batch request for %v", turingReqID)
+
+	ctx = h.enableTracingSpan(ctx, req)
```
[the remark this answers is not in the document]
leonlnj: yes, the check is within the function, i should move it out. thanks!

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: 1. Yes it can be, I have tried that previously. What I had in mind was to keep it consistent with XP's response where it was {"data": .....} in general. Then i added the a key to the request to keep similar on both req/response. I just look through the Turing swagger doc and it make more sense to use `[{"key": "request_1"}, {"key": "request_2"}, ...]` instead for consistency, missed this out. 
2. Will change both to the suggested, feels more consistent that way.

Thanks!

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: they are handled individually in `resp, httpErr := h.getPrediction(ctx, req, ctxLogger, batchRequestBody)`

If the router timeout is 1s , individual request will get a 408 fiber timeout at 1s.
In the event of route timeout, individual request will get a 500 timeout (ss in the last section of description)

With the router timeout applied on individual requests, it will appear as though it is applied to the main request. 
I left the behaviour as such because in the event of partial success/failures, it might be more ideal to return them as it is , rather than to timeout the entire batch request. 

The only issues is that Im not sure what is the http status code to return, in the event of partial failure, keeping it as 200 for now.

leonlnj:
Thanks, will look into the e2e before cleaning it up!

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: My intention was to differential invalid json and missing `batch_request`, since it could be a valid json and fail in unmarshalling due to missing key. 

I'll removed the `batch_request` and this can be combined.

engines/router/missionctl/handlers/batch_http_handler.go
krithika369: The example you have shown is that of the response: `resp.Header()`

> Maybe we can just mirror back the original request's Content-Type

But if you meant this:
* I am not sure if there we be cases where the request's content type header is missing.
* If there will be cases where the request content-type is json but the response type is text

So I feel it's better to copy from the response (unless Turing is crafting it, in which case it should determine the type).

ashwinath: Not in favour of mirroring it, since the router does not dynamically handle all kinds of content-type.

engines/router/missionctl/handlers/batch_http_handler.go
krithika369: I can't think of any. There is no uniqueness constraint on the Turing Request ID. However, I'm curious - is that the only distinguisher in the logs - i.e., the presence of multiple records with the same ID indicate it's a batch request? 

Should we instead append a suffix to the TuringReqID? So, for batch requests, they will be `{unique_hash}-{idx}` where the idx is their index in the original request. That way, we have a common pattern and a distinguisher for batch requests. wdyt?

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+type batchResponse struct {
+	StatusCode int         `json:"code"`
+	ErrorMsg   string      `json:"error,omitempty"`
+	Data       interface{} `json:"data,omitempty"`
+}
+
+type batchResult struct {
+	BatchResult []batchResponse `json:"batch_result"`
+}
```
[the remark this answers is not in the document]
ashwinath: Also actually is StatusCode really required? Can't it be taken from the HTTP header?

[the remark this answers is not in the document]
leonlnj: When i checked, it was not available from the route response, there was only application/content in the header. 

Also, these are for the individual resquest's, I cant think of any other way to depict them in case of partial failures.

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: I do feel that copying from response is neater, but my dilemma is that if there were multiple individual request, some with failure and some with success, it might be different content type. 

In most cases, I thought that returning a json with individual responses made the most sense to me, and hence I hardcoded to it to application/json initially.

leonlnj:
will reopen the PR once i patch up the virtualservice and e2e test, apologies for not doing it early!

engines/router/missionctl/handlers/batch_http_handler.go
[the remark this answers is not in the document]
ashwinath: Not sure what you mean it's here: https://github.com/gojek/turing/blob/5cdd2094c69ffeab327605d513eede288c5f4247/engines/router/missionctl/handlers/batch_http_handler.go#L97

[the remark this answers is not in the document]
leonlnj: yup that is for the header/status for the batch request, the status code is meant for individual responses from the route.

image

engines/router/missionctl/handlers/batch_http_handler.go
ashwinath: I think the other alternative is to reject non `content-type: application/json` in requests if you want to be strict about it, then it's safe to mirror.

engines/router/missionctl/handlers/batch_http_handler.go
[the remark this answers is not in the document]
ashwinath: no thats the payload not the http headers.

leonlnj:
updated the MR with the changes, thanks all once again

api/turing/cluster/servicebuilder/router.go
```diff
@@ -72,7 +72,7 @@ const (
 const (
 	defaultIstioGateway   = "istio-ingressgateway.istio-system.svc.cluster.local"
 	defaultGateway        = "knative-ingress-gateway.knative-serving"
-	defaultMatchURIPrefix = "/v1/predict"
+	defaultMatchURIPrefix = "/v1/predict,/v1/batch_predict"
```
romanwozniak, line 75: Would it be a better idea to define it as:
```go
var defaultMatchURIPrefixes = []string{"/v1/predict", "/v1/batch_predict"}
```
So you don't need to do the unnecessary string manipulations

api/turing/cluster/virtual_service.go
```diff
@@ -14,7 +14,7 @@ type VirtualService struct {
 	Endpoint        string            `json:"endpoint"`
 	DestinationHost string            `json:"destination_host"`
 	HostRewrite     string            `json:"host_rewrite"`
-	MatchURIPrefix  string            `json:"match_uri_prefix"`
+	MatchURIPrefix  []string          `json:"match_uri_prefix"`
```
romanwozniak, line 17: Maybe then also change the property field to `MatchURIPrefixes` to make it easy to read that multiple prefixes can be passed in

engines/router/missionctl/handlers/batch_http_handler.go
```diff
@@ -0,0 +1,106 @@
...
+	// Get the unique turing request id from the context
+	turingReqID, err := turingctx.GetRequestID(ctx)
+	if err != nil {
+		ctxLogger.Errorf("Could not retrieve Turing Request ID from context: %v",
+			err.Error())
+	}
+	ctxLogger.Debugf("Received batch request for %v", turingReqID)
+
+	if tracing.Glob().IsEnabled() {
+		ctx = h.enableTracingSpan(ctx, req)
```
romanwozniak, line 53: Can you make sure that distributed tracing works as intended with the batch request to Turing? I.e. that the tracing headers are propagated to processing individual requests and separate nested tracing spans are created for each individual request?

engines/router/missionctl/handlers/batch_http_handler.go
romanwozniak: Ok, this makes sense. Resolving this thread

engines/router/missionctl/handlers/batch_http_handler_test.go
```diff
@@ -0,0 +1,171 @@
...
+func createGenericMissionControl(t *testing.T) missionctl.MissionControl {
+	//Create missionctl with route for testing
+	missionCtl, err := missionctl.NewMissionControl(
+		nil,
+		&config.EnrichmentConfig{
+			Endpoint: "",
+			Timeout:  time.Second,
+		},
+		&config.RouterConfig{
+			//ConfigFile: filepath.Join("../testdata", "nop_default_router.yaml"),
```
romanwozniak, line 157: Leftover?

engines/router/missionctl/handlers/batch_http_handler_test.go
```diff
@@ -0,0 +1,171 @@
...
+			ConfigFile: filepath.Join("../testdata", "batch_router_test.yaml"),
```
romanwozniak, line 158: Maybe this ConfigFile can be passed as an argument to `createGenericMissionControl` function?

engines/router/missionctl/handlers/batch_http_handler_test.go
```diff
@@ -0,0 +1,171 @@
...
+			if test.expectedContentType == "application/json" {
+				assert.JSONEq(t, test.expectedResponseBody, string(result))
+			} else {
+				assert.Equal(t, test.expectedResponseBody, string(result))
+			}
+		})
+	}
+}
+
+func TestNewBatchHTTPHandlerBadRoute(t *testing.T) {
```
romanwozniak, line 91: Can you mergte `TestNewBatchHTTPHandlerBadRoute`  with `TestNewBatchHTTPHandler` by adding `missionCtl ` as a parameter of the test suite:
```go
tests := map[string]struct {
		payload              string
		expectedResponseBody string
		expectedStatusCode   int
		expectedContentType  string
                missionCtl: *missionctl.MissionControl
	}
```
and then dynamically constructing HTTP handler inside the test Run:
```go
for name, test := range tests {
	t.Run(name, func(t *testing.T) {
		batchHTTPHandler := NewBatchHTTPHandler(missionCtl)
		assert.NotNil(t, batchHTTPHandler)
		// ...
	})
}
```

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: added it, will look something like this
Screenshot 2021-11-25 at 9 20 13 AM

engines/router/missionctl/handlers/batch_http_handler.go
leonlnj: checked.
Screenshot 2021-11-24 at 7 41 09 PM
Screenshot 2021-11-24 at 7 13 21 PM

romanwozniak (approved):
Thanks for looking closely at all the PR comments, it LGTM now 🚀

PR #128, opened 2021-11-19
Remove db scripts out of helm

ashwinath:
Currently, the helm chart contains the database migration scripts. But we should not be duplicating these scripts into the helm charts and running a separate container just to initialise them. Embedding it in the application logic makes it easier to start the Turing API as a developer as well as removing the need for an init container.


api/turing/config/config.go
```diff
@@ -228,11 +228,12 @@ type SinglePageApplicationConfig struct {
...
-	Port     int    `validate:"required"`
-	User     string `validate:"required"`
-	Password string `validate:"required"`
-	Database string `validate:"required"`
+	Host                       string `validate:"required"`
+	Port                       int    `validate:"required"`
+	User                       string `validate:"required"`
+	Password                   string `validate:"required"`
+	Database                   string `validate:"required"`
+	MigrationScriptsFolderPath string `validate:"required"`
```
romanwozniak, line 236: super nit, but it can be just `MigrationsPath`or `MigrationsFolder`

api/turing/server/application.go
```diff
@@ -153,3 +148,49 @@ func Run() {
 		log.Errorf("Failed to start turing-api: %s", err)
 	}
 }
+
+func initDB(cfg *config.Config) (*gorm.DB, error) {
```
romanwozniak, line 152: It makes sense to move it into a separate package (`db`?) and re-use this part for the integration tests db initialization (https://github.com/gojek/turing/blob/main/api/turing/it/database/database.go)

romanwozniak: Also, you don't need the entire `*config.Config` here, you can simply pass an instance of `config.DatabaseConfig` instead

romanwozniak (commented):
Please refactor it to share the part of running the migrations between the main app and the integration test setup. Rather than that, I have no disagreement on how it's done

api/turing/config/config.go
ashwinath: Had a bit of a brain fart here. Yeah I'll do `MigrationsFolder` instead.

api/turing/server/application.go
ashwinath: Ok db package sounds better actually.

api/turing/it/database/database.go
```diff
@@ -9,15 +9,14 @@ import (
...
+	"github.com/gojek/turing/api/turing/config"
+	"github.com/gojek/turing/api/turing/db"
 	"github.com/gojek/turing/api/turing/log"
-	gomigrate "github.com/golang-migrate/migrate/v4"
-	"github.com/golang-migrate/migrate/v4/database/postgres"
-	_ "github.com/golang-migrate/migrate/v4/source/file"
 	"github.com/jinzhu/gorm"
 )
 
 func connectionString(db string) string {
```
romanwozniak, line 18: This can be moved to `db`package as well (as it's used there as well), with a modification:
```
func ConnectionString (cfg *config.DatabaseConfig) string {...}
```

api/turing/db/database.go
```diff
@@ -0,0 +1,68 @@
+package db
```
romanwozniak, line 1: I'd suggest to name the package `database` and the DB instance as `db`

api/turing/it/database/database.go
```diff
@@ -76,7 +59,15 @@ func CreateTestDatabase() (*gorm.DB, func(), error) {
 		}
 	}
 
-	if testDb, err = migrate(testDb, testDbName); err != nil {
+	dbCfg := &config.DatabaseConfig{
```
romanwozniak, line 62: you can actually initialize an instance of `config.DatabaseConfig` at `api/turing/it/database/config.go` as the user-provided input is being read there

ashwinath: I thought about it but the database always changes with a new test suite, so the object will have to be created upon invocation of `WithTestDatabase()`

romanwozniak: well, your global variables are guaranteed to be initialized before `WithTestDatabase` is invoked. I.e. your variables here: https://github.com/gojek/turing/blob/0e8eec8042bbb074240b9aa7d4f4fb2303a3ca9c/api/turing/it/database/config.go#L7-L13
are guaranteed to be initialized, so why can't you simply replace it with:
```go
var dbConfig = config.DatabaseConfig {
	Host:              getEnvOrDefault("DATABASE_HOST", "localhost"),
	Port:              strconv.Atoi(getEnvOrDefault("DATABASE_PORT", "5432")),
	User:              getEnvOrDefault("DATABASE_USER", "turing"),
	Password:          getEnvOrDefault("DATABASE_PASSWORD", "turing"),
	Database:          getEnvOrDefault("DATABASE_NAME", "turing"),
	MigrationsFolder:  "../db-migrations/",
}
```

ashwinath: Ya ok i did something like this but mutate as needed with a helper method.

ashwinath: The current state is quite confusing as 2 kinds of db are being juggled so i refactored it for readability as well.

api/turing/db/database.go
ashwinath: done

api/turing/it/database/database.go
ashwinath: done

romanwozniak (approved):
LGTM

PR #129, opened 2021-11-23
Fix Router details UI overflow

terryyylim:
**View before fix**
image

**View after fix**
image

This PR fixes:
- Overflow of elements when window size is small
- Abruptly truncated endpoint

### DIFF
--- a/ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.scss
+++ b/ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.scss
@@ -1,3 +1,8 @@
+.euiFlexGroup--justifyContentSpaceBetween {
+  justify-content: space-between;
+  flex-wrap: wrap;
+}
+
 .euiFlexItem--endpoint {
   max-width: calc(60% - 24px);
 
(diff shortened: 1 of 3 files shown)


ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.scss
```diff
@@ -1,3 +1,8 @@
+.euiFlexGroup--justifyContentSpaceBetween {
+  justify-content: space-between;
+  flex-wrap: wrap;
+}
```
romanwozniak, line 4: You should use `EuiFlexGroup` props for this, instead of custom styles: https://elastic.github.io/eui/#/layout/flex#flex-group-is-for-a-single-row-layout
```js
<EuiFlexGroup justifyContent="spaceBetween" wrap={true}>
...
</EuiFlexGroup>
```

romanwozniak:
The font size of the endpoint URL looks bigger in the "fixed" screenshot, inconsistent with other sections on the header

terryyylim:
> The font size of the endpoint URL looks bigger in the "fixed" screenshot, inconsistent with other sections on the header

Confirmed that it's just a discrepancy in screenshot, there's no change to actual UI.

romanwozniak (approved):
Thanks for the fix!

PR #131, opened 2021-11-23
Add debouncing to form validation

krithika369:
Recently, the Router Form has begun to display performance degradation, especially when attempting to type into the text fields when the form has already been touched (which in turn causes the validation to be run repeatedly, with every key stroke). A similar behavior is seen even with a very simple `yup` validation schema for all steps. Additionally, this slowness is also noticeable in older versions of the code, albeit less pronounced.

Introducing a debouncing delay of `500ms` to the form validation so that it will only be run at most once every 500ms.

#### Before:
before

#### After:
after


romanwozniak (approved):
LGTM!

PR #132, opened 2021-11-24
Add optional DisplayName in Experiment Engine props

krithika369:
Adding the `DisplayName` field to the engine props that can be utilised in the Turing UI instead of the `Name`, if set by the experiment engine.
* `engines/experiment/manager/types.go` - Add `DisplayName` property
* `ui/src/router/components/form/components/experiment_config/typeOptions.js` - Use display name in router configuration form
* `ui/src/router/components/configuration/components/experiment_config_section/credentials/CredentialsConfigSection.js` - Use display name in router details view

### DIFF
--- a/api/api/specs/experiment-engines.yaml
+++ b/api/api/specs/experiment-engines.yaml
@@ -132,6 +132,9 @@ components:
       properties:
         name:
           type: "string"
+        display_name:
+          type: "string"
+          description: "Optional display name to be used in the UI"
         type:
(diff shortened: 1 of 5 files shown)


romanwozniak (approved):
LGTM. Any ideas about these changes to `api/go.sum`?

krithika369:
> Any ideas about these changes to api/go.sum?

I had to do `go mod vendor` in the `api` directory to update the dependency on the `engines/experiment`. And it ended up cleaning up the `go.sum` file.

PR #133, opened 2021-11-24
Wrap debounced validation method with useMemo

krithika369:
Follow up from: https://github.com/gojek/turing/pull/131

If the debounced method is re-created upon every re-render, it ends up triggering itself. This causes some unexpected behaviours:
* Validation is not run properly for step # 2 onwards in the Create Router Form (`isValidated` is already set to `true` by one of the older spillover triggers)
* More than 1 validation triggers on the Edit Router Form > Next causes late updates to the UI, ultimately resulting in console errors related to unmounted component.

These are fixed by wrapping the debounced method in a `useMemo` hook.

**Reference:** https://dmitripavlutin.com/react-throttle-debounce/

### DIFF
--- a/ui/src/components/form/validation/multi_section_provider.js
+++ b/ui/src/components/form/validation/multi_section_provider.js
@@ -1,4 +1,10 @@
-import React, { useCallback, useContext, useEffect, useState } from "react";
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 
 import debounce from "lodash/debounce";
 import zip from "lodash/zip";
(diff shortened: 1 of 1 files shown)


ui/src/components/form/validation/multi_section_provider.js
```diff
@@ -49,27 +55,33 @@ export const MultiSectionFormValidationContextProvider = ({
...
-      zip(schemas, contexts).map(([schema, ctx]) => {
-        return !!schema
-          ? new Promise((resolve, reject) => {
-              schema
-                .validate(formData, {
-                  abortEarly: false,
-                  context: ctx,
+  // useMemo here so that debouncedValidate is not recreated frequently, causing the
+  // useEffect hook that initiates the validation to be triggered multiple times.
+  const debouncedValidate = useMemo(
```
romanwozniak, line 60: I actually wanted to point to this in the previous PR, but assumed that it was tested and this function re-creation doesn't cause any issues :) 

You can also move `debouncedValidate` outside of the functional component, so there there is only one and only instance of that function:
```js
const debouncedValidate = debounce((schemas, contexts, formData, setErrors, setIsValidated) => {
    Promise.all(
      zip(schemas, contexts).map(([schema, ctx]) => {
        return !!schema
          ? new Promise((resolve, reject) => {
              schema
                .validate(formData, {
                  abortEarly: false,
                  context: ctx,
                })
                .then(
                  () => resolve({}),
                  (err) => resolve(extractErrors(err))
                );
            })
          : Promise.resolve({});
      })
    )
      .then(setErrors)
      .then(() => setIsValidated(true));
  }, DEBOUNCE_INTERVAL_MS);

export const MultiSectionFormValidationContextProvider = ({...}) => {
  //...
 
  useEffect(() => {
    if (isTouched) {
      if (schemas) {
        debouncedValidate(schemas, contexts, formData, setErrors, setIsValidated]);
      } else {
        setIsValidated(true);
      }
    }
  }, [isTouched, schemas, contexts, formData]);
})

```

romanwozniak: Another option would be to define the validation function with `useCallback` hook to get the referential equality of that function:
```js
export const MultiSectionFormValidationContextProvider = ({...}) => {
  // ...  
  
  const validate = useCallback(() => {
    Promise.all(
      zip(schemas, contexts).map(([schema, ctx]) => {
        return !!schema
          ? new Promise((resolve, reject) => {
              schema
                .validate(formData, {
                  abortEarly: false,
                  context: ctx,
                })
                .then(
                  () => resolve({}),
                  (err) => resolve(extractErrors(err))
                );
            })
          : Promise.resolve({});
      })
    )
      .then(setErrors)
      .then(() => setIsValidated(true));
  }, [schemas, contexts, formData]);

  useEffect(() => {
    if (isTouched) {
      if (schemas) {
        debounce(validate, DEBOUNCE_INTERVAL_MS);
      } else {
        setIsValidated(true);
      }
    }
  }, [isTouched, schemas,validate]);
})
```

romanwozniak (approved):

ui/src/components/form/validation/multi_section_provider.js
krithika369: I clearly didn't test this properly yesterday. :(

Thanks for the suggestions. Moving the function outside of the component seems simple enough, without having to introduce any hooks. So I did that.

I also lowered the debounce interval from `500ms` to `300ms` now (`500` started to feel like the feedback was slow, 300 is better). The performance is comparable between the 2 values.

PR #137, opened 2021-11-29
Update documentation on Turing Init.

ashwinath:
The instructions for the turing-init chart was outdated as it was not published to the public repository then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,22 +345,27 @@ This is the recommended way to install the prerequisites on a production Kuberne
 To install the required components on your Kubernetes cluster, issue the following command:
 
 ```bash
-helm upgrade turing-init infra/charts/turing-init \
-    --set image.registry="<image registry>/" \
-    --set image.repository="<repository>" \
-    --set image.tag="<tag version>" \
+helm repo add turing https://turing-ml.github.io/charts
+helm upgrade turing-init turing/turing-init \
+    --namespace infrastructure \
     --install \
     --wait
 ```
 
 For it to be completely installed, you should check the init job has run successfully.
-To check, issue the command `kubectl get pod` and you should see something like this:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -363,6 +362,12 @@ turing-test-spark-operator-webhook-init-ph8ds   0/1     Completed   0          3
 turing-test-turing-init-init-pknvb              0/1     Completed   0          3m44s
 ```
 
+Programatically, you could check if the init job has run successfully with the following command:
+
+```bash
+kubectl wait -n infrastructure --for=condition=complete --timeout=10m job/turing-init-init
+```
```
romanwozniak, line 369: Maybe this can also be added under `Test the Deployment` section in `charts/turing-init/README.md`

infra/charts/turing-init/README.md.gotmpl
```diff
@@ -0,0 +1,45 @@
...
+{{ template "chart.versionBadge" . }}
+{{ template "chart.appVersionBadge" . }}
+
+{{ template "chart.description" . }}
+
+## Introduction
+
+This Helm chart installs the infrastructure components [Turing](https://github.com/gojek/turing) requires.
+
+{{ template "chart.requirementsSection" . }}
```
romanwozniak, line 12: This requirements section is a bit confusing, as users assume that they have to install required deps upfront. Same applies to `charts/turing`. Wdyt?

romanwozniak (approved):

README.md
ashwinath: Ok added.

infra/charts/turing-init/README.md.gotmpl
ashwinath: Hmm, I have some mixed opinions on this but I do think you are right. It's confusing and not required. Will remove for init + turing charts.

PR #140, opened 2021-12-02
Overridable OpenAPI Spec

ashwinath:
This PR allows overriding of OpenAPI spec in the turing charts so that we can change the experiment type enum without maintaining a separate set of openapi specificiations in our private repository. The other motivation is that we are now able to override the `servers[].url` to allow custom domains other than `http://localhost:8080/v1`.

However, in this PR, we have generalised it so that we can override any portion of the OpenAPI specification as long as it follows the OAS3 specification.

There is no longer a need to copy over the raw OpenAPI spec files (handwritten) and both the validation and the OpenAPI UI will use the same bundled OpenAPI spec file.

In the Github Actions, we now compile the OpenAPI into one bundle file and put it into the Turing API image instead of building it during the uber Turing API image step as it is required for the E2E tests.


Dockerfile
```diff
@@ -9,13 +9,10 @@ RUN apk add --no-cache bash
 USER ${TURING_USER}
 
 ARG TURING_UI_DIST_PATH=ui/build
-ARG SWAGGER_UI_DIST_PATH=api/api/swagger-ui-dist
 
 ENV TURINGUICONFIG_SERVINGDIRECTORY "./turing-ui"
-ENV OPENAPICONFIG_SWAGGERUICONFIG_SERVINGDIRECTORY "./swagger-ui"
```
romanwozniak, line 15: This variable is used by [spf13/viper](https://github.com/spf13/viper) to set a corresponding config value:
https://github.com/gojek/turing/blob/b603d60763fd73d7b0104d5e88d1402b15d0e468/api/turing/config/config.go#L340

If the config value is not passed, then turing API wouldn't serve Swagger UI:

https://github.com/gojek/turing/blob/b603d60763fd73d7b0104d5e88d1402b15d0e468/api/turing/server/application.go#L127

api/config-dev.yaml
```diff
@@ -65,9 +65,3 @@ MLPConfig:
 TuringUIConfig:
   ServingDirectory: ../ui/build
   ServingPath: /turing
-OpenapiConfig:
-  ValidationEnabled: true
-  SpecFile: api/openapi.yaml
-  SwaggerUIConfig:
-    ServingDirectory: api/swagger-ui-dist
-    ServingPath: /api-docs/
```
romanwozniak, line 73: What's the motivation for removing it?

api/turing/config/config.go
```diff
@@ -337,7 +340,10 @@ type OpenapiConfig struct {
 	// SpecFile specifies the file path containing OpenAPI v3 spec
 	SpecFile string
 	// Optional. Defines a configuration to be used for serving Swagger UI as a single-page app
+	// If this is provided, it will use the bundled yaml instead.
 	SwaggerUIConfig *SinglePageApplicationConfig
+	// Optional. Overrides the file before running the Swagger UI.
+	SwaggerUIConfigOverrideFile *string
```
romanwozniak, line 346: Shall it be named as `SpecOverrideFile` or something like lt? It's used not only by Swagger UI, but also for API validation

api/turing/config/config.go
```diff
@@ -20,6 +20,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
+// OpenapiBundleFile is the file name that is generated when the openapi bundle yaml file is generated.
+const OpenapiBundleFile = "openapi.bundle.yaml"
```
romanwozniak, line 24: I'm not sure if we should hardcode it in the go file. The name of the file is also specified in the make file: 
https://github.com/gojek/turing/blob/d8e0b2141f42d5883765e06d28c0116be2086d50/api/api/Makefile#L19
So with that, we will have two sources of truth of what filename should be used (and subsequently, we should update it in two places if we ever decide to rename this file). 
Is it possible to pass this a parameter?

api/turing/config/config.go
```diff
@@ -337,7 +340,10 @@ type OpenapiConfig struct {
 	// SpecFile specifies the file path containing OpenAPI v3 spec
 	SpecFile string
 	// Optional. Defines a configuration to be used for serving Swagger UI as a single-page app
+	// If this is provided, it will use the bundled yaml instead.
```
romanwozniak, line 343: I'm not sure if I can follow the logic here. Your PR assumingly makes it possible to path OAS3 spec file and use it for both API validation and serving Swagger UI. 

And with that, I'm not sure that I understand this comment clearly. `SwaggerUIConfig` was expected to provide the information on how to serve Swagger UI and it should work the same with or without patching of the spec

api/turing/config/config.go
```diff
@@ -460,8 +466,6 @@ func setDefaultValues(v *viper.Viper) {
 
 	v.SetDefault("OpenapiConfig::ValidationEnabled", "true")
 	v.SetDefault("OpenapiConfig::SpecFile", "api/openapi.yaml")
-	v.SetDefault("OpenapiConfig::SwaggerUIConfig::ServingDirectory", "")
-	v.SetDefault("OpenapiConfig::SwaggerUIConfig::ServingPath", "/api-docs/")
```
romanwozniak, line 464: Why do we need to remove this?

api/turing/config/testdata/config-2.yaml
```diff
@@ -27,6 +27,3 @@ Experiment:
         quxkey2-3: quxval2-3-add
     quux: 
       quuxkey1: quuxval1
-OpenapiConfig:
-  SwaggerUIConfig:
-    ServingPath: /swagger-ui
```
romanwozniak, line 32: Why remove this?

infra/charts/turing/templates/_helpers.tpl
```diff
@@ -101,6 +101,12 @@ MLPConfig:
 TuringEncryptionKey: {{ include "turing.encryption.key" . | quote }}
 Sentry:
   DSN: {{ .Values.turing.config.Sentry.DSN | default (include "turing.sentry.dsn" .) | quote }}
+OpenapiConfig:
+  ValidationEnabled: true
+  SwaggerUIConfigOverrideFile: /etc/openapi/override.yaml
+  SwaggerUIConfig:
+    ServingDirectory: swagger-ui
+    ServingPath: /api-docs/
```
romanwozniak, line 109: Having the default values only here is not enough, as it doesn't address a situation, when turing-api isn't deployed from the chart, but is running locally or in a docker container (for development/test purposes)

infra/e2e/turing.values.yaml
```diff
@@ -5,6 +5,14 @@ global:
 
 turing:
   uiConfig: null # Not needed as we only test API
+  # testing of openapi overriding feature
+  openApiComponentOverrides:
```
romanwozniak, line 9: `openApiComponentOverrides` -> `openApiSpecOverrides`. 
`openApiComponentOverrides` make it sound like you can only override `components` part of the spec file, which isn't true

romanwozniak (commented):
I'm still a bit unclear with this PR as it mixes changes to OAS3 spec files AND SwaggerUI serving configuration. Can we decouple these two steps? I.e, can we do it in this way:
1. When we build a `turing-api` image, can we also merge files from `turing/api/api` into a single YAML spec file and include it with the `turing-api` image (but we don't need Swagger UI at this point)
2. When we build a `turing` image, we can generate swagger-ui-dist directory and point it to bundled spec file (from step 1)
3. When turing-api server starts, it can merge existing `openapi.bundle.yaml` in-place with provided patch data

api/config-dev.yaml
ashwinath: 1. It is already set as the defaults and an example of how to override them is written in api/config/example.yaml. I would think that this config-dev.yaml is more of a minimal way of running them.
2. If the api/swagger-ui-dist is not generated, the api will error out because it selects which API validation on the following rule; if this config is not set, use the handwritten openapi yaml spec.

api/turing/config/config.go
ashwinath: I'm wondering why was it set in the first place, on a local setup, `/api-docs/` is non existent, if it's used in the deployment then I have edited the _helpers.tpl to reflect this value. And also as issues mentioned here: https://github.com/gojek/turing/pull/140#discussion_r760959454

api/turing/config/testdata/config-2.yaml
ashwinath: Leaving this thread as the central discussion point: https://github.com/gojek/turing/pull/140#discussion_r760959454

infra/charts/turing/templates/_helpers.tpl
ashwinath: Using this as the central discussion point: https://github.com/gojek/turing/pull/140#discussion_r760959454

infra/e2e/turing.values.yaml
ashwinath: Yes I think this is a fair point, will change this.

api/turing/config/config.go
ashwinath: > I'm not sure if I can follow the logic here. Your PR assumingly makes it possible to path OAS3 spec file and use it for both API validation and serving Swagger UI.

Not sure if I get what you mean but this is true, you can use both the same file for OpenAPI validation + serving Swagger UI.

> And with that, I'm not sure that I understand this comment clearly. SwaggerUIConfig was expected to provide the information on how to serve Swagger UI and it should work the same with or without patching of the spec

This is also true. You can choose to patch or not patch it and it will still work.

api/turing/config/config.go
ashwinath: I'm quite conflicted on this. One one hand this value will never be changed so there is no need to make this configurable. I'm not sure how your solution will be implemented because there will always be two places this name will be hardcoded into.

What would your suggestion be?

Dockerfile
ashwinath: It is set in the _helper.tpl file.

api/turing/config/config.go
ashwinath: ok done

ashwinath:
> I'm still a bit unclear with this PR as it mixes changes to OAS3 spec files AND SwaggerUI serving configuration. Can we decouple these two steps? I.e, can we do it in this way:
> 
>     1. When we build a `turing-api` image, can we also merge files from `turing/api/api` into a single YAML spec file and include it with the `turing-api` image (but we don't need Swagger UI at this point)
> 
>     2. When we build a `turing` image, we can generate swagger-ui-dist directory and point it to bundled spec file (from step 1)
> 
>     3. When turing-api server starts, it can merge existing `openapi.bundle.yaml` in-place with provided patch data

We can but I'm not sure how this solution would help. To me I only see the effects of having a smaller image during the Github Actions but I'm not sure if we gain anything significant. In terms of organisation wise I'm not sure if this does anything better also.

It also adds a lot of noise to this PR.

api/turing/config/config.go
ashwinath: I just remembered, because if these defaults are set, then it will never use the handwritten API specs when developing locally. Most of us will definitely forget to generate the bundled yaml config as per my logic of use bundled file if possible else use the handwritten api spec.

romanwozniak: > I'm wondering why was it set in the first place, on a local setup, /api-docs/ is non existent

`/api-docs/` is a _seerving path_ i.e. swagger-ui will be accessible under <host:port>/api-docs/

> if it's used in the deployment then I have edited the _helpers.tpl to reflect this value

Helm chart doesn't cover local development scenario (when the app runs from a local machine or via docker compose). The default value was set to keep it consistent with https://github.com/gojek/turing/blob/b603d60763fd73d7b0104d5e88d1402b15d0e468/api/turing/config/config.go#L459. If you want to standardize the way how the default values are set, then it should be consistent with other config values.

> I just remembered, because if these defaults are set, then it will never use the handwritten API specs when developing locally. 
We are talking about Swagger UI here, not about the spec file. So, the default value only sets under which path Swagger UI should be served (`ServingDirectory` is still empty, so by default there is noting to serve)

ashwinath: Ok the other alternative is to stat the file if it exists then serve, if it does not exist we give a warning

ashwinath: Actually I'll use your `spaCfg := cfg.OpenapiConfig.SwaggerUIConfig; spaCfg != nil && len(spaCfg.ServingDirectory) > 0` logic.

api/Dockerfile
```diff
@@ -22,6 +32,10 @@ FROM alpine:3.13
 ENV TURING_PORT "8080"
 ENV TURING_USER "turing"
 ENV TURING_USER_GROUP "app"
+ENV TURING_OPENAPI_BUNDLE_FILE "/app/openapi.bundle.yaml"
+ENV OPENAPICONFIG_OPENAPIBUNDLEFILENAME "openapi.bundle.yaml"
+# there is no swaggerui but there will be minimum openapi.bundle.yaml there.
+ENV OPENAPICONFIG_SWAGGERUICONFIG_SERVINGDIRECTORY "/app"
```
ashwinath, line 38: Not very satisfied by this so I'll be thinking of a solution for this.

ashwinath: Changed this to a runtime helm configuration instead of baking it into the docker image.

api/turing/config/config.go
ashwinath: Changed it to an envvar

api/Dockerfile
```diff
@@ -22,6 +32,8 @@ FROM alpine:3.13
 ENV TURING_PORT "8080"
 ENV TURING_USER "turing"
 ENV TURING_USER_GROUP "app"
+ENV TURING_OPENAPI_BUNDLE_FILE "/app/openapi.bundle.yaml"
+ENV OPENAPICONFIG_OPENAPIBUNDLEFILENAME "openapi.bundle.yaml"
```
romanwozniak, line 36: Nit: shall you re-use `openapi.bundle.yaml` value via docker build arg or env var? It's also hardcoded at Line 7. Would be nice to define it only once in case if we will ever need to change it

api/turing/server/api.go
```diff
@@ -69,13 +70,20 @@ func AddAPIRoutesHandler(r *mux.Router, path string, appCtx *api.AppContext, cfg
 
 func openapiValidationMiddleware(apiPath string, cfg *config.Config) (mux.MiddlewareFunc, error) {
 	if cfg.OpenapiConfig != nil && cfg.OpenapiConfig.ValidationEnabled {
+		// Choose between bundled openapi yaml file or development files
```
romanwozniak, line 73: Can we reduce the complexity here and instead just update `.env.development` file to point to development files, while for production builds (i.e. image distributions) use bundled spec file?

api/turing/server/application.go
```diff
@@ -117,6 +119,19 @@ func Run() {
 	// HealthCheck Handler
 	AddHealthCheckHandler(r, "/v1/internal", db)
 
+	// Override the bundle file with user specified overrides.
+	if spaCfg := cfg.OpenapiConfig.SwaggerUIConfig; spaCfg != nil && len(spaCfg.ServingDirectory) > 0 {
```
romanwozniak, line 123: Why are user-specified overrides only applied when `SwaggerUIConfig` is configured? Openapi spec file is not directly related to the swagger UI. Swagger UI is just a convenience added on top of the spec file

api/turing/server/api.go
ashwinath: Removed and reworked.

api/turing/server/application.go
ashwinath: Removed and reworked.

api/turing/config/example.yaml
```diff
@@ -209,7 +209,10 @@ TuringUIConfig:
 # Openapi validation configuration
 OpenapiConfig:
   ValidationEnabled: true
-  SpecFile: "api/openapi.yaml"
+  # This file will be ignored by the OpenAPI validator if SwaggerUIConfig is filled.
```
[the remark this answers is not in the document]
ashwinath: Removed and reworked.

ashwinath:
@romanwozniak I have reworked the configuration into a much simpler way. The following changes have been made:

1. openapi.bundle.yaml is commited we guarantee that the developer will have the bundle files. This can be regnerated by just running `make` in `/api/api` folder.
2. You can provide an override yaml file. The reason why I chose file over configuration was because there is a bug/harmful feature introduced that [lower cased all the keys]()https://github.com/spf13/viper/issues/373), I could not use the fix because i needed to change to spf13/viper instead of ory/viper and i need at least [go 1.16 to make it work](https://github.com/spf13/viper/issues/1161)
3. Configuration is much easier now, it only introduces an additional config that takes in the override file.
4. We serve the new merged in memory YAML, merged or not, and serve it under `/static/openapi.bundle.yaml` (configurable).

romanwozniak:
> 2\. I could not use the fix because i needed to change to spf13/viper instead of ory/viper and i need at least [go 1.16 to make it work](https://github.com/spf13/viper/issues/1161)

Now I wonder why do we use `ory/viper` in the first place (that's a rhetorical question)

api/turing/server/application.go
```diff
@@ -117,8 +118,18 @@ func Run() {
 	// HealthCheck Handler
 	AddHealthCheckHandler(r, "/v1/internal", db)
 
+	// Override the bundle file with user specified overrides and serve files
+	openAPISpecBytes, err := utils.MergeTwoYamls(
+		cfg.OpenapiConfig.SpecFile,
+		*cfg.OpenapiConfig.SpecOverrideFile,
```
romanwozniak, line 124: Can you implement the patching logic as a method of `config.OpenapiConfig` struct?  Something like this:
```go
func (c * OpenapiConfig) SpecData() ([]byte, error) {
  return utils.MergeTwoYamls(c.SpecFIle, c. SpecOverrideFile)
}
```

The power of abstractions helps you to manage the complexity of certain parts of your application by encapsulating the implementation details within the bounds of the abstraction. I.e. I likely don't want to know anything about the specfile or the patching logic of it at the root level of my application, but what I need is a laconic contract on how to retrieve my API specification.

Also, I'm not sure if passing `openAPISpecBytes` into `AddAPIRoutesHandler` is necessary as you already have an instance of `Config` passed into this method, meaning that you can retrieve API spec data from the `cfg.OpenapiConfig`

api/turing/config/example.yaml
```diff
@@ -209,10 +209,13 @@ TuringUIConfig:
...
+  SpecFile: api/openapi.yaml
+  SpecOverrideFile: api/override-sample.yaml
   SwaggerUIConfig:
     ServingDirectory: static/swagger-ui
     ServingPath: /api-docs
+    YAMLServingURL: /static/openapi.bundle.yaml
```
romanwozniak, line 218: Isn't it expected to be at the same level as `SwaggerUIConfig` and not nested under it?

api/turing/config/example.yaml
```diff
@@ -209,10 +209,13 @@ TuringUIConfig:
 # Openapi validation configuration
 OpenapiConfig:
   ValidationEnabled: true
-  SpecFile: "api/openapi.yaml"
+  # This file will be ignored by the OpenAPI validator if SwaggerUIConfig is filled.
```
romanwozniak, line 212: Shall this comment be uprated/removed?

api/turing/config/config.go
```diff
@@ -330,14 +330,20 @@ type MLPConfig struct {
...
 
+// OpenapiConfig contains the settings for the OpenAPI specs used for validation and Swagger UI
 type OpenapiConfig struct {
 	// ValidationEnabled specifies whether to use OpenAPI validation middleware,
 	// which validates HTTP requests against the spec.
 	ValidationEnabled bool
 	// SpecFile specifies the file path containing OpenAPI v3 spec
 	SpecFile string
 	// Optional. Defines a configuration to be used for serving Swagger UI as a single-page app
+	// If this is provided, it will use the bundled yaml instead.
```
romanwozniak, line 341: Outdated comment

infra/charts/turing/templates/openapi-override-configmap.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
ashwinath: not strictly necessary but will prevent creation if it's blank.

api/turing/server/static.go
```diff
@@ -80,5 +81,13 @@ func (h SPAHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 func ServeSinglePageApplication(r *mux.Router, path string, appDir string) {
 	r.PathPrefix(path).Handler(NewSinglePageApplicationHandler(path, appDir))
+}
 
+func ServeOpenAPIYAML(r *mux.Router, url string, openAPISpecBytes []byte) {
```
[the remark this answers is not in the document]
ashwinath: what should the media type be? YAML is not considered some standard media type, but will put it as `application/x-yaml`.

Ref: https://stackoverflow.com/questions/332129/yaml-media-type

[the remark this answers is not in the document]
ashwinath: Will not do `Nit 2` unless there's a compelling reason (i.e. we are serving any other byte arrays) to abstract the logic for serving byte arrays.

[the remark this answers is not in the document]
ashwinath: on second thought i'll do it like how Github does it, type is plain.

api/turing/config/example.yaml
ashwinath: ya good catch.

romanwozniak (approved):

api/turing/server/application.go
ashwinath: Done

api/turing/config/config.go
```diff
@@ -333,14 +334,23 @@ type MLPConfig struct {
...
+	// Config file to be used for serving swagger.
 	SwaggerUIConfig *SinglePageApplicationConfig
+	// Where the merged spec file yaml should be served.
+	YAMLServingPath string
+	// Optional. Overrides the file before running the Swagger UI.
+	SpecOverrideFile *string
+}
+
+func (c *OpenapiConfig) SpecData() ([]byte, error) {
+	return utils.MergeTwoYamls(c.SpecFile, *c.SpecOverrideFile)
```
romanwozniak, line 353: Wouldn't this cause a nil pointer exception when you try to de-reference `nil` `SpecOverrideFile`?

ashwinath: Yeah it would. Let me correct that. Thanks

PR #141, opened 2021-12-03
Removing experiment and treamtent labels from experiment engine metrics

leonlnj:
This PR removes the experiment and treatments for experiment engine metrics.

The flexibility to allow experiment engine to log additional labels which might cause a huge cardinality. 

Also, it is recommended not use dimensions with high cardinality - 
https://prometheus.io/docs/practices/naming/
https://www.robustperception.io/cardinality-is-key
```
CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, 
which can dramatically increase the amount of data stored.
Do not use labels to store dimensions with high cardinality (many different label values), 
such as user IDs, email addresses, or other unbounded sets of values.
```

### DIFF
--- a/engines/router/missionctl/experiment/interceptors.go
+++ b/engines/router/missionctl/experiment/interceptors.go
@@ -37,8 +37,8 @@ func (i *MetricsInterceptor) AfterCompletion(
 	err error,
 ) {
 	labels := map[string]string{
-		"status":     metrics.GetStatusString(err == nil),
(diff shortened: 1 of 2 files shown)


krithika369:
I don't think we want to do this universally and disable these labels for all experiment engines. Rather, we should not set these labels for the problematic internal ones (i.e. XP).

romanwozniak (approved):

PR #143, opened 2021-12-09
Support external experiment engines through the RPC plugins

romanwozniak:
### Context:
At the moment of writing, Turing comes with no publicly available experiment engines, that can be used for setting up experiments and running them with Turing routers. This means, that if users want to deploy their routers with an enabled experiment engine, they first need to implement required interfaces for the [experiment manager](https://github.com/gojek/turing/blob/c6884d5464439584f447927e23707732407282a8/engines/experiment/manager/manager.go#L13) / [experiment runner](https://github.com/gojek/turing/blob/c6884d5464439584f447927e23707732407282a8/engines/experiment/runner/runner.go#L22), then programmatically register their engine implementation and then build turing-api/turing-router from sources to take this into an effect. This is a long and error-prone process that makes the onboarding experience not ideal, at least. 
This PR is our first step to improving it, by adding support of external experiment engine plugins (implemented with https://github.com/hashicorp/go-plugin). With that, plugin, that implements experiment engine contract, can be written, tested, and compiled separately from the Turing codebase, and then supplied to turing-api/turing-router, as an executable binary, at the runtime. 

Simple implementation of a NoOp experiment engine as plugin, is added here: [experiment/examples/plugins/nop](https://github.com/romanwozniak/turing/tree/experiment_engine_plugins/engines/experiment/examples/plugins/nop)

NOTE: Support of plugin-based experiment engines is added on top of the existing way of including engines at the compile-time. It was done for backward compatibility and for the usecases, when the overhead of host<>plugin communication over the local network is not acceptable due to the latency requirements.

### Changes:
* This PR introduces some subtle changes to the `runner.ExperimentRunner` interface to make it return a `Treatment` `struct`, instead of an `interface`, as (de-)serialization of interfaces with [`gob`](https://pkg.go.dev/encoding/gob) is tricky.


engines/router/missionctl/fiberapi/fan_in_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/gojek/fiber"
 	fiberhttp "github.com/gojek/fiber/http"
 	"github.com/gojek/turing/engines/experiment/runner"
+	runnerV1 "github.com/gojek/turing/engines/experiment/v1/runner"
```
leonlnj, line 18: curious, what is the benefit of alias-ing in this fashion with the version? to make it more verbose which version we are using when importing?

romanwozniak: Since both packages (`github.com/gojek/turing/engines/experiment/runner` and `github.com/gojek/turing/engines/experiment/v1/runner`) are sharing the same name (`runner`), we can't import both of them without using an alias. Hence, alias `runnerV1` suggests that this is the `v1/runner` package.

engines/experiment/factory.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/gojek/turing/engines/experiment/plugin"
+	"github.com/gojek/turing/engines/experiment/runner"
+	v1 "github.com/gojek/turing/engines/experiment/v1"
+	"github.com/mitchellh/mapstructure"
+	"go.uber.org/zap"
+)
+
+// EngineFactory interface defines methods for accessing manager/runner
+// of a given experiment engine
+type EngineFactory interface {
```
krithika369, line 16: Though we are combining the manager and runner here, they won't be initialised together, right? For example, EngineFactory's `GetExperimentRunner` call must be made for the runner to be initialised and there wont be any initialisation of the corresponding experiment manager within the router. Is my understanding correct?

engines/experiment/plugin/manager/rpc.go
```diff
@@ -0,0 +1,45 @@
...
+// rpcServer serves the implementation of a ConfigurableExperimentManager
+type rpcServer struct {
+	Impl ConfigurableExperimentManager
+}
+
+func (s *rpcServer) Configure(cfg json.RawMessage, _ *interface{}) (err error) {
+	return s.Impl.Configure(cfg)
+}
+
+func (s *rpcServer) GetEngineInfo(_ interface{}, resp *manager.Engine) error {
```
krithika369, line 42: What about the rest of the experiment manager methods other than `GetEngineInfo`? Will they be implemented in a future PR ?

engines/experiment/plugin/runner/rpc.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (c *rpcClient) GetTreatmentForRequest(
+	ctx context.Context,
+	log runner.Logger,
+	header http.Header,
+	payload []byte,
+) (*runner.Treatment, error) {
+	req := GetTreatmentRequest{
+		Context: ctx,
```
krithika369, line 28: You were saying earlier that there were issues serializing the logger and Go context. Is this working now?

krithika369 (approved):
Looks great! Added some comments / questions for clarification.

engines/experiment/plugin/runner/rpc.go
romanwozniak: Changes to the runner interface will be introduced in a separate PR. I didn't want to mix it in into a somehow big PR. 

>  Is this working now?
It works for imported experiment engines (compile-time plugins), but doesn't work for the RPC plugins.

engines/experiment/plugin/manager/rpc.go
romanwozniak: Currently `ExperimentManager` interface has only a [single `GetEngineInfo` method.](https://github.com/gojek/turing/blob/c6884d5464439584f447927e23707732407282a8/engines/experiment/manager/manager.go#L15) and this is what is implemented here. 
As for the `StandardExperimentManager` and `CustomExperimentManager`, I'll need to give it another thought, as I'm not sure if the current way of casting the types through the `adapter` package is the best way with respect to the plugin implementation.

engines/experiment/factory.go
romanwozniak: Yup, that's right.

engines/experiment/examples/plugins/nop/runner.go
```diff
@@ -0,0 +1,23 @@
+package nop
```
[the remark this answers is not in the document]
romanwozniak: >Is this how we are expected to build the binary for the plugin

Building a plugin's binary is no different compared to building any other Go app. So something like this will do the job:
```shell
go build -o bin/nop ./cmd
```

Developer docs are out of scope for now and will be updated once we settle with all the changes to the manager/runner interfaces.

engines/experiment/plugin/config.go
```diff
@@ -0,0 +1,25 @@
+package plugin
+
+import (
+	"github.com/gojek/turing/engines/experiment/plugin/manager"
+	"github.com/gojek/turing/engines/experiment/plugin/runner"
+	"github.com/hashicorp/go-plugin"
+)
+
+const (
+	ManagerPluginIdentified = "experiment_manager"
```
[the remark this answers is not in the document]
krithika369: fied vs fier 😬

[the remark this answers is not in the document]
romanwozniak: lol, I got it :)

engines/experiment/plugin/manager/rpc.go
krithika369: Okay. Got it.

PR #145, opened 2021-12-14
Do not set knative resource percentage annotation if value <= 0

krithika369:
The annotation `queue.sidecar.serving.knative.dev/resourcePercentage` will not be set on the Knative service if the configured value is <= 0.


romanwozniak (approved):
LGTM 🚀

PR #146, opened 2021-12-15
Add UserContainerLimitRequestFactor config to determine resource limit from request

krithika369:
Currently, the memory and CPU limit for the Knative and Kubernetes deployments (Router, Enricher, Ensembler, Fluentd) for the Turing router are using 2X the request.

This PR introduces the `KnativeServiceDefaults.UserContainerLimitRequestFactor` config so we may vary the Knative deployments' resource limit value, based on the request, using the multiplier (applies to Enricher, Ensembler and Router).

The default Fluentd resources are generally under-utilised. The resource limit is now set to match the request value, by using factor = 1.


api/go.sum
```diff
@@ -754,7 +754,6 @@ github.com/hashicorp/go-cleanhttp v0.5.1/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtng
 github.com/hashicorp/go-hclog v0.0.0-20180709165350-ff2cf002a8dd/go.mod h1:9bjs9uLqI8l75knNv3lV1kA55veR+WUPSiKIWcQHudI=
 github.com/hashicorp/go-hclog v0.8.0/go.mod h1:5CU+agLiy3J7N7QjHK5d05KxGsuXiQLrjA0H7acj2lQ=
 github.com/hashicorp/go-hclog v0.9.2/go.mod h1:5CU+agLiy3J7N7QjHK5d05KxGsuXiQLrjA0H7acj2lQ=
-github.com/hashicorp/go-hclog v0.14.1 h1:nQcJDQwIAGnmoUWp8ubocEX40cCml/17YkF6csQLReU=
```
krithika369, line 757: These changes are caused by `go mod tidy`

romanwozniak:
Can we make limit/request factor configurable? I'm not sure about making it as strict as 1:1, it makes sense to keep some 0.1-0.2 buffer on top of the requested value. 
You can use this implementation for your reference: https://github.com/knative/serving/blob/353f8fb7fbb690e579ff9975ac2c433d8b733d5d/pkg/reconciler/revision/resources/queue.go#L114

krithika369:
@romanwozniak thanks for the comment - I agree with the concern. I've updated the code with your suggestion. Could you have another look?

romanwozniak (approved):
Thanks @krithika369, lgtm!

PR #147, opened 2021-12-20
Update ExperimentManager interface and implement plugin methods for StandardExperimentManager

romanwozniak:
### Context:
This PR is a continuation of https://github.com/gojek/turing/pull/143, where we introduced the general structure for RPC-based experiment engine plugins. This PR updates [ExperimentManager](https://github.com/gojek/turing/pull/147/files#diff-2b339cceca084cf09dc64b9da63fc21f984d14e487cce0a8c8fdf0d3eece6089) interface definition to make it compatible with the plugin implementation and extends RPC plugin (client/server) to support methods of the `StandardExperimentManager` interface.

### Changes:
* `ValidateExperimentConfig`/`GetExperimentRunnerConfig` methods now standardized and moved under `ExperimentManager`. Previously, signatures of these methods were different for `StandardExperimentManager` and `CustomExperimentManager`, which was tricky to support in RPC plugin, because [rpcClient](https://github.com/gojek/turing/pull/147/files#diff-cea59aa522be9a65d9c9d6f7b7664806d2a3e415a41a425ee250b357770188f6) should implement methods of both standard/custom ExperimentManager (to support communication with both types).
* [rpc_server](https://github.com/gojek/turing/pull/147/files#diff-f6b776c4c917954b288811dbe8f976ba328c6228b0943dfeb1197703cca59c85)/[rpc_client](https://github.com/gojek/turing/pull/147/files#diff-cea59aa522be9a65d9c9d6f7b7664806d2a3e415a41a425ee250b357770188f6) were extended to support methods of the `StandardExperimentManager` interface
* The rest is simply changes to existing unit tests and new tests to cover added code parts


api/turing/models/experiment_engine.go
```diff
@@ -6,25 +6,21 @@ import (
...
-	// Type of Experiment Engine. Currently supports "litmus", "xp" and "nop".
-	Type ExperimentEngineType `json:"type"`
+	// Type of Experiment Engine
+	Type string `json:"type"`
 	// Config contains the configs for the selected experiment engine (other than "nop").
-	// For standard experiment engine managers, the config can be cast into TuringExperimentConfig type.
-	Config interface{} `json:"config,omitempty"`
+	// For standard experiment engine managers, the config can be unmarshalled into
+	// manager.TuringExperimentConfig type.
+	Config json.RawMessage `json:"config,omitempty"`
```
krithika369, line 23: If using `json.RawMessage` here, the `Config` value saved in the DB will be a byte array, right?
Also, if I'm not wrong, in the current state of the code, the router / router version API would show the byte data.
I think it would be nice to see the actual config in both places.

krithika369: This would also require the existing data in our DB to be migrated (less of a concern). IMO `interface{}` in the GORM model is not a bad idea. I believe `map[string]interface{}` is more conventional for JSON.

romanwozniak: > If using json.RawMessage here, the Config value in the experiment_engine column in the DB will be displayed as a byte array, right?

yes, but it's not any different from how it works right now, because the entire  [ExperimentEngine](https://github.com/gojek/turing/blob/a1fa8c2d115f7a8fc28ac4b278bb7fa7d87bc0b4/api/turing/models/experiment_engine.go#L31) is stored as a byte array, not just `Config` value.

> Also, if I'm not wrong, in the current state of the code, the router / router version API would show the byte data.
I think it would be nice to see the actual config in both places.

This is something for me to check.

> This would also require the existing data in our DB to be migrated (less of a concern).

No, not required, because of the point above ^

> IMO interface{} in the GORM model is not a bad idea. I believe map[string]interface{} is more conventional for JSON.

So are you suggesting to use stick to `interface{}` here?

krithika369: > yes, but it's not any different from how it works right now

Hmmm.. It _is_ different. In the DB, the `experiment_engine` column is `jsonb` and we marshal the entire data before storing. However, when the jsonb column is queried, we see the decoded data.

But, if ExperimentEngine.Config is a byte array (`json.RawMessage`) in itself, I'm afraid we are going to see the same byte array in the DB, for example, through the `psql` command line.

CMIIW.

> So are you suggesting to use stick to interface{} here?

I now think `map[string]interface{}` is more apt (which is something I hadn't considered earlier, when changing the Config to `interface{}`) if we know it will be json data. Our OpenAPI specs state that the config field will be json.

romanwozniak: > Hmmm.. It _is_ different. In the DB, the `experiment_engine` column is `jsonb` and we marshal the entire data before storing. However, when the jsonb column is queried, we see the decoded data.
> 
> But, if ExperimentEngine.Config is a byte array (`json.RawMessage`) in itself, I'm afraid we are going to see the same byte array in the DB, for example, through the `psql` command line.
> 
> CMIIW.

I'm slightly confused by this. 
The type of `experiment_engine` remains `jsonb`, so you're expected to be able to query and see it as JSON data. Also, it doesn't matter if `ExperimentEngine.Config` is `interface{}` or `json.RawMessage` because in both cases the way how it is passed to the DB's driver is defined by the same [`ExperimentEngine.Value()`](https://github.com/gojek/turing/blob/a1fa8c2d115f7a8fc28ac4b278bb7fa7d87bc0b4/api/turing/models/experiment_engine.go#L30) method, which in both cases just marshals the entire `ExperimentEngine` instance (including its `Config` filed) into a byte array (`[]byte`). Essentially, the `Config` filed (no matter if it's `interface{}` or `json.RawMessage`) is being serialized into a byte array, which in case of `json.RawMessage` is even easier as it's already serialized.

> Also, if I'm not wrong, in the current state of the code, the router / router version API would show the byte data.
I think it would be nice to see the actual config in both places.

What I described above stands correct for the API logic as well. The API response [is serialized into a form of JSON](https://github.com/gojek/turing/blob/a1fa8c2d115f7a8fc28ac4b278bb7fa7d87bc0b4/api/turing/api/response.go#L32) using stdlib's json encoder, so the end result is the same for both `interface{}` and `json.RawMessage`.

romanwozniak: Querying `router_versions` of the existing deployment:
Screenshot 2021-12-21 at 7 48 54 PM

Querying `router_versions` of the local deployment with changes introduced in this PR:
Screenshot 2021-12-23 at 10 42 31 PM

krithika369: After poking around a bit, I realise my source of confusion was in using `[]byte` vs `json.RawMessage`. The former shows something like:
Screenshot 2021-12-24 at 8 49 59 PM

```
type ExperimentEngine struct {
	Type string `json:"type"`
	Config []byte `json:"config,omitempty"` // instead of json.RawMessage
}
```

This is something I had attempted earlier, before deciding to go with `interface{}` for the config. But, I had tried to double marshal the data, which apparently returns the base64 encoded string, and wrongly concluded that would be the same behavior for what happens when the data is saved to the DB. What I had failed to consider and take note of is, marshaling the `[]byte` behaves differently from marshaling `json.RawMessage`, even though the latter is an alias for the former.

A simplified example:
Screenshot 2021-12-24 at 9 01 08 PM

As you've elaborated, the `json.RawMessage` works as intended, both in the saved value in the DB as well the API response. Thanks for clarifying.

engines/experiment/plugin/rpc/manager/rpc_client.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (c *rpcClient) Configure(cfg json.RawMessage) error {
+	return c.Call("Plugin.Configure", cfg, new(interface{}))
+}
+
+func (c *rpcClient) GetEngineInfo() (resp manager.Engine) {
+	err := c.Call("Plugin.GetEngineInfo", new(interface{}), &resp)
+	if err != nil {
+		// err should be propagated upstream, but it's currently not
```
krithika369, line 23: Should we change the function signature of `GetEngineInfo` to return an error instead ?

engines/experiment/plugin/rpc/manager/rpc_client.go
```diff
@@ -0,0 +1,71 @@
...
+	return c.Call("Plugin.ValidateExperimentConfig", cfg, new(interface{}))
+}
+
+func (c *rpcClient) GetExperimentRunnerConfig(cfg json.RawMessage) (resp json.RawMessage, err error) {
+	err = c.Call("Plugin.GetExperimentRunnerConfig", cfg, &resp)
+	return
+}
+
+func (c *rpcClient) IsCacheEnabled() (resp bool) {
+	_ = c.Call("Plugin.IsCacheEnabled", new(interface{}), &resp)
```
krithika369, line 42: Same comment as with `GetEngineInfo` - should we change the signature to capture the error? Perhaps an error from the `rpcClient.IsCacheEnabled` can be discarded, but that decision could be made by the caller.

krithika369 (approved):
The experiment manager interface looks much more concise now. Thanks for the changes. LGTM!

engines/experiment/plugin/rpc/manager/rpc_client.go
romanwozniak: Yup, perhaps it's a better way to deal with it. Addressed

api/turing/validation/validator.go
```diff
@@ -94,18 +94,25 @@ func validateLogConfig(sl validator.StructLevel) {
 }
 
 func newExperimentConfigValidator(expSvc service.ExperimentsService) func(validator.StructLevel) {
+	supportedEngines := make(map[string]bool)
+	supportedEnginesList := []string{models.ExperimentEngineTypeNop}
```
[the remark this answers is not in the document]
romanwozniak: Not sure if I got it right. `supportedEngines` is there for a quick lookup if the engine is supported, but `supportedEnginesList` is only used for reporting the validation error message as (`strings.Join(supportedEngines, ",")`)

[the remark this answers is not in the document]
romanwozniak: I converted `supportedEnginesList` into string here: https://github.com/gojek/turing/pull/147/commits/ad10b183c313c3b0760ac7520527f1de9683e14b so hopefully it's less confusing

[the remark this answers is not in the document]
krithika369: Yes. Given that the valid types are nop + list of supported engines, having nop in one list and not in another similarly named list didn't seem immediately apparent. The new change is clearer. Thanks.

PR #148, opened 2022-01-05
Add list router sdk

DeadlyCoconuts:
Context:
Currently the Turing SDK only supports batch experiments by covering only operations related to ensemblers and ensembling jobs, and we are now expanding its scope to cover online experiments by extending the SDK to cover operations involving other Turing components.

Changes:
This PR involves implementing a functionality corresponding to one of the API endpoints involving routers, i.e. listing the routers belonging to a given project.

- A new SDK class `Router` has been created for users to interact with (this interface exposes some attributes such as `id`, `name`, `project_id`, `config`, etc.)
- The schema in the OpenAPI spec for `ExperimentConfig` has been changed to remove usage of the `oneOf` keyword, which isn't currently supported by the OpenAPI generator 
- The schemas in the OpenAPI spec for certain components in `Router` and its constituent components have been modified slightly to allow for `null` responses and other minor schema differences
- A new test script for the new `Router` class' `list` method with its corresponding test fixtures (which can be reused when extending the SDK to other components)

Note that the massive number of new line additions stem from the autogenerated OpenAPI classes.


api/api/specs/experiment-engines.yaml
```diff
@@ -270,23 +270,6 @@ components:
...
-          type: array
-          items:
-            $ref: "#/components/schemas/Experiment"
-        variables:
-          $ref: "#/components/schemas/ExperimentVariables"
-
-    CustomExperimentConfig:
-      type: "object"
-      not:
-        $ref: "#/components/schemas/StandardExperimentConfig"
```
romanwozniak, line 288: I understand that you are removing this because of some limitations of https://github.com/OpenAPITools/openapi-generator, but this is not a viable solution to remove these schemas completely as they (especially `StandardExperimentConfig`) are helpful source of the API documentation. 
You need to explore other possible solutions to the limitation. Some existing parts of the SDK are already using `oneof` here: https://github.com/gojek/turing/blob/1f22856d7f5fd647ef3bbb75fe889ab73e29d1d0/api/api/specs/ensemblers.yaml#L131-L138
(however I'm not sure if it matches the issue you observed)

api/api/specs/routers.yaml
```diff
@@ -423,12 +424,13 @@ components:
...
           format: "date-time"
           readOnly: true
-        project:
+        project_id:
           type: "integer"
         environment_name:
           type: "string"
         endpoint:
           type: "string"
+          nullable: true
```
romanwozniak, line 433: this likely should be `readOnly`, instead of `nullable`, as it's not editable by users

sdk/turing/generated/model/ensembler1.py
```diff
@@ -0,0 +1,195 @@
...
+)
+
+def lazy_import():
+    from turing.generated.model.ensembler_docker_config import EnsemblerDockerConfig
+    from turing.generated.model.ensembler_standard_config import EnsemblerStandardConfig
+    globals()['EnsemblerDockerConfig'] = EnsemblerDockerConfig
+    globals()['EnsemblerStandardConfig'] = EnsemblerStandardConfig
+
+
+class Ensembler1(ModelNormal):
```
romanwozniak, line 36: There are two models named `Ensembler` in the openapi spec now, hence this `Ensembler1` name here.  
I would suggest to rename https://github.com/gojek/turing/blob/1f22856d7f5fd647ef3bbb75fe889ab73e29d1d0/api/api/specs/routers.yaml#L680
into `EnsemblerConfig`, to follow the name of the DB table where these entities are stored.

api/api/specs/routers.yaml
DeadlyCoconuts: Yep, okay that makes sense; I was thinking of the case whereby routers get created but fail (and have no endpoint) or aren't deployed and `readOnly` still works perfectly in that case.

api/api/specs/experiment-engines.yaml
DeadlyCoconuts: Hmm, okay I made that change thinking that we'll not be explicitly enforcing specific values of `ExperimentEngineType` using an enum. I was previously planning to turn the original `ExperimentConfig` schema into a 'base class' and allow the derived schemas `StandardExperimentConfig` and `CustomExperimentConfig` to inherit from it directly, while using a discriminator on the `ExperimentEngineType` to determine which child schema to use:

```
StandardExperimentConfig:
  allOf:
    - $ref: "#/components/schemas/ExperimentConfig"
    - type: "object"
      properties:
        config:
          type: "object"
          properties:
            project_id:
              type: "integer"
            client:
              $ref: "#/components/schemas/ExperimentClient"
            experiments:
              type: array
              items:
                $ref: "#/components/schemas/Experiment"
            variables:
              $ref: "#/components/schemas/ExperimentVariables"

CustomExperimentConfig:
  allOf:
    - $ref: "#/components/schemas/ExperimentConfig"
    - type: "object"

ExperimentConfig:
  type: "object"
  required:
    - type
  properties:
    type:
      $ref: "#/components/schemas/ExperimentEngineType"
  discriminator:
    propertyName: type
    mapping:
      'nop': "#/components/schemas/StandardExperimentConfig"
      'xp': "#/components/schemas/CustomExperimentConfig"
      'litmus': "#/components/schemas/StandardExperimentConfig"

# Enums
ExperimentEngineType:
  type: "string"
  default: "nop"
```

Would this be a more viable option?

-----

I think the example on the `Ensembler` using `oneOf` didn't encounter any issues with the OpenAPI generator because it only contained 1 schema to choose from (it inherently behaves as an `allOf` in that case), unlike the original specification of `ExperimentConfig`.

api/api/specs/routers.yaml
```diff
@@ -655,6 +666,8 @@ components:
         - port
         - env
       properties:
+        id:
+          type: "integer"
```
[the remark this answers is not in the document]
DeadlyCoconuts: Thanks for bringing that up, should a similar change be applied to the `id` (an integer) in `Event`, `Ensembler` and `Router` too?

api/api/specs/experiment-engines.yaml
romanwozniak: > Would this be a more viable option?

Actually, no, because the list of supported exp engines is dynamic and can vary from one deployment to another. It seems like the support of `oneof` is [finally coming](https://github.com/OpenAPITools/openapi-generator/issues/10812) to openapi generator, so I think for now we can change `ExperimentConfig` to 
```yaml
ExperimentConfig:
      type: "object"
      required:
        - type
      properties:
        type:
          $ref: "#/components/schemas/ExperimentEngineType"
        config:
          type: "object"
```
But at the same time, we can keep `#/components/schemas/StandardExperimentConfig` and  `#/components/schemas/CustomExperimentConfig` definitions in this spec so when `oneof` keyword is supported we can revert it back.

api/api/specs/routers.yaml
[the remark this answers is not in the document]
romanwozniak: For consistency, yes, however it's not crucial. I think the reason why `type: "integer"` is used somewhere is because it's impossible to use `readOnly: true` together with `$ref`

sdk/turing/generated/model/ensembler1.py
DeadlyCoconuts: Ok I'll be renaming this class to `RouterEnsemblerConfig` in order to prevent overlapping with the `EnsemblerConfig` for batching jobs.

PR #150, opened 2022-01-07
Update ci teardown to cluster init

leonlnj:
# Summary
Refactor existing Turing CI workflow, to remove tear down of cluster and do it as an e2e test part of cluster-init workflow.

# Modification 
`.github/actions/cluster-init/action.yaml` - New github action to be called by both Turing and cluster-init workflow to init the cluster. It will contain inputs to optionally build an image or to use an built jar.
The steps involved are:
1. Setup k3d cluster
2. (Optional) Build and save Cluster-Init Image
3. Push image to local registry
4. Install cluster

`.github/workflows/cluster-init.yaml` - Updated to include an e2e test.
Steps
1.  Trigger `cluster-init` github action 
2. Deploy an knative service for smoke test
3. Tear down infra

`.github/workflows/turing.yaml` - Extracted installation of cluster as github action and to call the action instead. Removed tearing down of cluster.

### DIFF
--- a/.github/actions/cluster-init/action.yaml
+++ b/.github/actions/cluster-init/action.yaml
@@ -34,7 +34,7 @@ runs:
     - name: "Setup local k8s cluster"
       uses: AbsaOSS/k3d-action@v1.5.0
       with:
-        cluster-name: ${{ input.cluster_name }}
+        cluster-name: ${{ inputs.cluster_name }}
         use-default-registry: true
         args: >-
           --servers 1
@@ -60,7 +60,7 @@ runs:
 
     - name: Publish images to local registry
       env:
-        DOCKER_REPOSITORY: ${{ input.local_registry }}/${{ github.repository }}
+        DOCKER_REPOSITORY: ${{ inputs.local_registry }}/${{ github.repository }}
         CLUSTER_INIT_VERSION: ${{ steps.build-image.outputs.cluster-init-version }}
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


romanwozniak:
@leonlnj I understand that this is a draft PR, but in case you haven't considered it yet, then my early feedback would be to share common cluster setup steps between `cluster-init/test-e2e` and  `turing/test-e2e` workflows with a [custom Github Action](https://docs.github.com/en/actions/creating-actions/about-custom-actions). This action can encapsulate repeated steps of starting up k3d cluster, adding cluster-init Docker image into a K3d's local image registry, and cluster provisioning with the cluster-init job.

gojek/turing already contains one custom action here: https://github.com/gojek/turing/blob/main/.github/actions/release-rules/action.yaml

leonlnj:
ah thanks, i was still looking around on how to extract common code, that helps!

.github/actions/cluster-init/action.yaml
```diff
@@ -0,0 +1,102 @@
...
+        cluster-name: ${{ inputs.cluster_name }}
+        use-default-registry: true
+        args: >-
+          --servers 1
+          --agents 3
+          --port 80:80@loadbalancer
+          --k3s-server-arg "--no-deploy=traefik,metrics-server"
+
+    - name: Build Docker image
+      if: ${{ inputs.build_image == 'true' }}
```
leonlnj, line 50: Need to do a string check, as it is actually a string and not a boolean - https://github.com/actions/runner/issues/1483

.github/actions/cluster-init/action.yaml
```diff
@@ -0,0 +1,102 @@
...
+    description: 'Knative Istio Version'
+    default: ''
+  local_registry:
+    required: true
+    description: 'Endpoint of local registry'
+    default: ''
+  build_image:
+    required: false
+    description: '(Optional) Flag to build image from code'
+    default: false
```
romanwozniak, line 28: Since we are already building `cluster-init` image in Turing CI workflow, I would rather unify the flow and make it similar to Turing CI. To achieve that, we need:
1. Remove `build_image` input option from this action
2. Add explicit step `build-cluster-init` into `cluster-init` workflow, that builds the image and stores it as `.tar` artifact
3. Make `cluster-init` action to import that docker image into a local registry (it's already done, by using `cluster_init_version` variable)
4. Update `publish` step of the `cluster-init` workflow to load the Docker image from `.tar` artefact, instead of building it again with `make build-image`.

romanwozniak (approved):
Added a suggestion on how to make `cluster-init` workflow aligned with the main Turing CI workflow (to structure it as `build -> test-e2e -> publish`). The rest looks good!

PR #152, opened 2022-01-12
Add create router SDK functionality

DeadlyCoconuts:
### Context:
Currently the Turing SDK only supports batch experiments by covering only operations related to ensemblers and ensembling jobs, and we are now expanding its scope to cover online experiments by extending the SDK to cover operations involving other Turing components. Previously we've already implemented the SDK to supporting the listing of routers in PR #148, and this PR includes additional functionalities involving the setting up of routers.

### Features:
This PR involves implementing a functionality corresponding to one of the API endpoints involving routers, i.e. creating a new router given a router configuration.

- New SDK classes have been created for users to interact with (Route, TrafficRuleCondition, TrafficRule, LogConfig, ResourceRequest, Enricher, RouterEnsemblerConfig, etc.) and to build their own RouterConfig instance incrementally using these components

Example:
```python
router_config = RouterConfig(
        environment_name="id-dev",
        name="router-1",
        routes=[
            Route(
                id="model-a",
                endpoint="http://predict-this.io/model-a",
                timeout="100ms"
            ),
            Route(
                id="model-b",
                endpoint="http://predict-this.io/model-b",
                timeout="100ms"
            )
        ],
        rules=None,
        default_route_id="test",
        experiment_engine=ExperimentConfig(
            type="nop",
            config={
                'variables':
                        [
                            {'name': 'order_id', 'field': 'fdsv', 'field_source': 'header'},
                            {'name': 'country_code', 'field': 'dcsd', 'field_source': 'header'},
                            {'name': 'latitude', 'field': 'd', 'field_source': 'header'},
                            {'name': 'longitude', 'field': 'sdSDa', 'field_source': 'header'}
                        ],
                'project_id': 102
            }
        ),
        resource_request=ResourceRequest(
            min_replica=0,
            max_replica=2,
            cpu_request="500m",
            memory_request="512Mi"
        ),
        timeout="100ms",
        log_config=LogConfig(
            result_logger_type=ResultLoggerType.NOP,
            table="abc.dataset.table",
            service_account_secret="not-a-secret"
        ),
        enricher=Enricher(
            image="asia.test.io/model-dev/echo:1.0.2",
            resource_request=ResourceRequest(
                min_replica=0,
                max_replica=2,
                cpu_request="500m",
                memory_request="512Mi"
            ),
            endpoint="/",
            timeout="60ms",
            port=8080,
            env=[
                EnvVar(
                    name="test",
                    value="abc"
                )
            ]
        ),
        ensembler=DockerRouterEnsemblerConfig(
            id=1,
            image="asia.test.io/gods-test/turing-ensembler:0.0.0-build.0",
            resource_request=ResourceRequest(
                min_replica=1,
                max_replica=3,
                cpu_request="500m",
                memory_request="512Mi"
            ),
            endpoint=f"http://localhost:5000/ensembler_endpoint",
            timeout="500ms",
            port=5120,
            env=[],
        )
    )
```
- The SDK classes also double as objects that automatically get created (as properties of Router instances) when users receive responses from the Turing API containing Router information

Example:
```python
# retrieving a list of routers using the SDK method list() from PR #148 
routers = turing.Router.list()

# get an instance of one of the Router objects returned
my_router = routers[0]

# access attributes of Router instances (such as RouterConfig and Route) and manipulate them directly
my_router.config.routes[0].timeout = "50ms"
```
- These SDK components returned can then be retrieved and reused as individual components for other RouterConfig configurations
- Tests involving all the newly created components

### Modifications
- `sdk/turing/router/*` - new Python files containing the new SDK classes for router components
- `sdk/turing/router/router.py` - addition of a new `Router.create` method to create a new Turing router with a given router configuration
- `sdk/turing/router/session.py` - addition of a new `create_router` method used by `Router.create` to call the API method associated with creating a new Turing router
- `api/api/specs/*` - addition of regex expressions that will be enforced by API classes generated by OpenAPI Generator; acts as a central source of truth for validating user input
- ` sdk/turing/generated/model/*` - autogenerated API classes changed by the OpenAPI generator given the changes made to the OpenAPI specs


sdk/turing/experiment_config.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @property
+    def config(self):
+        return self._config
+
+    @config.setter
+    def config(self, config):
+        if config is not None and 'project_id' in config:
+            config['project_id'] = int(config['project_id'])
+        self._config = config
```
DeadlyCoconuts, line 29: This `project_id` check is a little awkward; it's there because when `project_id` gets returned as a response within the `config` attribute, its value gets parsed as a `float` by the autogenerated API classes, which when sent back as part of a request, receives a 400 Bad Request message since the API expects an `int` for the `project_id` variable.

sdk/turing/router/config/resource_request.py
```diff
@@ -0,0 +1,89 @@
+import turing.generated.models
+from turing.generated.model_utils import OpenApiModel
+from turing.router.config.common.schemas import CpuRequestSchema, MemoryRequestSchema
+
+
+class ResourceRequest:
```
terryyylim, line 6: Considering that most new SDK classes you've created are used for storing data, you might want to consider the use of `dataclasses` which comes in-built with dunder methods `__init__`, `__repr__`, `__eq__` methods that will simplify the generation and storage of attributes.
```
@dataclasses.dataclass
class ResourceRequest:
    min_replica: int
    max_replica: int
    cpu_request: str
    memory_request: str
```

terryyylim (commented):
Thanks for the extensive test suites @DeadlyCoconuts , left a few comments for your consideration!

sdk/tests/conftest.py
```diff
@@ -380,7 +432,103 @@ def router_version(
...
+                        [
+                            {'name': 'order_id', 'field': 'fdsv', 'field_source': 'header'},
+                            {'name': 'country_code', 'field': 'dcsd', 'field_source': 'header'},
+                            {'name': 'latitude', 'field': 'd', 'field_source': 'header'},
+                            {'name': 'longitude', 'field': 'sdSDa', 'field_source': 'header'}
+                        ],
+                'project_id': 102
+            }
+        ),
+        resource_request=turing.router.config.resource_request.ResourceRequest(
```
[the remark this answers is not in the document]
DeadlyCoconuts: Ok sounds good! I've just refactored all the tests in the router module to use classes imported using import statements for the SDK classes so the code should a lot less verbose now. The autogenerated classes are still used via imported modules, i.e. `turing.generated.models.[class_name]` because they usually have the same names as those of the SDK classes...

Another alternative is to import them (the autogenerated classes) with aliases, e.g. ApiResourceRequest or GenResourceRequest, but I'm not sure if that'll be confusing/too wordy.

sdk/turing/router/config/log_config.py
```diff
@@ -0,0 +1,195 @@
...
+    NOP = "nop"
+    CONSOLE = "console"
+    BIGQUERY = "bigquery"
+    KAFKA = "kafka"
+
+    def to_open_api(self) -> OpenApiModel:
+        return turing.generated.models.ResultLoggerType(self.value)
+
+
+class LogConfig:
```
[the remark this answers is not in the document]
DeadlyCoconuts: Hmm I was actually intending for the user the use the child classes (of `LogConfig`), `BigQueryLogConfig` and `KafkaLogConfig` for the time being, to instantiate `LogConfig` objects. These classes only take in the params relative to `bigquery_config` or `kafka_config` respectively and leave the other as None. 

But yeah I'll come up with an additional safeguard to verify the concern you mentioned, so that the user doesn't get away with manipulating the config attributes of `LogConfig` without having the `result_logger_type` set.

sdk/turing/router/config/resource_request.py
DeadlyCoconuts: That's a great idea! I've just refactored all of the SDK classes relating to the router configs as `dataclass`-es. They should now all come with the cheap `__repr__` methods that should be convenient for users.

sdk/tests/router/config/common/schema_test.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+@pytest.mark.parametrize(
+    "value,expected", [
+        pytest.param(
+            "!@#$%^&*()",
+            InvalidKafkaTopicException
+        ),
+        pytest.param(
+            "is_this_a_topic?",
```
romanwozniak, line 131: 613eys

sdk/turing/router/config/common/schemas.py
```diff
@@ -0,0 +1,148 @@
...
+    regex_exp = None
+
+    @classmethod
+    @abstractmethod
+    def verify_regex(cls, value):
+        pass
+
+
+class DockerImageSchema(Schema):
+    regex_exp = r"^([a-z0-9]+(?:[._-][a-z0-9]+)*(?::\d{2,5})?\/)?([a-z0-9]+(?:[._-][a-z0-9]+)*\/)*([a-z0-9]+(?:[._-][a-z0-9]+)*)(?::[a-z0-9]+(?:[._-][a-z0-9]+)*)?$"
```
romanwozniak, line 15: Conceptual question: Can we add all the regex into openapi spec itself (with `patten` keyword) and rely on openapi-generator to do validation for us?
Also, client side validation is good to have, but absolutely not essential to the SDK. It's different from the client-side validation on UI – the reason why we added it to UI is to validate form data dynamically, before user submitted request to the server (i.e. to prevent users from filling the entire form if they provided wrong input to one of the fields). This is motivation doesn't work for the SDK flow, as we will only validate the entire form after python script is executed.
So on the good side, client-side validation helps to report user-friendly error messages, but the drawback of it is that now we have to maintain validation schemas at three different places – API for server-side validation, UI and SDK – for client side validation. Would be nice to see if we can at least merge API and SDK validation into a single source of truth (i.e openapi spec itself)

DeadlyCoconuts: Apparently this is possible, according to the OpenAPI specs [here](https://swagger.io/docs/specification/data-models/data-types/):

> pattern
> The pattern keyword lets you define a regular expression template for the string value. Only the values that match this template will be accepted. The regular expression syntax used is from JavaScript (more specifically, ECMA 262). Regular expressions are case-sensitive, that is, [a-z] and [A-Z] are different expressions. For example, the following pattern matches a Social Security Number (SSN) in the 123-45-6789 format:

I don't think I've seen any issues with using OpenAPI Generator with regex patterns (at least with a quick search on Google) so it is possible to consider unifying all the regex patterns in the OpenAPI spec.

DeadlyCoconuts: As for the client-side validation, I was thinking maybe some users would use the SDK on notebooks so a quick exception thrown out while building the configs (cell by cell) would've been helpful but thanks for clarifying again about the primary usage of the SDK in the form of python script :)

DeadlyCoconuts: Ok it seems like the autogenerated classes validate the text values with the regex expressions correctly; I'll begin moving the schemas to the OpenAPI spec.

api/api/specs/common.yaml
```diff
@@ -29,6 +29,7 @@ components:
       properties:
         name:
           type: "string"
+          pattern: '^[a-zA-Z0-9_]*$'
```
DeadlyCoconuts, line 32: The regex expressions look slightly different from those of the UI as the case insensitive flag `i` doesn't seem to work with OpenAPI generator. As a workaround, I enumerated the different letter cases (i.e. added `A-Z`) in each of the expressions where the case insensitive flag was used.

romanwozniak (approved):
This is great, thanks @DeadlyCoconuts. Could you please also include samples on how to use this SDK under https://github.com/gojek/turing/tree/main/sdk/samples?

DeadlyCoconuts:
> This is great, thanks @DeadlyCoconuts. Could you please also include samples on how to use this SDK under https://github.com/gojek/turing/tree/main/sdk/samples?

Alright sure, I'll add them in in the next PR :)

PR #153, opened 2022-01-13
Update Helm chart to support deployment of Turing App with experiment engine plugins

romanwozniak:
### Context:
In the previous PRs (https://github.com/gojek/turing/pull/143, https://github.com/gojek/turing/pull/147, https://github.com/gojek/turing/pull/149 to name a few) we added support of external experiment engines implemented as RPC plugins. This PR introduces a base Docker image to make it simpler to package and distribute experiment engine plugins and adds necessary changes to Turing's Helm chart to support the deployment of Turing application with such plugins.

### Details:
Experiment engine plugins are supported via [hashicorp/go-plugins](https://github.com/hashicorp/go-plugin), where each plugin is running in a separate process and host application (turing-app) communicates with the plugin over the network. Plugin's process is started by the host application, meaning that the plugin's executable binary should be accessible from the host app. 
To achieve precisely this, we [added a new `emptyDir` volume](https://github.com/romanwozniak/turing/blob/938ee0b09d7c3fd226afb282f43467792ba2431b/infra/charts/turing/templates/deployment.yaml#L94-L95) into the deployment, where plugin executables are expected to be placed. 

Now Turing's user can package their plugin as a docker image:
```
docker build . \
  -t ghcr.io/myrepo/my-exp-engine:latest \
  -f engines/experiment/plugin.Dockerfile \
  --build-arg PLUGIN_NAME=my-exp-engine \
  --build-arg PLUGIN_BINARY=path/to/my/plugin/binary
```

And then deploy it with Turing, by providing `turing.experimentEngines` configuration via Helm values file:
```yaml
turing:
  experimentEngines:
  - name: my-exp-engine
    type: rpc-plugin
    rpcPlugin:
      image: ghcr.io/myrepo/my-exp-engine:latest
    options: {}
    # plugin's configuration  
```

Under the hood, the docker image with the plugin's binary will be added as an `initContainer` on the Turing app deployment, and the binary will be copied over to the `plugins-volume`, where it can be accessed by the host application.

#### Note:
This PR touches only the deployment of Turing app, meaning that the plugin will not be available to the Turing router, after the one is deployed. This will be addressed separately in the next PR.


infra/charts/turing/templates/_helpers.tpl
```diff
@@ -80,6 +80,43 @@
...
 {{- end -}}
 
+{{- define "turing.plugins.directory" -}}
+/app/plugins
+{{- end -}}
+
+{{- define "turing.plugins.initContainers" -}}
+{{ if .Values.turing.experimentEngines }}
+initContainers:
+{{ range $expEngine := .Values.turing.experimentEngines }}
```
leonlnj, line 90: this is to support multiple experiment engines at run time?

krithika369 (approved):

infra/charts/turing/templates/_helpers.tpl
romanwozniak: That's nothing new, Turing can be deployed with multiple experiment engines, and it's up to users to decide which experiment engine to use during the router configuration step

infra/charts/turing/values.yaml
```diff
@@ -100,6 +100,16 @@ turing:
...
     useInClusterConfig: false
 
+  # -- Turing Experiment Engines configuration
+  experimentEngines: {}
+  # Example:
+  # - name: my-exp-engine
+  #   type: rpc-plugin
+  #   rpcPlugin:
+  #     image: ghcr.io/turing/my-exp-engine:latest
+  #   options:
```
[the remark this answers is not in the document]
romanwozniak: Correct. This configuration is later placed (or merged with, if some configuration already exists there) under `turing.config.Experiment.<engine name>`: https://github.com/romanwozniak/turing/blob/938ee0b09d7c3fd226afb282f43467792ba2431b/infra/charts/turing/templates/_helpers.tpl#L142-L148

PR #154, opened 2022-01-17
Use MLP components in the forms

krithika369:
This MR removes the following local form components and replaces their usage with the respective components from the MLP library @ `v1.4.19`:
- ui/src/components/accordion_form
- ui/src/components/form
- ui/src/components/multi_steps_form


romanwozniak (approved):
Thanks!

ui/src/router/edit/components/VersionComparisonView.js
```diff
@@ -46,6 +46,7 @@ export const VersionComparisonView = ({
 
       <EuiFlexItem>
         <StepActions
+          currentStep={1}
```
krithika369, line 49: The decision to show "Cancel" vs "Previous" buttons is done in the library component based on the step number ([ref](https://github.com/gojek/mlp/blob/main/ui/packages/lib/src/components/multi_steps_form/StepActions.js#L20)), as opposed to comparing the value of `onCancel`, as was being done in the Turing component.

PR #155, opened 2022-01-18
Add router management functions to SDK

DeadlyCoconuts:
### Context:
Currently the Turing SDK only supports batch experiments by covering only operations related to ensemblers and ensembling jobs, and we are now expanding its scope to cover online experiments by extending the SDK to cover operations involving other Turing components. Previously we've already implemented the SDK to supporting the listing of routers in PR #148, and the setting up of routers in PR #152. This PR introduces 10 additional SDK methods to support the deletion, retrieval, update, deployment, undeployment of routers and router versions, as well as the listing of deployment events involving a router.

### Features:
- New SDK methods have been created for the `Router` SDK class to support the following methods: 
```
- router deletion
- router retrieval by `router_id`
- router update
- router deployment
- router undeployment
- listing of router versions
- router version deletion
- router version retrieval
- router version deployment
- listing of router deployment events
```
- New SDK child classes `RouterVersion` and `RouterVersionLogConfig` (from their parent `RouterConfig` and `LogConfig` classes introduced in PR #152) to support the read-only operations that return `RouterVersion` objects. In particular the SDK `RouterVersion` class allows users to extract an SDK `RouterConfig` object (without the read-only elements) lying beneath, which can be reused and manipulated as any other `RouterConfig` object 

Example:
```python
# retrieving a router's version using the SDK get_version method
latest_version = my_router.get_version(10)

# extract the RouterConfig object beneath the returned RouterVersion object
latest_config = latest_version.get_config()

# manipulate the extracted RouterConfig object directly (from PR #152)
latest_config.routes[0].timeout = "50ms"

# reuse these RouterConfig objects with other SDK methods
my_router.update(latest_config)
```
- Tests involving all the newly created components
- A sample script showing how the SDK methods can be used 

### Fixes 
- Fix the `update` instance method of the SDK `Router` class so that it updates *all* the attributes of the `Router` instance (`monitoring_url`, `status`, `endpoint`, etc. as opposed to only `config`), according to the `RouterDetails` instance returned by Turing API

### Modifications
- `sdk/turing/router/config/router_version.py` - new `RouterVersion` class 
- `sdk/turing/router/config/log_config.py` - addition of a new `RouterVersionLogConfig` class
- `sdk/turing/router/config/experiment_config.py` - minor modifications to support the creation of OpenAPI `ExperimentConfig` instances while having `config` set as `None`
- `sdk/turing/router/config/router_config.py` - addition of a new method to convert a `RouterConfig`'s (private) attributes into a dict, so that `RouterVersion` can use it to instantiate new `RouterConfig` objects
- `sdk/turing/router/router.py` - addition of new SDK methods corresponding to the 10 new functions mentioned above
- `sdk/turing/router/session.py` - addition of new methods to support the 10 new functions mentioned above
- `api/api/openapi-sdk.yaml` - addition of all the other API endpoints for routers
- `api/api/specs/routers.yaml` - minor modifications to several schemas which were not aligned with the actual Turing API output
- ` sdk/turing/generated/model/*` - autogenerated API classes changed by the OpenAPI generator given the changes made to the OpenAPI specs (~2000 lines)
- `sdk/tests/*` - tests for the new methods and classes
- `sdk/samples/router/general.py` - sample file to demonstrate the use of the all the SDK methods developed for routers
- `sdk/samples/router/create_from_existing_router.py` - sample file to demonstrate how a router config can be retrieved from an existing router to create a new router


api/api/specs/routers.yaml
```diff
@@ -401,9 +401,12 @@ paths:
...
               schema:
-                type: "array"
-                items:
-                  $ref: "#/components/schemas/Event"
+                type: "object"
+                properties:
+                  events:
+                    type: "array"
+                    items:
+                      $ref: "#/components/schemas/Event"
```
DeadlyCoconuts, line 409: Somehow the original schema doesn't align with the responses I'm getting from the API, so I'm aligning this with that.

sdk/turing/router/config/experiment_config.py
```diff
@@ -32,7 +33,11 @@ def config(self, config):
...
 
     def to_open_api(self) -> OpenApiModel:
+        kwargs = {}
+        if self.config is not None:
+            kwargs['config'] = self.config
+
         return turing.generated.models.ExperimentConfig(
             type=self.type,
-            config=self.config
+            **kwargs
```
DeadlyCoconuts, line 42: The autogenerated `ExperimentConfig` class doesn't take `None` for self.config, so I'm hiding away `config` into `kwargs`.

sdk/samples/router/general.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # 2. List all routers
+    routers = turing.Router.list()
+    for r in routers:
+        if r.name == new_router.name:
+            my_router_id = r.id
+        print(r)
+
+    # 3. Get the router you just created using the router_id obtained
+    my_router = turing.Router.get(my_router_id)
```
DeadlyCoconuts, line 204: Some of these steps require waiting for the steps above to 'complete' before they can be run. For example, the one above (step 3) requires the router created in step 1 to be fully deployed before it's able to retrieve a non-`None` router config. I was wondering if I should write the sample in a notebook so that it's clearer that these steps should be run incrementally instead of a continuous script.

romanwozniak:
@gojek/turing-dev a friendly reminder to review this PR

sdk/samples/router/general.py
terryyylim: Alternatively, you could consider writing a generic function which uses `time.sleep` to check whether router has been fully deployed prior to continuing the script.

sdk/samples/router/general.py
```diff
@@ -0,0 +1,265 @@
...
+            cpu_request="500m",
+            memory_request="512Mi"
+        ),
+        endpoint="/echo",
+        timeout="60ms",
+        port=8080,
+        env=[],
+    )
+
+    # # Create the RouterConfig instance
```
terryyylim, line 178: nit: bunch of extra `#` for the comments

sdk/samples/router/general.py
```diff
@@ -0,0 +1,265 @@
...
+    events = my_router.get_events()
+    for e in events:
+        print(e)
+
+    # 12. Delete this router (using its router_id)
+    turing.Router.delete(my_router_id)
+    # # Check if the router still exists
+    for r in turing.Router.list():
+        if r.name == my_router_config.name:
+            print("Oh my, this router still exists!")
```
terryyylim, line 259: It'll be neater to use logging package instead of print statements.

sdk/turing/router/config/log_config.py
```diff
@@ -210,3 +210,17 @@ def to_open_api(self) -> OpenApiModel:
             serialization_format=self.serialization_format.value
         )
         return super().to_open_api()
+
+
+class RouterVersionLogConfig(LogConfig):
```
terryyylim, line 215: Just curious, is there any reason why this implementation and `BigQueryLogConfig`, `KafkaLogConfig` took a different implementation without dataclasses?

sdk/samples/router/general.py
DeadlyCoconuts: Ok I've fixed the `#`es, thanks for pointing that out!

sdk/samples/router/general.py
DeadlyCoconuts: Hmm I was thinking using a basic `print` would be simpler for a sample file that's only meant to be read by users as a rough guide, but okay, using `logging` is fine too :)

sdk/samples/router/general.py
ashwinath: Ya I agree with @terryyylim you could write a blocking function to ensure that the router has been deployed and expose it as part of an Router method.

sdk/samples/router/general.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # 4. Update the router with the new router config
+    my_router.update(my_router_config)
+
+    # 5. List all the router config versions of your router
+    my_router_versions = my_router.list_versions()
+    for ver in my_router_versions:
+        if ver.status == 'undeployed':
+            first_ver_no = ver.version
+        if ver.status == "deployed":
```
ashwinath, line 226: `elif` is preferred here because `ver.status` can be either value but never both at the same time

sdk/samples/router/general.py
```diff
@@ -0,0 +1,265 @@
...
+            latest_ver_no = ver.version
+        print(ver)
+
+    # 6. Deploy a specific router config version (the first one we created)
```
ashwinath, line 230: I would suggest for points 6-8 that you print some debugging information to feedback to the user what is happening when you undeploy and deploy.

sdk/samples/router/general.py
ashwinath: I think for sample this would suffice. Only if it's part of the core SDK then I would suggest using a logger so that the user can disable or set the levels on the logger.

sdk/tests/conftest.py
```diff
@@ -223,17 +223,38 @@ def generic_kafka_config():
...
-        custom_metrics_enabled=True,
-        fiber_debug_log_enabled=True,
-        jaeger_enabled=True,
-        result_logger_type=generic_result_logger_type,
-        bigquery_config=generic_bigquery_config,
-        kafka_config=generic_kafka_config
-    )
+@pytest.fixture(params=["kafka", "bigquery", "others"])
+def log_config(generic_log_level, generic_result_logger_type, generic_bigquery_config, generic_kafka_config, request):
+    if request.param == "kafka":
```
ashwinath, line 228: One trick you can do to remove the verbosity is something like this:
```python
params = dict(
    log_level=generic_log_level,
    custom_metrics_enabled=True,
    fiber_debug_log_enabled=True,
    jaeger_enabled=True,
    result_logger_type=turing.generated.models.ResultLoggerType("kafka"),
    bigquery_config=None,
    kafka_config=generic_kafka_config,
)
if request.param == "bigquery":
    params = {result_logger_type: turing.generated.models.ResultLoggerType("kafka"), **params}
...
return turing.generated.models.RouterVersionLogConfig(**params)
```

sdk/turing/router/config/experiment_config.py
ashwinath: Is there a reason why an empty dictionary would not work?

sdk/samples/router/general.py
DeadlyCoconuts: Ok oops I think I might not have explained it carefully, but this file is really just meant as a sample, so the methods like `get` above wouldn't normally require waiting for the router to be deployed before being able to run (I might imagine someone wanting to get a router that's still `pending`).

What's particular in this sample is that I've laid out the various SDK functions in a certain order so that it demonstrates to the user the creation of *a* router, followed by the retrieval of *that* router by its id, and then the updating of *that* router, etc. So it just so happens that in this 'demo', this `get` should ideally retrieve a router that's already deployed because the next SDK function in the demo is one that requires it (or some other router whose existence I can't guarantee when the user runs the sample) to be deployed, which isn't really the case in normal usage.

sdk/turing/router/config/router_config.py
```diff
@@ -230,3 +233,10 @@ def _verify_no_duplicate_routes(self):
             raise turing.router.config.route.DuplicateRouteException(
                 f"Routes with duplicate ids are specified for this traffic rule. Duplicate id: {most_common_route_id}"
             )
+
+    def to_dict(self):
+        att_dict = {}
+        for m in inspect.getmembers(self):
+            if not inspect.ismethod(m[1]) and not m[0].startswith('_'):
```
ashwinath, line 240: Would be nice to know what m[1] is and m[0] is without looking at the documentation. In this case m[0] is name and m[1] is value. Naming the [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) makes code easier to read. 

Consider this alternative:

```python
# This is better outside of the function
NAME_INDEX = 0
VALUE_INDEX = 1
if not inspect.ismethod(m[VALUE_INDEX]) and not m[NAME_INDEX].startswith('_'):
    att_dict[m[NAME_INDEX]] = m[VALUE_INDEX]
```

It's much easier to read this.

sdk/samples/router/general.py
DeadlyCoconuts: But yes alternatively, it's possible to implement some blocking function into these SDK methods, but that'll fundamentally change the way they operate, e.g. `get` would only work (or would at least wait until) on routers that are `deployed`, instead of being able to retrieve routers on any kind of status (`pending`, `undeployed`, or `failed`)

sdk/turing/session.py
```diff
@@ -12,8 +14,10 @@
     EnsemblersPaginatedResults, \
     EnsemblingJobPaginatedResults, \
     IdObject, \
+    Router, \
     RouterDetails, \
-    RouterConfig
+    RouterConfig, \
+    RouterVersion
```
ashwinath, line 20: no action required but you can actually use parenthesis also: https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses

sdk/turing/router/config/router_config.py
DeadlyCoconuts: Thanks for the tip; I'll take up that suggestion and put that in :)

sdk/samples/router/general.py
ashwinath: > But yes alternatively, it's possible to implement some blocking function into these SDK methods, but that'll fundamentally change the way they operate,

You don't have to bake it into the existing methods, you can create a new method and that method only blocks and does nothing else.

DeadlyCoconuts: Ahh that looks elegant enough for the sample, alright I'll implement that :) Thanks guys!

sdk/samples/router/general.py
```diff
@@ -0,0 +1,308 @@
...
+    response = my_router.undeploy()
+    print(f"7. You have undeployed router {response['router_id']}.")
+
+    # Wait for the router to get undeployed
+    try:
+        my_router.wait_for_status(RouterStatus.UNDEPLOYED)
+    except TimeoutError:
+        raise Exception(f"Turing API is taking too long for router {my_router.id} to get undeployed.")
+
+    time.sleep(30)
```
DeadlyCoconuts, line 269: This call of `sleep` is a little awkward; it's there to somehow prevent this odd case where the deployment of a router fails when it is undeployed and immediately redeployed.

sdk/turing/router/config/experiment_config.py
DeadlyCoconuts: That's a good point; I've just tried it out with an empty dictionary and it works perfectly!

sdk/tests/conftest.py
DeadlyCoconuts: Ok I've refactored it slightly differently so as to ensure `kafka_config` or `bigquery_config` gets properly assigned when their corresponding `result_logger_type` is set:

```python
result_logger_type = generic_result_logger_type.value if request.param == "others" else request.param

params = dict(
    log_level=generic_log_level,
    custom_metrics_enabled=True,
    fiber_debug_log_enabled=True,
    jaeger_enabled=True,
    result_logger_type=turing.generated.models.ResultLoggerType(result_logger_type),
    bigquery_config=None,
    kafka_config=None
)

if request.param == "kafka":
    params["kafka_config"] = generic_kafka_config
elif request.param == "biggquery":
    params["bigquery_config"] = generic_bigquery_config

return turing.generated.models.RouterVersionLogConfig(**params)
```

sdk/turing/router/config/log_config.py
DeadlyCoconuts: My initial concern was with the subclasses (of `LogConfig`) `BigQueryLogConfig`, `KafkaLogConfig` and `RouterVersionLogConfig`  having non-default values for certain arguments, which would create an issue because the default constructor generated for these classes would insert these arguments without default values _after_ those with default values, which would be invalid. 

However, it appears that it's not an issue if the constructors are defined manually, and if the additional attributes are intialised in them. I've thus added the dataclass decorator for these subclasses, as well as for other instances where I originally thought would've posed problems.

api/api/specs/experiment-engines.yaml
```diff
@@ -296,6 +296,12 @@ components:
           $ref: "#/components/schemas/ExperimentEngineType"
         config:
           type: "object"
+        plugin_config:
+          type: "object"
+          nullable: true
+          properties:
+            image:
+              type: "string"
```
DeadlyCoconuts, line 304: Updated OpenAPI specs and added support for `plugin_config` returned as part of `ExperimentConfig` by Turing, as implemented in #158.

romanwozniak: I'm not sure if this needs to be exposed to the users. But for the very least, it should be `read-only`, as it creates a security concern (i.e. Turing users can specify an arbitrary plugin, that will be executed inside the cluster and cause damage. Plugins configuration should be in a full control of Turing operators, ie. those who are deploying Turing).

DeadlyCoconuts: Okay, I'll not specify the schema of `plugin_config` then, and leave it as a arbitrary `read_only` object.

DeadlyCoconuts: Actually, since it's going to be `read-only`, I think I'd just specify the spec for `ExperimentConfig` as follows to remove any mention of `plugin_config` in the OpenAPI spec, as well as in the SDK:

```yaml
ExperimentConfig:
  type: "object"
  required:
    - type
  properties:
    type:
      $ref: "#/components/schemas/ExperimentEngineType"
    config:
      type: "object"
  additionalProperties:
    type: "object"
    readOnly: true
    nullable: true
```

sdk/samples/router/general.py
```diff
@@ -0,0 +1,320 @@
...
+    )
+
+    # Create a log config for the router
+    log_config = LogConfig(
+        result_logger_type=ResultLoggerType.NOP
+    )
+
+    # Create an enricher for the router
+    enricher = Enricher(
+        image="asia.gcr.io/gods-dev/echo:1.0.2",
```
romanwozniak, line 146: That's on a private GCR. Please replace it with a publicly accessible image and keep a note to avoid adding any Gojek specific stuff to the OSS repo

sdk/samples/router/general.py
```diff
@@ -0,0 +1,320 @@
...
+
+    # Wait for the latest version to get deployed
+    try:
+        my_router.wait_for_version_status(RouterStatus.DEPLOYED, latest_ver_no)
+    except TimeoutError:
+        raise Exception(f"Turing API is taking too long for router {my_router.id} with version {latest_ver_no} to get "
+                        f"deployed.")
+
+    # Wait for the dependencies of the first version to be fully undeployed
+    time.sleep(15)
```
romanwozniak, line 254: What is it for? Can it be done as: ?
```python
my_router.wait_for_version_status(RouterStatus.UNDEPLOYED, first_ver_no)
```

sdk/samples/router/general.py
```diff
@@ -0,0 +1,320 @@
...
+    print(f"7. You have undeployed router {response['router_id']}.")
+
+    # Wait for the router to get undeployed
+    try:
+        my_router.wait_for_status(RouterStatus.UNDEPLOYED)
+    except TimeoutError:
+        raise Exception(f"Turing API is taking too long for router {my_router.id} to get undeployed.")
+
+    # Wait for the dependencies of the router to be fully undeployed
+    time.sleep(10)
```
romanwozniak, line 278: Have you seen some issues if this sleep statement is not added? I think the expectation is that router's status is changed to `UNDEPLOYED` only after all the dependencies are undeployed

sdk/samples/router/general.py
DeadlyCoconuts: Ok got that; will keep that in mind in the future.

sdk/samples/router/general.py
DeadlyCoconuts: Yeah I was telling the rest of the team on Thursday that occasionally the deployment of the router/version would take a very long time before failing if it immediately follows an undeployment (without the `sleep` statement). It doesn't occur all time, but somehow spacing out the two actions seems to prevent this from happening. I'll try to see if I can recreate that happening.

sdk/samples/router/general.py
```diff
@@ -0,0 +1,320 @@
...
+    deleted_router_id = turing.Router.delete(my_router.id)
+    print(f"12. You have just deleted the router with id: {deleted_router_id}")
+
+    # Check if the router still exists
+    for r in turing.Router.list():
+        if r.name == my_router_config.name:
+            raise Exception("Oh my, this router still exists!")
+
+
+if __name__ == '__main__':
```
romanwozniak, line 318: This is great!
Can we also explicitly cover (probably, in a separate file) the scenario, when user creates a NEW router, by using an existing router as a base and slightly tweaking its config? Something like:

```python
router_1 = turing.Router.get(my_router.id)

my_router_config = router_1.config

# Modify something in the router config
my_router_config.enricher.env.append(
        EnvVar(
                name="another",
                value="value"
            )
    )

router_2 = turing.Router.create(my_router_config)
```

romanwozniak:
@terryyylim @ashwinath any more comments from you? Are you comfortable approving this PR? I haven't reviewed it end-to-end, so I rely on you to get this approved

sdk/samples/router/general.py
DeadlyCoconuts: Actually this is also similar to the issue right below; I was kind of trying to test both 'potential sources' of error (either with something undeploying being undeployed completely/properly, or with something being deployed not being deployed completely/properly) in both places.

sdk/samples/router/general.py
DeadlyCoconuts: Ok, as Krithika mentioned, so apparently the deployment failure is caused by the knative podautoscaler encountering some delay in clearing some metrics during the undeployment. 

I suppose in that case it's almost impossible for the client to foresee such a scenario from occurring, so should we just include that sleep statement in the sample or should we build in some safeguards in the methods themselves to prevent these? I was thinking the former would be simpler because it's unlikely most of the users would be switching the same routers on and off endlessly

sdk/samples/router/general.py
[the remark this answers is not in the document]
DeadlyCoconuts: Ok! I'll add another explicit example showing that other potential use case!

ashwinath:
> @gojek/turing-dev a friendly reminder to review this PR

I'll rereview it soon.

sdk/samples/router/general.py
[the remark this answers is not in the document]
DeadlyCoconuts: Just added this in.

sdk/turing/router/config/log_config.py
```diff
@@ -105,6 +105,7 @@ class InvalidResultLoggerTypeAndConfigCombination(Exception):
     pass
 
 
+@dataclass
```
ashwinath, line 108: Is there a reason why the dataclass decorator is added?

DeadlyCoconuts: I did it to align its implementation with that of its child class `LogConfig`, which is also a `dataclass`. I'm not sure if your comment is similar to Terence's one above on the same file, but essentially I thought it'd be nice for users to get the free `__repr__` and `__eq__` implementations when manipulating these `LogConfig` derivatives.

ashwinath: Ah ok, no it's not a trick question because I haven't used that decorator before.

ashwinath (approved):
Looks good, just address @romanwozniak's comments and it should be good to go

DeadlyCoconuts:
Thanks everyone for the really detailed and in-depth review and comments

sdk/turing/router/config/log_config.py
DeadlyCoconuts: Ahaha okay xD

sdk/samples/router/general.py
DeadlyCoconuts: -Same comment as the one above-

sdk/samples/router/general.py
DeadlyCoconuts: The sleep statements have now been wrapped into the blocking functions in router to hide this away from the clients,

PR #156, opened 2022-01-18
Allow Swagger UI to use file instead of serving it on a URL

ashwinath:
The main motivation for this PR is that whenever we deploy our application, swagger-ui is [referencing a fixed URL for openapi.bundle.yaml](https://github.com/gojek/turing/blob/1de5d8397d885d8342fc71b9a477a034c9d4a885/scripts/swagger-ui-generator/swagger-ui-generator.sh#L136), which does not allow for any URL rewrites when using a reverse proxy. Letting Swagger UI handle it based on file names allows the use of relative paths (which can be manipulated by a reverse proxy) rather than fixed static paths.

### DIFF
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -343,8 +343,8 @@ type OpenapiConfig struct {
 	SpecFile string
 	// Config file to be used for serving swagger.
 	SwaggerUIConfig *SinglePageApplicationConfig
-	// Where the merged spec file yaml should be served.
-	YAMLServingPath string
+	// Where the merged spec file yaml should be saved.
+	MergedSpecFile string
 	// Optional. Overrides the file before running the Swagger UI.
 	SpecOverrideFile *string
 }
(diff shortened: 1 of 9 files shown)


api/turing/config/config.go
```diff
@@ -354,6 +356,21 @@ func (c *OpenapiConfig) SpecData() ([]byte, error) {
...
 }
 
+// GenerateSpecFile gets the spec data and writes to a file
+func (c *OpenapiConfig) GenerateSpecFile() error {
+	b, err := c.SpecData()
+	if err != nil {
+		return err
+	}
+
+	err = os.MkdirAll(filepath.Dir(c.MergedSpecFile), 0755)
```
romanwozniak, line 366: minor, but wouldn't `0644` be enough here?

romanwozniak (approved):
lgtm

api/turing/config/config.go
ashwinath: I think it would be enough, but the rest of the folders in the image are 755 so I thought I would keep it consistent.

PR #157, opened 2022-01-19
Fix turing docker

ashwinath:
Cherry picked from @romanwozniak to fix the docker builds. Turing API was not able to write to the swagger ui folder because the permissions were tied to root rather than the turing user.

Since the CI doesn't cover this, I have built and tested the image locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,36 @@
 ARG TURING_API_IMAGE
+FROM bash:alpine3.14 as swagger-ui
+
+WORKDIR /app
+
+ARG openapi_spec_file="/app/swagger-ui/openapi.bundle.yaml"
+
+# Build swagger ui deps
+COPY ./scripts/swagger-ui-generator /app/swagger-ui-generator
+
+RUN cd /app/swagger-ui-generator && ./swagger-ui-generator.sh \
+    --spec-file ${openapi_spec_file} \
+    --output /app/swagger-ui
+
 FROM ${TURING_API_IMAGE}
+
+ARG openapi_spec_file=/app/swagger-ui/openapi.bundle.yaml
+ARG turing_ui_dist_path=ui/build
 
 ARG TURING_USER=${TURING_USER:-turing}
 ARG TURING_USER_GROUP=${TURING_USER_GROUP:-app}
 
 ENV TURINGUICONFIG_SERVINGDIRECTORY "/app/turing-ui"
 
-USER root
-RUN apk add --no-cache bash
-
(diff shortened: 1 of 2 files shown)


romanwozniak:
> Since the CI doesn't cover this, I have built and tested the image locally

There is a manual trigger for running CI on your branch and publish images into `ghcr.io/<your_repo>/turing:<version>`:
Screenshot 2022-01-19 at 11 36 15

Dockerfile
```diff
@@ -1,34 +1,36 @@
...
 
 ARG TURING_USER=${TURING_USER:-turing}
 ARG TURING_USER_GROUP=${TURING_USER_GROUP:-app}
 
-ENV TURINGUICONFIG_SERVINGDIRECTORY "/app/turing-ui"
+ARG openapi_spec_file=/app/swagger-ui/openapi.bundle.yaml
+ARG turing_ui_dist_path=ui/build
 
-USER root
-RUN apk add --no-cache bash
```
romanwozniak, line 10: `bash` is still required, because this image uses a custom entrypoint, that requires bash shell

ashwinath: Interesting, ok

ashwinath:
> or running CI on your branch and publish images into `ghcr.io/<your_repo>/turing:<ver`

Oh I mean the runtime part, where the full image runs the Turing API binary and performs the merge.

Dockerfile
romanwozniak: I expected it to be reproducible when you tested it locally

ashwinath: i used /bin/sh with my mounted config

ashwinath: Hmm doesn't seem like it, I'm removing this bash requirement: https://github.com/ashwinath/turing/blob/d929942aa09f03fc78c206167c9ca1135af9e2a3/infra/charts/turing/templates/deployment.yaml#L65-L74

romanwozniak: Entrypoint is defined inside the Dockerfile. Args you posted are passed into it as arguments:
https://github.com/gojek/turing/blob/70fb3a0b63aac2dd9d3f8d972588c9eedb16eda9/Dockerfile#L34

ashwinath: ok, I missed this part. the way i tested it was like this `docker run --net host --rm -ti -v ${PWD}:/stuff --entrypoint /bin/sh asia.gcr.io/gods-dev/ashwin-hack/turing:15`. Which was why i missed it out.

romanwozniak (approved):

PR #158, opened 2022-01-19
Support deployment of Turing routers with RPC Experiment Engine plugins

romanwozniak:
### Context:
https://github.com/gojek/turing/pull/153 made it possible to deploy Turing API with experiment engines implemented as RPC plugins, however, these plugins were not propagated into turing routers, deployed by the API. This PR addresses this part.

### Details
Deployment of Turing API with RPC plugins is supported through the k8s `initContainer`s: during the deployment, plugin binaries from init containers are copied into a shared `emptyDir` volume, which later is mounted into the main `turing-api` container. However, neither [`initContainer`s](https://github.com/knative/serving/issues/12024), nor [shared volumes](https://github.com/knative/serving/issues/11664) are currently fully supported by the Knative (which is used for deploying routers), hence it requires a workaround to overcome these limitations. 

There are two parts to this workaround:
* Support of `PluginURL` configuration [was added into the ExperimentEngine factory](https://github.com/romanwozniak/turing/blob/b0058e5186e47e851ccdcda8a2b491201a30f0e4/engines/experiment/factory.go#L47-L60), to make it possible for the factory to fetch plugin's binary from the provided URL
* Together with the router, Turing API deploys `plugins-server` k8s service (`nginx` server), that serves plugin executables as static files via HTTP

So if turing-router is configured with RPC-based experiment engine, then during the initialization, router fetches plugin's binary from the `plugins-server`, stores it inside the container, and then starts the regular communication between host (turing-router) and the plugin (experiment-runner).


.github/workflows/helm-chart.yaml
```diff
@@ -8,6 +8,18 @@ on:
...
+    inputs:
+      owner:
+        description: The GitHub user or org that owns this repository
+        type: string
+        required: true
+      repository:
+        description:  The GitHub repository
+        type: string
+        required: true
+        default: "charts"
```
romanwozniak, line 21: To make it possible to manually trigger `publish-chart` pipeline from the dev branch

.github/workflows/turing.yaml
```diff
@@ -36,10 +36,130 @@ env:
   ARTIFACT_RETENTION_DAYS: 7
 
 jobs:
+  build-api:
```
romanwozniak, line 39: Changed the flow of CI to speed it up.
**Before:**
`test -> build -> e2e-test -> publish`
**Now:**
`build -> test & e2e -> publish` 

This reduces the average pipeline execution from 20m to 14m

krithika369: Since `test` has no dependency on `build`, can it be `build & test -> e2e -> publish` ? With this, the expensive e2e can be skipped if the unit tests fail.

api/turing/config/config.go
```diff
@@ -259,23 +263,9 @@ type RouterDefaults struct {
...
-	// specification for routers.
-	//
-	// For example:
-	// {"experiment_engine_a": {"endpoint": "http://engine-a.com", "timeout": "500ms"},
-	//  "experiment_engine_b": {"endpoint": "http://engine-b.com", "timeout": "250ms"} }
-	Experiment map[string]interface{}
-
+	FluentdConfig       *FluentdConfig
 	MonitoringURLFormat *string
+	Plugins             map[string]*ExperimentEnginePluginConfig `validate:"dive"`
```
krithika369, line 268: Nit: `Plugins` doesn't convey the purpose immediately. Change to `ExperimentEnginePlugins` or just `ExperimentEngines` ?

api/turing/config/config_test.go
```diff
@@ -178,84 +179,72 @@ func TestLoad(t *testing.T) {
...
 				},
-				RouterDefaults: &RouterDefaults{
+				RouterDefaults: &config.RouterDefaults{
 					LogLevel: "INFO",
-					FluentdConfig: &FluentdConfig{
+					FluentdConfig: &config.FluentdConfig{
 						Tag:                  "turing-result.log",
 						FlushIntervalSeconds: 60,
 					},
-					Experiment: map[string]interface{}{
```
krithika369, line 214: Is there no corresponding `Plugins` config test now?

api/turing/models/events.go
```diff
@@ -79,3 +79,50 @@ func (e *Event) SetRouter(router *Router) {
 func (e *Event) SetVersion(version uint) {
 	e.Version = version
 }
+
+type EventChannel struct {
```
krithika369, line 83: Hmmm.. I think the `models` package has been implicitly reserved for the representation of the DB models. I agree that the event channel related methods were not appropriate for the `utils` package, as they were placed earlier. I wonder if they can be moved into the service layer instead, in `api/turing/service/event_service.go` or even a new package `events` of its own.

infra/charts/turing/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.2.4-dev0.1
+version: 0.2.4-dev0.2
```
krithika369, line 4: General question: I see we are using the pre-release `-dev` labelling on the version now. What's our convention with this? In general, the chart versions would not be coupled to the app releases, right? How do we meaningfully track the chart version along with the app version (I guess this is something we only worry about for breaking changes to the chart)?

krithika369 (approved):
Left some small comments / questions. LGTM!

.github/workflows/turing.yaml
romanwozniak: I thought about this, but ended up with the current setup for two reasons:
* `build & test -> e2e -> publish` pipeline will take  roughly 3.5mins longer to complete, because `build & test` step will require `max(jobs in build&test)` time to complete, which is ~6min, as compared to `max(jobs in build)` time (~2.5min)
* There is no limits on parallel jobs or runner minutes in GitHub Actions, so `expensive` here is not very relevant

infra/charts/turing/Chart.yaml
romanwozniak: There is no convention as of now. I started using this `-dev` suffix in https://github.com/gojek/turing/pull/153 to make a point, that this chart version is not intended to be a stable chart release (as it didn't support the deployment with plugins fully). Also, I'm publishing these dev charts into my personal GitHub repo for testing, before the PR is fully done, so I can test the deployment with plugins in the GKE environment.
I will update the helm chart with a stable version before merging this PR.

api/turing/models/events.go
romanwozniak: I moved it out from `utils` package, because (after my changes from this PR) it caused a cyclic import issue. Agree with `event_service.go` as a better place for it, will make a change. Thanks!

api/turing/config/config_test.go
romanwozniak: I intend to add more tests to this PR, to cover the `plugins_server` part of `servicebuilder` package. Will address it there too, thanks!

PR #161, opened 2022-01-26
Removing references to any experiment engines from the codebase

romanwozniak:
### Context:
When Turing codebase was initially open-sourced, it still contained some mentions of the private experiment engines in its codebase. Together with this, it had a special treatment of `Litmus` experiment engine, which resulted in a requirement to set `LITMUS_PASSKEY` environment variable in turing routers. This PR removes such legacy code.


ashwinath (approved):
From what I understand this just looks like no major logic changes and just removal of Litmus into a generic experiment engine.

PR #163, opened 2022-01-28
Add assets for router versions.

ashwinath:
Fixes the assets missing from the [Turing router history page](https://github.com/gojek/turing/blob/main/docs/how-to/viewing-routers/history.md).

Old:
Old

New:
This PR


krithika369 (approved):
Could you also help to run a check on the asset names and what's referenced in the `docs/` folder? For example, in my local version of the code, I now have 63 hits for `.png` but only 50 images in the `.gitbook/assets` folder. Just wondering if we missed something else on the first import. Other than this, LGTM. Thank you!

ashwinath:
> Could you also help to run a check on the asset names and what's referenced in the `docs/` folder? For example, in my local version of the code, I now have 63 hits for `.png` but only 50 images in the `.gitbook/assets` folder. Just wondering if we missed something else on the first import. Other than this, LGTM. Thank you!

Some are used more than once so it won't be a 1:1 relation but rather a 1 png to many references relation.

PR #164, opened 2022-02-08
Update abstract PyFunc to utilise returned treatment_config

deadlycoconuts:
### Context
Turing SDK offers an abstract class `PyFunc` from which the concrete `PyFuncEnsembler` is implemented. In particular, this `PyFunc` class implements the abstract method `predict` from its parent `mlflow.pyfunc.PythonModel`, which gets called whenever the `mlflow` model is used to generate predictions (or in this case, ensembling results).

This `predict` method partially implements the batch ensembling logic, i.e. it performs some simple `DataFrame` manipulations by separating a single input `DataFrame` into `DataFrames` corresponding to `features`, `predictions` and `treatment_config`, that subsequently get passed to a user-defined `ensemble` method, which contains the rest (and the crux) of the ensembling logic.

Previously, the `treatment_config` field has always been unused, since in batch ensembling, a treatment can equivalently be defined in the features columns, hence the `None` value being passed to `ensemble` when it gets called by `predict`.

However, as we are moving to implement real-time pyfunc ensemblers that ultimately use the same `PyFuncEnsembler` class (which would allow a user to use the very same ensembler in both batch and real-time ensembling, when the same batch columns/live payload naming convention is used), there is a need to handle any `treatment_config` that appears in a live-ensembling request. 

This PR thus aims to expose this `treatment_config` to the user in the `ensemble` method (which has previously always been a null object), allowing users to manipulate the payload from a live Turing router with a user defined ensembling method in a more intuitive manner.

### Extra Context
In batch ensembling, `predictions` are currently passed to `predict` through the `model_input` argument. These predictions are contained in columns with the `__predictions__` header, due to a tabular join operation upstream by other batch ensembling components. In order to not break the existing naming convention of columns in `model_input`, the future engine for the real-time ensembler will similarly pass `predictions` as well as `treatment_config` as columns in `model_input` with the same naming convention.

### Features
- Extraction of `treatment_config` data that gets passed to `predict` as part of the `model_input` argument into a separate `DataFrame`, which then gets passed to the `ensemble` method as the `treament_config` argument:
- Refactoring of a few lines of existing code used to extract column names with the `__[prefix_name]__` prefix into a static method:
```python
@staticmethod
def _get_columns_with_prefix(df: pandas.DataFrame, prefix: str):
    selected_columns = {
        col: col[len(prefix):]
        for col in df.columns if col.startswith(prefix)
    }
    return selected_columns
```

### Modifications
- `sdk/turing/ensembler.py` - addition of logic to consider `treatment_config` being passed to `predict` as part of the `model_input` argument

### DIFF
--- a/sdk/turing/ensembler.py
+++ b/sdk/turing/ensembler.py
@@ -57,8 +57,8 @@ class PyFunc(EnsemblerBase, mlflow.pyfunc.PythonModel, abc.ABC):
 
     def predict(self, context, model_input: pandas.DataFrame) -> \
             Union[numpy.ndarray, pandas.Series, pandas.DataFrame]:
-        prediction_columns = PyFunc._get_columns_with_header(model_input, PyFunc.PREDICTION_COLUMN_PREFIX)
-        treatment_config_columns = PyFunc._get_columns_with_header(model_input, PyFunc.TREATMENT_CONFIG_COLUMN_PREFIX)
+        prediction_columns = PyFunc._get_columns_with_prefix(model_input, PyFunc.PREDICTION_COLUMN_PREFIX)
+        treatment_config_columns = PyFunc._get_columns_with_prefix(model_input, PyFunc.TREATMENT_CONFIG_COLUMN_PREFIX)
 
         return model_input \
             .rename(columns=prediction_columns) \
@@ -71,10 +71,10 @@ class PyFunc(EnsemblerBase, mlflow.pyfunc.PythonModel, abc.ABC):
                    ), axis=1, result_type='expand')
 
     @staticmethod
-    def _get_columns_with_header(df: pandas.DataFrame, header: str):
+    def _get_columns_with_prefix(df: pandas.DataFrame, prefix: str):
         selected_columns = {
-            col: col[len(header):]
-            for col in df.columns if col.startswith(header)
+            col: col[len(prefix):]
+            for col in df.columns if col.startswith(prefix)
         }
         return selected_columns
 
--- a/sdk/turing/ensembler.py
+++ b/sdk/turing/ensembler.py
@@ -40,8 +40,8 @@ class PyFunc(EnsemblerBase, mlflow.pyfunc.PythonModel, abc.ABC):
     Abstract implementation of PyFunc Ensembler.
     It leverages the contract of mlflow's PythonModel and implements its `predict` method.
     """
-
     PREDICTION_COLUMN_PREFIX = '__predictions__'
+    TREATMENT_CONFIG_COLUMN_PREFIX = '__treatment_config__'
 
     def load_context(self, context):
         self.initialize(context.artifacts)
@@ -57,19 +57,26 @@ class PyFunc(EnsemblerBase, mlflow.pyfunc.PythonModel, abc.ABC):
 
     def predict(self, context, model_input: pandas.DataFrame) -> \
             Union[numpy.ndarray, pandas.Series, pandas.DataFrame]:
-        prediction_columns = {
-            col: col[len(PyFunc.PREDICTION_COLUMN_PREFIX):]
-            for col in model_input.columns if col.startswith(PyFunc.PREDICTION_COLUMN_PREFIX)
-        }
+        prediction_columns = PyFunc._get_columns_with_header(model_input, PyFunc.PREDICTION_COLUMN_PREFIX)
+        treatment_config_columns = PyFunc._get_columns_with_header(model_input, PyFunc.TREATMENT_CONFIG_COLUMN_PREFIX)
 
         return model_input \
             .rename(columns=prediction_columns) \
+            .rename(columns=treatment_config_columns) \
             .apply(lambda row:
                    self.ensemble(
-                       features=row.drop(prediction_columns.values()),
+                       features=row.drop(prediction_columns.values()).drop(treatment_config_columns.values()),
                        predictions=row[prediction_columns.values()],
-                       treatment_config=None
+                       treatment_config=row[treatment_config_columns.values()]
                    ), axis=1, result_type='expand')
+
+    @staticmethod
+    def _get_columns_with_header(df: pandas.DataFrame, header: str):
+        selected_columns = {
+            col: col[len(header):]
+            for col in df.columns if col.startswith(header)
+        }
+        return selected_columns
 
 
 @ApiObjectSpec(turing.generated.models.Ensembler)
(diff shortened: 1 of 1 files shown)


sdk/turing/ensembler.py
```diff
@@ -57,20 +57,27 @@ def initialize(self, artifacts: dict):
...
                    self.ensemble(
-                       features=row.drop(prediction_columns.values()),
+                       features=row.drop(prediction_columns.values()).drop(treatment_config_columns.values()),
                        predictions=row[prediction_columns.values()],
-                       treatment_config=None
+                       treatment_config=row[treatment_config_columns.values()]
                    ), axis=1, result_type='expand')
 
+    @staticmethod
+    def _get_columns_with_header(df: pandas.DataFrame, header: str):
```
romanwozniak, line 74: Is it `header` or `prefix`? What's header in this context?

romanwozniak (approved):
Generally, it LGTM, but I have some concerns such as:
1. Even though it's expected that PyFunc ensemblers might be not as performant as custom ensemblers implemented with more performant languages, we should still try to squeeze the maximum out of it and avoid unnecessary pandas DF manipulation. 
2. We can afford to introduce breaking changes to the current PyFunc interface (`ensemble` method) if we need it to achieve better performance. 
3. `treatment_config` in the `ensemble` method is unstructured `dict`, however it's expected to be of a certain structure: https://github.com/gojek/turing/blob/main/docs/how-to/create-a-router/configure-ensembler.md. It could be a good time to update the `PyFunc` interface to reflect this.

sdk/turing/ensembler.py
deadlycoconuts: Yeah I think `prefix` is the right word I was intending to use; I'll update the names in a bit

sdk/turing/ensembler.py
```diff
@@ -26,9 +26,16 @@ def ensemble(
...
+        :param treatment_config: Optional[pandas.Series], representing the configuration of a
+                treatment, that should be applied to a given record/payload.
+
+                For batch experiments, the `treatment_config` can be `None`. If an attribute
+                representing a treatment plan of all the treatments and their respective
+                configuration is set using the `initialize` method, the `treatment_config`
+                for any treatment can be accessed during all calls made to `ensemble`.
+
+                For real-time experiments, the `treatment_config` is either a pandas.Series
+                object, or None, if the experiment engine is not configured.
```
deadlycoconuts, line 38: @romanwozniak I'm not sure if this what you referred to by updating the interface with the changes to `treatment_config`. Basically it's now (soon) going to be a `pandas.Series` object containing treatment config info, just as how `features` and `predictions` contain data within as a `pandas.Series` object. If you were referring to something like a payload unpacking step for `treatment_config` because we're expecting a payload with a standard structure, that part's going to be handled by the future real-time ensembler engine.

romanwozniak: > If an attribute representing a treatment plan of all the treatments and their respective configuration is set using the `initialize` method, the `treatment_config` for any treatment can be accessed during all calls made to `ensemble`.

I'm not sure if I follow this. Can we for now don't touch `treatment_config` for batch experiments at all? This requires a much deeper understanding of how batch experiments are run and how to connect ensembling jobs with the experiment engine.

> Basically it's now (soon) going to be a pandas.Series object containing treatment config info, just as how features and predictions contain data within as a pandas.Series object. If you were referring to something like a payload unpacking step for treatment_config because we're expecting a payload with a standard structure, that part's going to be handled by the future real-time ensembler engine.

I can't remember saying, that `treatment_config` should necessarily be a `pandas.Series`. I think we'll get a better understanding of how this all clicks together once we are moving closer to the integration of all pieces together (basically, after pyfunc ensembler server is ready and after you'll be able to package it as a docker image and deploying it as a custom ensembler in Turing)

deadlycoconuts: > I'm not sure if I follow this. Can we for now don't touch treatment_config for batch experiments at all? This requires a much deeper understanding of how batch experiments are run and how to connect ensembling jobs with the experiment engine.

Okay right now we aren't touching anything involving the batch experiments, so there shouldn't be any issue with the current changes on batch ensembling. 

As to what I was trying to describe in what you quoted, I was having in mind this batch ensembling example I found which shows an implementation of an ensembler with the `PyFunc` interface. It looks something like this:
```python
class MyEnsembler(PyFunc):
    def initialize(self, artifacts: dict):
        # store information about all treatments
        import yaml
        with open(artifacts[ENSEMBLER_CONFIG_PATH_ARTIFACT_KEY], 'r') as f:
            self.config = yaml.safe_load(f)

    def ensemble(self, features: pd.Series, predictions: pd.Series, treatment_config: Optional[dict]) -> Any:
        treatment = features['treatment']
        # access the treatment config stored in the self.config attribute, relevant to the treatment 
        treatment_config = self.config['treatments'][treatment]  
        
        if 'model_name' in treatment_config:
            # do something prescribed by the stored treatment_config
        else:
            # do something default
```
Essentially I was trying to say that for batch experiments, the current way `treatment_config`s are stored is through an attribute that gets initialised.

> I can't remember saying, that treatment_config should necessarily be a pandas.Series. I think we'll get a better understanding of how this all clicks together once we are moving closer to the integration of all pieces together (basically, after pyfunc ensembler server is ready and after you'll be able to package it as a docker image and deploying it as a custom ensembler in Turing)

I think I was trying to say that `treatment_config` is going to be a pandas.Series, if `ensemble` is called by the real-time ensembling engine. And you're right I think the details will be clearer when I bring up the new implementation of the real-time ensembling web service.

romanwozniak: > Essentially I was trying to say that for batch experiments, the current way treatment_configs are stored is through an attribute that gets initialised.

This is not how it's expected to work in the long run, hence we should not include this into the codebase.
In the long run, the gap between experiment engine and batch ensembler should be bridged and treatment configs should be retrieved from the exp engine before ensembling happens, probably as a separate transformation on the spark DataFrame. i.e. something like this:
```
treatmentsDF = spark.DataFrame(source).apply(fetch_treatment_udf)

result = spart.DataFrame(source) \
  .join(predictionsDF) \
  .join(treatmentsDF) \
  .apply(ensemble_udf)
```

deadlycoconuts: Okay then I'll remove the part of the docstrings that mention about the batch ensembling `treatment_config` workaround. Thanks a lot for the clarification!

PR #165, opened 2022-02-11
Create ensembler web service

deadlycoconuts:
### Context
In PR #164, the abstract class `PyFunc` was modified in preparation for the introduction of a real-time pyfunc ensembler engine. 

This PR introduces the aforementioned real-time ensembler engine which is in effect a web service wrapped around by a container that serves real-time ensembling requests using an ensembler defined in the `mlflow` pyfunc flavour. This ensembler is to be created using the Turing SDK by implementing the same `PyFunc` class currently used for implementing batch ensemblers. Hence with these changes, a user would be able to implement an pyfunc ensembler that works in both batch and real-time ensembling (if the same batch columns/live payload naming convention is used).

### Features
- New real-time ensembler engine that runs the `ensemble` method of child classes of the `PyFunc` class
- A web service using the `tornado` framework to serve ensembling requests sent from a Turing router, by using the real-time ensembler engine
- A Docker image to package the web service as a base image, with a separate Docker image (to build on to the former) to load the user-defined ensembler artefacts and run the web service in a container
- A CI/CD workflow on GitHub for the new engine

### Main Additions
- `engines/real-time-ensembler/pyfunc_ensembler_runner/ensembler_runner.py` - Class that holds the real-time ensembler engine
- `engines/real-time-ensembler/pyfunc_ensembler_runner/handler.py` - Class containing the HTTP handler (according to the `tornado` framework) that handles post requests from a Turing router
- `engines/real-time-ensembler/pyfunc_ensembler_runner/server.py` - Class containing the web service
- `engines/real-time-ensembler/pyfunc_ensembler_runner/__main__.py` - Main entry point that runs the web service
- `engines/real-time-ensembler/Dockerfile` - Base image for the web service; does not contain artefacts of the pyfunc ensembler
- `engines/real-time-ensembler/app.Dockerfile` - Docker image that builds upon the base image by loading artefacts of a given pyfunc ensembler
- `sdk/turing/ensembler.py` - Changes to the `predict` method of the `PyFunc` class to process the incoming arguments differently according to their input type (`pandas.DataFrame` for batch ensembling vs `dict` for real-time ensembling)
- `.github/workflows/real-time-ensembler.yaml` - GitHub workflow that tests and publishes the base Docker image containing the real-time ensembler engine


engines/real-time-ensembler/app.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+COPY --from=builder /ensembler ./ensembler
+COPY --from=builder /pyfunc_ensembler_runner ./pyfunc_ensembler_runner
+COPY --from=builder /run.sh /run.sh
+COPY --from=builder /venv /venv
+
+RUN /bin/bash -c ". /venv/bin/activate && \
+    python -m pyfunc_ensembler_runner --mlflow_ensembler_dir /ensembler --dry_run"
+
+CMD ["/bin/bash", "./run.sh"]
```
romanwozniak, line 30: Can we simply activate conda env by default so all of the commands will be executed in that env already activated? In this case you don't need this extra `run.sh` file. 

This is how it was done in batch ensemble image: https://github.com/gojek/turing/blob/c1303487fde3af8704fa03cf561e969db013530e/engines/batch-ensembler/Dockerfile#L96

engines/real-time-ensembler/pyfunc_ensembler_runner/ensembler_runner.py
```diff
@@ -0,0 +1,85 @@
...
+        return treatment_config
+
+    @staticmethod
+    def _flatten_json(y: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Helper function to normalise a nested dictionary into a dictionary of depth 1 with names following the
+        convention: key_1.key_2.key_3..., with a period acting as a delimiter between nested keys
+
+        Items in lists have their names rendered using their index numbers within the lists they are found in.
+        """
```
romanwozniak, line 62: I'm not very comfortable with this. It seems like `dict` could be a better container for this data, compared to `pandas.DataFrame/pandas.Series`. The reason for this is:
1. Latency – for real-time ensemblers, the latency requirement is much more critical, compared to batch ensembling jobs. Having all this json manipulation and transformation into a pandas containers is likely something we'd better avoid
2. This "flat-json" is not something the end-users would expect to work with. Also, JSON keys could have `.` in their names, which is not handled here

With that said, re-working the interface of the PyFunc ensembler makes more sense, and then `batch-ensembler` can be updated too, to transform pandas data into `dict`

engines/real-time-ensembler/pyfunc_ensembler_runner/handler.py
```diff
@@ -0,0 +1,31 @@
...
+        response = self.ensembler.predict(request)
+
+        response_json = orjson.dumps(response)
+        self.write(response_json)
+        self.set_header("Content-Type", "application/json; charset=UTF-8")
+
+    @staticmethod
+    def validate_request(request: Any) -> Dict[str, Any]:
+        try:
+            body = orjson.loads(request.body)
```
romanwozniak, line 25: Is this the same parser, that @gojek/merlin is using in their PyFunc model server?

engines/real-time-ensembler/requirements.txt
```diff
@@ -0,0 +1,5 @@
+argparse>=1.4.0
+cloudpickle==1.2.2
+orjson==2.6.8
+tornado==6.1
+file:../../sdk
```
romanwozniak, line 5: I think we need to consider moving to versioned sdk packages. I.e., instead of using the latest sources of the Python SDK package, we can add it as a PyPi wheel dependency (`turing-sdk==0.2.0`). 
Maybe this can be done once we fix the contract of the `PyFunc` ensembler.

.github/workflows/real-time-ensembler.yaml
```diff
@@ -0,0 +1,109 @@
+name: engines/real-time-ensembler
+
+on:
+  # Automatically run CI on Release and Pre-Release tags and main branch
+  # (only if there are changes to relevant paths)
+  push:
+    tags:
+      - "real-time-ensembler/v[0-9]+.[0-9]+.[0-9]+*"
```
romanwozniak, line 8: nit: I understand that you followed the naming convention of `batch-ensembler`, hence picked up the `real-time-ensembler` name for this module. Maybe we should consider changing this to `pyfunc-ensembler-job` and `pyfunc-ensembler-service` for a better representation of what it does?

deadlycoconuts: Ok those parts have been refactored; all mentions of `batch-ensembler` or `real-time-ensembler`s in the entire turing repo have been worked around to their new names/references `pyfunc-ensembler-job` and `pyfunc-ensembler-service` respectively.

engines/real-time-ensembler/pyfunc_ensembler_runner/handler.py
deadlycoconuts: Yep that's right

engines/real-time-ensembler/app.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+# Use conda-pack to create a standalone enviornment
+# in /venv:
+RUN conda-pack -n real-time-ensembler -o /tmp/env.tar && \
+  mkdir /venv && cd /venv && tar xf /tmp/env.tar && \
+  rm /tmp/env.tar
+
+RUN /venv/bin/conda-unpack
+
+FROM debian:latest
```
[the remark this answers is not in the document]
deadlycoconuts: Oh yeah nice catch; I'll replace that image with its slim version.

engines/real-time-ensembler/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM continuumio/miniconda3 AS builder
+
+RUN wget -qO- https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-367.0.0-linux-x86_64.tar.gz  | tar xzf -
+ENV PATH=$PATH:/google-cloud-sdk/bin
+COPY . .
+COPY ./temp-deps/sdk ./../../sdk
+
+RUN conda env create -f ./environment.yaml &&  \
+    conda env update --name real-time-ensembler --file /ensembler/conda.yaml && \
```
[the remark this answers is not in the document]
deadlycoconuts: Oops seems like I've missed that out; I've updated the Dockerfiles to utilise `CONDA_ENV_NAME` as a variable carrying the name `pyfunc_ensembler_service`.

engines/real-time-ensembler/app.Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+COPY --from=builder /ensembler ./ensembler
+COPY --from=builder /pyfunc_ensembler_runner ./pyfunc_ensembler_runner
+COPY --from=builder /run.sh /run.sh
+COPY --from=builder /venv /venv
+
+RUN /bin/bash -c ". /venv/bin/activate && \
+    python -m pyfunc_ensembler_runner --mlflow_ensembler_dir /ensembler --dry_run"
```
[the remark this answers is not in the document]
deadlycoconuts: Oh it's just an option to run the web service (to load the ensembler from the mlflow registry and the runner) without actually serving it. It's... actually kinda pointless but I observed @gojek/merlin having it for their `pyfunc-server`s so I was wondering if that option would serve some other separate/greater purpose, which was why I included it to be safe. 

But I can remove it also since I can't seem to find any reason why we should run it once without serving it to ensure it's working, before actually running the service again for serving.

engines/real-time-ensembler/app.Dockerfile
deadlycoconuts: Okay that's possible too, alright I'll do everything within the `Dockerfile`

engines/real-time-ensembler/requirements.txt
deadlycoconuts: What was the reason why we decided to work with importing the `sdk` directory directly instead of relying on versioned packages?

romanwozniak: Because there were no `sdk` before we started working on batch ensembling jobs :) Hence, 1) Nothing was published to PyPi at that stage 2) The `sdk` interfaces were fluid and has been changed frequently, together with changes to the `pyfunc-ensembler-job` implementation. So, it made sense to keep them together by making `batch-ensembler` always be dependent on the latest version of the SDK

engines/real-time-ensembler/app.Dockerfile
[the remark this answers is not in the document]
romanwozniak: > I was wondering if that option would serve some other separate/greater purpose, which was why I included it to be safe.

I think the reason is that if you run this command at the image building step, then failure of this `dry-run` step will prevent image from being even published and hence we will not spend more time trying to deploy the image if it's misconfigured

engines/real-time-ensembler/pyfunc_ensembler_runner/ensembler_runner.py
deadlycoconuts: Okay with regards to the latency concerns for real-time ensemblers, I've reworked (overloaded) the `predict` method of the `PyFunc` base class a little to make it work differently depending on whether it gets passed a `dict` (when called by a real-time ensembler engine) or a `pandas.DataFrame` (when called by a batch-ensembler engine) as its argument:

https://github.com/deadlycoconuts/turing/blob/1bce96a181324636367236df3e9ad6d98c285c68/sdk/turing/ensembler.py#L73

With this change, I've completely removed any of those `pandas.DataFrame` transformations like `flatten_json` or other preprocessing table operations that might introduce latency in the real-time ensemblers.

---
While it'd be nice to unify both the batch and real-time ensembling use cases to act on and return `dict` inputs, the `predict` method, as an inherited method of the `mlflow.pyfunc.PythonModel` abstract class, gets utilised by other downstream methods such as `mlflow.pyfunc.spark_udf`, that forces the `predict` method to follow a certain contract by taking `pandas.DataFrame`s as input and output. 

In particular, the current batch ensembling engine uses `mlflow.pyfunc.spark_udf`, which necessitates at least some implementation of `predict` that takes in and outputs `pandas.DataFrame`s. For now I've kept our original implementation intact to prevent the batch ensembler from breaking, but I'll see if I can find a solution that's more elegant.

engines/real-time-ensembler/requirements.txt
deadlycoconuts: Ah alright, in this case for this PR, I'll leave the dependency on the module of the SDK as it is, i.e. `file:../../sdk` since there's actually a tiny change made to the SDK module in this PR. 

Once this PR has been merged, I'll cut a new release for the SDK, and after that raise a separate PR to update the `requirements.txt` to utilise the version for that latest release. This should permanently move us away from having the actual SDK module in the same repo as a dependency.

romanwozniak: Cool. Yes, we can definitely move the the PyPi dependency once the changes to SDK are finalised

sdk/turing/ensembler.py
```diff
@@ -55,20 +55,27 @@ def initialize(self, artifacts: dict):
...
-        return model_input \
-            .rename(columns=prediction_columns) \
-            .rename(columns=treatment_config_columns) \
-            .apply(lambda row:
-                   self.ensemble(
-                       features=row.drop(prediction_columns.values()).drop(treatment_config_columns.values()),
-                       predictions=row[prediction_columns.values()],
-                       treatment_config=row[treatment_config_columns.values()]
-                   ), axis=1, result_type='expand')
+    def predict(self, context, model_input: Union[pandas.DataFrame, Dict[str, Any]]) -> \
```
romanwozniak, line 58: Can we split this method into two private methods with strict signature types? I.e.
```
def _ensemble_batch(self, model_input: pandas.DataFrame) -> ...

def _ensemble_service(self, model_input: Dict[str, Any]) -> ...
```
And then call these methods from inside the `predict` methods, based on the actual type of the `model_input`

romanwozniak (approved):
This looks great. Thanks, @deadlycoconuts!

engines/real-time-ensembler/app.Dockerfile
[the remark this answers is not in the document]
deadlycoconuts: Thanks for the clarification! Hmm but yet again in this particular instance, there isn't much difference between a `dry-run` and a 'normal-run' (i.e. we aren't calling additional methods or running any compute-heavy stuff), so there really isn't much opportunity to 'catch' an error early by performing the `dry-run` -- anything that fails in the dry run will also make the 'normal-run' fail, which stops the build process, and by extension any deployment. 

Maybe it'll just be a tiny bit more elegant without the `dry-run`?

sdk/turing/ensembler.py
deadlycoconuts: Ok that's done :)

deadlycoconuts:
Alright thanks @romanwozniak for the useful comments; I'm gonna merge this now 🚀

PR #168, opened 2022-02-17
[BUGFIX] Fix experiment engine plugins in routers with the traffic rules 

romanwozniak:
### Context: 
Turing supports [Traffic Rules](https://github.com/gojek/turing/blob/main/docs/how-to/create-a-router/configure-traffic-rules.md) via Fiber's [LazyRouter](https://github.com/gojek/fiber/blob/main/lazy_router.go), where each route in this LazyRouter is another Fiber component (either [EagerRouter](https://github.com/gojek/fiber/blob/main/eager_router.go) (when standard or no ensembler is used) or [Combiner](https://github.com/gojek/fiber/blob/main/combiner.go) (when Docker ensembler is used)). The configuration of each route contains all the information that this traffic rule's route requires, including the routes activated when this traffic rule is matched and the configuration of the experiment engine, if it's configured on the router. The entire configuration could look something like this:

```yaml
id: my-router
routes:
  - id: traffic-split-default
    routes:
      - endpoint: https://run.mocky.io/v3/5a431baf-95e8-4e45-a3e4-c5018a0ec107
        id: control
        timeout: 2s
        type: PROXY
    strategy:
      properties:
        default_route_id: control
        experiment_engine: my-engine
        experiment_engine_properties:         
          plugin_url: http://plugins-server:8080/plugins/my-engine
          timeout: 1s
      type: fiber.DefaultTuringRoutingStrategy
    type: EAGER_ROUTER
  - id: traffic-split-0
    routes:
      - endpoint: https://run.mocky.io/v3/5a431baf-95e8-4e45-a3e4-c5018a0ec107
        id: control
        timeout: 2s
        type: PROXY
      - endpoint: https://run.mocky.io/v3/fb120af8-0127-4fde-8e42-288958b37ad4
        id: route-a
        timeout: 1.5s
        type: PROXY
    strategy:
      properties:
        default_route_id: control
        experiment_engine: my-engine
        experiment_engine_properties:         
          plugin_url: http://plugins-server:8080/plugins/my-engine
          timeout: 1s
      type: fiber.DefaultTuringRoutingStrategy
    type: EAGER_ROUTER  
strategy:
  properties:
    default_route_id: traffic-split-default
    rules:
      - conditions:
          - field: X-Route
            field_source: header
            operator: in
            values: ["A"]
        route_id: traffic-split-0
  type: fiber.TrafficSplittingStrategy
type: LAZY_ROUTER
```  

### Problem:
From the example above, it's seen, that `routes.0.strategy` == `routes.1.strategy` which is expected because the experiment configuration is common for each traffic rule within the same router. The problem, however, happens when the router is using an RPC plugin to interact with the experiment engine. 

In this case, each traffic rule will launch the same RPC plugin in a separate process, which isn't really necessary. Also, when the router is configured to download the plugin from a remote URL (see. `routes.0.strategy.properties.experiment_engine_properties.plugin_url`), in certain cases, it fails to initialize the ExperimentRunner, because, during the initialization of multiple traffic rule, each route's Fiber component tries to download the plugin binary into the same location which may leave the binary in a broken state.

### Fix:
This PR fixes the described issue by keeping track of the experiment engines factories that have already been initialized and preventing new factories from being created when the same factory already exists in the turing-router process. This is achieved by storing the references to the initialized factory objects in the local key-value registry.

### DIFF
--- a/engines/experiment/plugin/rpc/factory.go
+++ b/engines/experiment/plugin/rpc/factory.go
@@ -3,15 +3,23 @@ package rpc
 import (
 	"encoding/json"
 	"fmt"
+	"net/url"
+	"path"
 	"reflect"
 	"sync"
 
+	"github.com/gojek/turing/engines/experiment/config"
 	"github.com/gojek/turing/engines/experiment/manager"
+	"github.com/gojek/turing/engines/experiment/pkg/utils"
 	"github.com/gojek/turing/engines/experiment/plugin/rpc/shared"
 	"github.com/gojek/turing/engines/experiment/runner"
 	"github.com/hashicorp/go-plugin"
+	"github.com/mitchellh/hashstructure/v2"
 	"go.uber.org/zap"
 )
+
+var factoriesmu sync.Mutex
+var factories = make(map[string]*EngineFactory)
 
 // EngineFactory implements experiment.EngineFactory and creates experiment manager/runner
 // backed by net/rpc plugin implementations
@@ -74,7 +82,44 @@ func (f *EngineFactory) GetExperimentRunner() (runner.ExperimentRunner, error) {
 	return f.runner, nil
 }
 
-func NewFactory(pluginBinary string, engineCfg json.RawMessage, logger *zap.SugaredLogger) (*EngineFactory, error) {
+func NewFactory(name string, cfg config.EngineConfig, logger *zap.SugaredLogger) (*EngineFactory, error) {
+	factoriesmu.Lock()
+	defer factoriesmu.Unlock()
+
+	// get a hash of the engine's configuration and use it as a fingerprint
+	cfgHash, err := hashstructure.Hash(cfg, hashstructure.FormatV2, nil)
+	if err != nil {
+		return nil, err
+	}
+	factoryKey := fmt.Sprintf("%s-%d", name, cfgHash)
+
+	if engineFactory, ok := factories[factoryKey]; ok {
+		return engineFactory, nil
+	}
+
+	engineCfg, err := cfg.RawEngineConfig()
+	if err != nil {
+		return nil, err
+	}
+	if cfg.PluginBinary != "" {
+		factories[factoryKey], err = NewFactoryFromBinary(cfg.PluginBinary, engineCfg, logger)
+	} else if cfg.PluginURL != "" {
+		factories[factoryKey], err = NewFactoryFromURL(cfg.PluginURL, engineCfg, logger)
+	} else {
+		err = fmt.Errorf("either `plugin_url` or `plugin_binary` must be specified")
+	}
+
+	if err != nil {
+		return nil, err
+	}
+	return factories[factoryKey], nil
+}
+
+func NewFactoryFromBinary(
+	pluginBinary string,
+	engineCfg json.RawMessage,
+	logger *zap.SugaredLogger,
+) (*EngineFactory, error) {
 	rpcClient, err := Connect(pluginBinary, logger.Desugar())
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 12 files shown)


deadlycoconuts:
Looks good to me! Not sure if you'd want someone else to approve it though 😅

deadlycoconuts (approved):
Looks good to me! Not sure if you'd want someone else to approve it though 😅

engines/experiment/plugin/rpc/factory.go
```diff
@@ -3,16 +3,24 @@ package rpc
...
 	"reflect"
 	"sync"
 
+	"github.com/gojek/turing/engines/experiment/config"
 	"github.com/gojek/turing/engines/experiment/manager"
+	"github.com/gojek/turing/engines/experiment/pkg/utils"
 	"github.com/gojek/turing/engines/experiment/plugin/rpc/shared"
 	"github.com/gojek/turing/engines/experiment/runner"
 	"github.com/hashicorp/go-plugin"
+	"github.com/mitchellh/hashstructure/v2"
```
deadlycoconuts, line 17: Probably insignificant question, but is there a reason we're using this package to generate hashes over others?

romanwozniak: No strong preferences, but we are already using https://github.com/mitchellh/mapstructure hence I didn't think too long before bringing this package in as well

engines/experiment/plugin/rpc/factory.go
```diff
@@ -74,7 +82,44 @@ func (f *EngineFactory) GetExperimentRunner() (runner.ExperimentRunner, error) {
 	return f.runner, nil
 }
 
-func NewFactory(pluginBinary string, engineCfg json.RawMessage, logger *zap.SugaredLogger) (*EngineFactory, error) {
+func NewFactory(name string, cfg config.EngineConfig, logger *zap.SugaredLogger) (*EngineFactory, error) {
+	factoriesmu.Lock()
+	defer factoriesmu.Unlock()
+
+	// get a hash of the engine's configuration and use it as a configuration's fingerprint
+	cfgHash, err := hashstructure.Hash(cfg, hashstructure.FormatV2, nil)
```
krithika369, line 90: Is the name not enough to uniquely identify which experiment engine has been seen before? Why go through the trouble of hashing and comparing all properties?

deadlycoconuts: Actually, what's the intended behaviour if the `plugin_url` is the same across various routes, but if the other parameters in `strategy` are different?

krithika369: Hmm... I think from the config part, the `experiment_engine` is the engine name and `experiment_engine_properties` are the `cfg` (@romanwozniak CMIIW). And I don't think we will ever configure different `experiment_engine_properties` for the same `experiment_engine` which is why I wonder why the name alone cannot be used to determine whether an exp engine has been seen before (and therefore being initialised already).

```yaml
        experiment_engine: my-engine
        experiment_engine_properties:         
          plugin_url: http://plugins-server:8080/plugins/my-engine
          timeout: 1s
```

romanwozniak: > Actually, what's the intended behaviour if the plugin_url is the same across various routes, but if the other parameters in strategy are different?

Since `fiber.yml` is constructed by the Turing API, this scenario is not possible. But in case turing-router is deployed manually with a hand-crafted config file, then different instances of the `EngineFactory` should be created and that's the reason why we use a hash of the entire config struct.

krithika369: Hmm.. In that case, since the local download path is constructed like this: `filename := fmt.Sprintf("./%s", path.Base(downloadURL.Path))`, I'd think we should then limit to comparing this effective `filename` value (timeout shouldn't matter to re-trigger the download and not sure if the name does either). Anyway, not a big issue since routers created from our app wont encounter this.

romanwozniak: I simplified this example, in the real-world scenario you'll get more properties to compare:
```yaml
experiment_engine: my-engine
experiment_engine_properties:         
    plugin_url: http://plugins-server:8080/plugins/my-engine
    timeout: 1s
    key1: value1
    key2: value2
   ...
```

Since all engine configuration [is stored inside the factory object](https://github.com/romanwozniak/turing/blob/01bc1266067d1491578a8a2f6b2173cf0a4735e1/engines/experiment/plugin/rpc/factory.go#L32), we need to compare all the properties. Otherwise, in the situation when you have the same plugin, but configured with the different config values (only possible in the manual router deployment), then in both cases, the same factory object will be used, hence the experiment runner would not be properly configured.

But I think that in this scenario, it would be nice to download the plugin's binary under a different path to prevent conflicting downloads.

PR #170, opened 2022-02-22
Automate ensembler image building with Turing API

deadlycoconuts:


api/turing/api/deployment_controller.go
```diff
@@ -231,6 +235,65 @@ func (c RouterDeploymentController) deployRouterVersion(
...
+			return "", fmt.Errorf("failed to find specified ensembler: %w", err)
+		}
+
+		pyFuncEnsembler := ensembler.(*models.PyFuncEnsembler)
+
+		// Build image corresponding to the retrieved ensembler
+		request := imagebuilder.BuildImageRequest{
+			ProjectName: project.Name,
+			// Create a unique resource name that contains partial versioning (part of the run_id hash)
+			ResourceName: pyFuncEnsembler.Name + "-" + pyFuncEnsembler.RunID[:5],
```
deadlycoconuts, line 255: This has the side effect of ensuring that whenever a pyfunc ensembler gets updated separately by a user, redeploying a router would ensure that the new pyfunc ensembler gets rebuilt as part of the new web ensembler service image, instead of relying on a cached version. The arbitrary slice of the original hash is done due to constraints in name length.

api/turing/models/ensembler.go
```diff
@@ -27,7 +28,7 @@ type EnsemblerType string
 const (
 	EnsemblerStandardType EnsemblerType = "standard"
 	EnsemblerDockerType   EnsemblerType = "docker"
-	EnsemblerTypePyFunc   EnsemblerType = "pyfunc"
+	EnsemblerPyFuncType   EnsemblerType = "pyfunc"
```
deadlycoconuts, line 31: Renamed the constants to create more consistency in the naming, though technically, the naming conventions originally didn't quite serve the same purpose... `EnsemblerTypePyFunc` was used only for batch ensemblers, which were a separate thing from the other two types which are only used for ensembler configs for routers.

api/turing/models/ensembler.go
```diff
@@ -79,3 +89,22 @@ func (c EnsemblerDockerConfig) Value() (driver.Value, error) {
...
+	return json.Marshal(c)
+}
+
+// Scan implements sql.Scanner interface so database tools like go-orm knows how to de-serialize the struct object
+// from the database
+func (c *EnsemblerPyFuncRefConfig) Scan(value interface{}) error {
+	return json.Unmarshal(value.([]byte), &c)
+}
+
+const (
```
krithika369, line 105: Are these constants better placed in the service layer? Per the previous suggestion to move the image building logic into `api/turing/service/router_deployment_service.go`, these consts can move there too. Don't think they need to be in the `models` package.

krithika369 (commented):
Thanks for the detailed summary, @deadlycoconuts ! I took a quick look and left some comments / questions. You may also want to get another set of eyes on this MR.

Could you also help clean up the README? There is a section showing the `config-dev.yaml` contents. But I think it can be removed, with a link to the config file, which you have already updated with the required configs for this feature.

Lastly, I imagine there will some user docs being updated, for the Pyfunc ensembler creation and router deployment (similar to batch ensemblers, both for the public docs and our internal docs; although batch docs are currently only internal - we can sort that out later). Will this be a part of the UI card? If not, please capture this in the backlog.

Thank you!

api/turing/api/deployment_controller.go
deadlycoconuts: It's an arbitrary number to pick a subset of the entire RunID because the the RunID is pretty long. 

At 32 characters, we'll end up using more than half of the max character length for pod names (63) if we were to specify the entire RunID in the pod name of the image builder, leaving out less than half of the characters available for other information that can go into the name, i.e. job type, project name, ensembler name, ensembler id, the random 5-char alphanumeric string that k8s adds to pods, and the dashes that go in between these stuff.

Right now the naming convention for image building jobs go like `"service-builder-[projectName]-[ensemblerName]-[5 chars of the RunID]-[ensemblerID]-[k8s string]"`, but I think to make things simpler and shorter, maybe I'll throw out elements such as ensembler name and shorten the `service-builder` part. If we want to cut things further I'll have to insert these data as annotations into the pod. Not sure if we're sacrificing readability if we go that direction.

api/api/openapi.bundle.yaml
```diff
@@ -2023,8 +2032,10 @@ components:
         max_replica:
           type: integer
         cpu_request:
+          pattern: ^(\d{1,3}(\.\d{1,3})?)$|^(\d{2,5}m)$
```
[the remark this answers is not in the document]
deadlycoconuts: 😅  it came from one of the previous MRs

api/config-dev.yaml
```diff
@@ -32,6 +32,25 @@ BatchEnsemblingConfig:
         Limits:
           CPU: "1"
           Memory: 1Gi
+EnsemblerServiceBuilderConfig:
+  DefaultEnvironment: dev
+  ImageBuildingConfig:
```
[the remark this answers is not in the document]
deadlycoconuts: Whoa awesome tip there! Thanks! :D

api/db-migrations/000010_add_py_func_ref_config.down.sql
```diff
@@ -0,0 +1,4 @@
+-- Removes py_func_ref_config from table.
+-- Hence, this migration involves data loss for ensemblers with type that is "pyfunc"
+ALTER TABLE ensembler_configs DROP COLUMN py_func_ref_config;
```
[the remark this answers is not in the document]
deadlycoconuts: Yeah the naming's a bit clumsy, I'll get that changed.

api/turing/api/deployment_controller.go
```diff
@@ -231,6 +235,65 @@ func (c RouterDeploymentController) deployRouterVersion(
...
+		eventsCh.Write(
+			models.NewInfoEvent(
+				models.EventStageDeployingDependencies,
+				"pyfunc ensembler with project_id: %d and ensembler_id: %d built/retrieved successfully",
+				*routerVersion.Ensembler.PyFuncRefConfig.ProjectID,
+				*routerVersion.Ensembler.PyFuncRefConfig.EnsemblerID,
+			),
+		)
+		// Create a new docker config for the ensembler with the newly generated image
+		routerVersion.Ensembler.DockerConfig = &models.EnsemblerDockerConfig{
```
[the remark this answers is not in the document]
deadlycoconuts: Yep that's right :)

api/turing/api/request/request.go
```diff
@@ -148,6 +149,29 @@ func (r CreateOrUpdateRouterRequest) BuildRouterVersion(
 		if r.Config.Ensembler.Type == models.EnsemblerStandardType && r.Config.Ensembler.StandardConfig == nil {
 			return nil, errors.New("missing ensembler standard config")
 		}
+		if r.Config.Ensembler.Type == models.EnsemblerPyFuncType {
+			if r.Config.Ensembler.PyFuncRefConfig == nil {
+				return nil, errors.New("missing ensembler pyfunc reference config")
+			}
+
+			// Verify if the ensembler given by its ProjectID and EnsemblerID exist
+			ensembler, err := ensemblersSvc.FindByID(
```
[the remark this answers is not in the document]
deadlycoconuts: Nope there doesn't seem to be an API endpoint for deleting ensemblers. Not sure if this was a conscious design choice though.

api/api/openapi.bundle.yaml
```diff
@@ -682,7 +682,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/inline_response_200'
```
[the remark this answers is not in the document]
deadlycoconuts: Okay I've just managed to clean the OpenAPI specs up from any remnants of `inline_response`s by either creating new named schemas or using existing ones.

api/turing/cluster/servicebuilder/router.go
```diff
@@ -199,7 +199,8 @@ func (sb *clusterSvcBuilder) buildRouterEnvs(
 			{Name: envEnricherTimeout, Value: ver.Enricher.Timeout},
 		}...)
 	}
-	if ver.Ensembler != nil && ver.Ensembler.Type == models.EnsemblerDockerType {
+	if ver.Ensembler != nil && ((ver.Ensembler.Type == models.EnsemblerPyFuncType && ver.Ensembler.
+		DockerConfig != nil) || ver.Ensembler.Type == models.EnsemblerDockerType) {
```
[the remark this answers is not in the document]
deadlycoconuts: Actually yes, because should the original deployment process fail due to an image building failure, there wouldn't be an associated Docker config for the pyfunc ensembler specified. Turing API will then attempt to undeploy the current router version, and part of the undeployment process involves calling `buildRouterEnvs`, which is the function which you quoted above. 

If the check for `ver.Ensembler.DockerConfig != nil` didn't happen, we'll end up accessing an invalid memory address when trying to access variables in the `DockerConfig` just shortly afterwards: https://github.com/gojek/turing/blob/deda5df4145c328e235204d8858507379a1fafc3/api/turing/cluster/servicebuilder/router.go#L208

api/turing/cluster/servicebuilder/router.go
```diff
@@ -432,7 +433,8 @@ func buildFiberConfig(
 	// If ensembler is set and is of Docker type, use "combiner" router
 	// Else, "eager" router is used.
 	var routerConfig fiberconfig.Config
-	if ensembler != nil && ensembler.Type == models.EnsemblerDockerType {
+	if ensembler != nil && ((ensembler.Type == models.EnsemblerPyFuncType && ensembler.DockerConfig != nil) ||
```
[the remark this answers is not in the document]
deadlycoconuts: Same issue as the reply above, but I think given that in this case we aren't accessing anything with a `DockerConfig`, it's probably safe to remove the `ensembler.DockerConfig != nil` part of the check.

api/turing/api/deployment_controller.go
```diff
@@ -231,6 +235,65 @@ func (c RouterDeploymentController) deployRouterVersion(
 		}
 	}
 
+	// Retrieve pyfunc ensembler if pyfunc ensembler is specified
+	if routerVersion.Ensembler != nil && routerVersion.Ensembler.Type == models.EnsemblerPyFuncType {
```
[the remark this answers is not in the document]
deadlycoconuts: Hmm yeah that sounds like a nice way to do it! It seems nicer to hide the `EnsemblerServiceBuilder` away in a `DeploymentService` struct instead of letting it sit at the `AppContext` level. The only odd part is probably where `DeployRouterVersion` would appear to be fiddling a bit with ensembler configuration instead of only managing k8s resources, but I guess that's a tiny detail we can overlook 😅

api/turing/models/ensembler.go
deadlycoconuts: Oh yea that's true, or another way I can think of is to dump them in `config_dev.yaml` together with `EnsemblerServiceBuilderConfig`. Or would that be overkill?

krithika369: Haha `EnsemblerServiceBuilderConfig` is probably an overkill for now.

api/turing/api/deployment_controller.go
krithika369: > Not sure if we're sacrificing readability if we go that direction.

Yes, that would be sacrificing the readability which I agree is not such a good idea. Let's stick to the current implementation. Could you add a comment that the `[:5]` is arbitrary?

Also:
* Should Run ID go to the end of the name, to also be consistent with the batch builder naming?
* Can we drop the `builder-` part of the name, which doesn't seem to add too much value and can help shorten the name a bit?

api/turing/api/request/request.go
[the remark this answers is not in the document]
krithika369: Ok.. I'm not sure either. Could you create a card for this (delete for batch / real-time ensemblers) in the backlog, just to track it?

api/turing/api/deployment_controller.go
deadlycoconuts: > Could you add a comment that the [:5] is arbitrary?

Yep sure.

> Can we drop the builder- part of the name, which doesn't seem to add too much value and can help shorten the name a bit?

That's true, besides, the namespace where the builder jobs get launched in is `image-builder` anyway so another `-builder-` is probably redundant.

> Should Run ID go to the end of the name, to also be consistent with the batch builder naming?

Hmm okay. I think to bypass the entire pod naming length issue altogether, I'll leave the pod naming conventions as they are without the Run ID, i.e. keeping `"service-builder-[projectName]-[ensemblerName]-[ensemblerID]-[k8s string]"` and `"job-builder-[projectName]-[ensemblerName]-[ensemblerID]-[k8s string]"`. On the other hand, I'll only inject the _entire_ Run ID into the docker image name (the one generated by the builder) instead, which will end up looking something like `"[registryName]/[projectName]-[ensemblerName]-[RunID]-service"` and `"[registryName]/[projectName]-[ensemblerName]-job"` respectively instead. Since docker image names have a way longer char limit (128), the Run ID's length isn't going to pose any problem. 

To do this I'll probably refactor the `BuildImageRequest` struct a little to carry an additional field `RunID`, which will be the `RunID` associated with the particular pyfunc ensembler.

api/turing/api/request/request.go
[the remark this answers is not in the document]
ashwinath: I can't remember why there isn't a delete ensembler API for this or maybe I wasn't kept in the loop. Ensemblers I believe was done by Roman where as I did the orchestration. There is a delete for the orchestration part.

api/turing/api/deployment_controller.go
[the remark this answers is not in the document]
deadlycoconuts: Now that I've tried refactoring the stuff a little, it appears that it might be a bit complex trying to move `EnsemblerServiceBuilder` into `DeploymentService`... since this makes the service package where `DeploymentService` is from dependent on the imagebuilder package, which itself is dependent on the service package.. creating a cyclical dependency.

api/turing/cluster/servicebuilder/router.go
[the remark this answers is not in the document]
deadlycoconuts: Hmm... that is true xD Maybe there was originally a concern that an ensembler config not having the type `docker`, but with a `DockerConfig` getting all the way to that stage? But yeah I don't think that's possible, so I'll just update those checks with that suggestion you have.

api/turing/api/deployment_controller.go
```diff
@@ -231,6 +235,65 @@ func (c RouterDeploymentController) deployRouterVersion(
...
+		ensembler, err := c.EnsemblersService.FindByID(
+			*routerVersion.Ensembler.PyfuncConfig.EnsemblerID,
+			service.EnsemblersFindByIDOptions{
+				ProjectID: routerVersion.Ensembler.PyfuncConfig.ProjectID,
+			})
+		if err != nil {
+			return "", fmt.Errorf("failed to find specified ensembler: %w", err)
+		}
+
+		pyFuncEnsembler := ensembler.(*models.PyFuncEnsembler)
```
ashwinath, line 249: I think it's good practice to use the 2nd return value when casting.
```go
pyFuncEnsembler, ok := ensembler.(*models.PyFuncEnsembler)
if !ok {
    return "", fmt.Errorf("failed to cast ensembler: %w", err)
}
```

Your code will panic if it's wrongly casted for some reason.

deadlycoconuts: Ahh okay, thanks for pointing that out; I'll go fix that.

api/turing/api/deployment_controller.go
ashwinath: While in this case slicing a string may seem ok since we know that RunID falls within the ASCII range, if not it's better to use a Rune slice instead (https://go.dev/blog/strings).

api/turing/api/ensemblers_api_it_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
ashwinath, line 1: Why is this changed?

api/turing/api/ensembling_job_api_it_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
ashwinath, line 1: Same for this, why is this changed?

api/turing/api/ensemblers_api_it_test.go
deadlycoconuts: Oh yeah tbh they probably got left there by my IDE or the linter, I should be able to remove them right away.

api/turing/api/ensembling_job_api_it_test.go
deadlycoconuts: ^Same as above, some remnants left by the IDE probably.

api/turing/cluster/servicebuilder/router.go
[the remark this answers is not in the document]
ashwinath: I'd even consider writing a function to check if it's either a EnsemblerPyFuncType or a EnsemblerDockerType so that it doesn't look that gory.

api/turing/database/integration_config.go
```diff
@@ -1,3 +1,4 @@
+//go:build integration
```
ashwinath, line 1: Same for this, why the change?

api/turing/service/router_deployment_service.go
```diff
@@ -351,7 +351,9 @@ func (ds *deploymentService) createServices(
 	}
 
 	// Ensembler
-	if routerVersion.Ensembler != nil && routerVersion.Ensembler.Type == models.EnsemblerDockerType {
+	if routerVersion.Ensembler != nil &&
+		((routerVersion.Ensembler.Type == models.EnsemblerPyFuncType && routerVersion.Ensembler.DockerConfig != nil) ||
```
ashwinath, line 355: I see this check being used quite often and maybe there should be a utility function that checks if its either a EnsemblerPyFuncType or EnsemblerDockerType

api/turing/api/ensemblers_api_it_test.go
deadlycoconuts: Okay I've just checked, it was my IDE running gofmt automatically when I saved these files.

engines/pyfunc-ensembler-service/app.Dockerfile
```diff
@@ -8,7 +8,8 @@ ARG FOLDER_NAME
 RUN gsutil -m cp -r ${MODEL_URL} .
 
 # Install dependencies required by the user-defined ensembler
-RUN /bin/bash -c ". activate ${CONDA_ENV_NAME} && conda env update --name ${CONDA_ENV_NAME} --file /${HOME}/${FOLDER_NAME}/conda.yaml"
+RUN /bin/bash -c ". activate ${CONDA_ENV_NAME}"
+RUN conda env update --name ${CONDA_ENV_NAME} --file ./${FOLDER_NAME}/conda.yaml
```
ashwinath, line 12: Does this actually work? The reason why it was done inline is because the states do not share when running in a new docker layer. i.e. running `/bin/bash -c ". activate ${CONDA_ENV_NAME}"` then closing the shell and then running `conda env update --name ${CONDA_ENV_NAME} --file ./${FOLDER_NAME}/conda.yaml`.

If this works isn't line 11 redundant?

api/turing/database/integration_config.go
deadlycoconuts: I've just removed this.

engines/pyfunc-ensembler-service/app.Dockerfile
deadlycoconuts: Hmm it actually works, so you're right I think - line 11 is indeed probably redundant. I'll give it a try and see how it goes.

api/turing/service/router_deployment_service.go
deadlycoconuts: That works too, but I think given the discussion above where we're changing the logic of the check to verify if an ensembler has a docker config or not, I'm not sure if it's excessive to replace `if routerVersion.Ensembler != nil && routerVersion.Ensembler.DockerConfig != nil` with `if hasDockerConfig(routerVersion.Ensember)` :/

api/turing/api/deployment_controller.go
deadlycoconuts: Hmm okay I think I might've created this tiny confusion over pod/image names in this thread. For pod names, there should no longer be a concern over name lengths because I'm planning to remove RunID from them altogether. 

Furthermore, the unique names concern only exists for image names because that's how Turing API checks for images that have already been created to avoid recreating an image with the same ensembler. To address this, I'll be adding the entire RunID to the image name so that serves to distinguish between images built with ensemblers with the same ensembler name, ensemblerID, projectID but with different versions (because a user can update an ensembler with a different ensembling logic). 

The confusion/problem arose when I was initially trying to splice the RunID with the ensembler name as a *new* ensembler name used by the existing name generators, hence confounding both naming concerns together. This was initially done to avoid having to refactor the existing `imageBuilder` struct since the original batch ensembling images don't have RunIDs in their image names to distinguish themselves (i.e. if I understood the existing implementation correctly, a user rebuilding a batch ensembling job with an updated ensembler would probably not have a new image rebuilt because an image containing the very same project and ensembler name already exist in the registry, leading to the cached image being used for the ensembling job, which would be incorrect). 

Okay apologies for the really long comment 😅  but I hope this clears things up!

api/turing/api/deployment_controller.go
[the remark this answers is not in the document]
krithika369: Haha. Seems the only usage of the `service` package in the `imagebuilder` is a `service.EnsemblerFolder` from the `createKanikoJob` method, which IMO should not be directly referenced here in the first place (a generic image builder's method shouldn't need to know about the ensembler service). So, I think we can safely break this dependency and pass the `service.EnsemblerFolder` constant in as an argument to the imagebuilder's BuildImage().

api/turing/service/router_deployment_service.go
ashwinath: Definitely not excessive, I actually prefer the latter because shows the intent of the person writing it. `hasDockerConfig()` has more meaning than `A && B`.

api/turing/cluster/servicebuilder/router.go
```diff
@@ -199,7 +199,7 @@ func (sb *clusterSvcBuilder) buildRouterEnvs(
 			{Name: envEnricherTimeout, Value: ver.Enricher.Timeout},
 		}...)
 	}
-	if ver.Ensembler != nil && ver.Ensembler.Type == models.EnsemblerDockerType {
+	if ver.HasDockerConfig() {
```
deadlycoconuts, line 202: Simplified this to make the intention of the check explicit

api/turing/service/router_deployment_service.go
```diff
@@ -96,11 +104,20 @@ func (ds *deploymentService) DeployRouterVersion(
...
 	ensemblerServiceAccountKey string,
+	pyfuncEnsembler *models.PyFuncEnsembler,
 	experimentConfig json.RawMessage,
 	eventsCh *EventChannel,
 ) (string, error) {
 	var endpoint string
 
+	// If pyfunc ensembler is specified as an ensembler service, build/retrieve its image
+	if pyfuncEnsembler != nil {
+		err := ds.buildEnsemblerServiceImage(pyfuncEnsembler, project, routerVersion, eventsCh)
```
deadlycoconuts, line 115: `DeploymentService` now does the weightlighting for image building.

engines/pyfunc-ensembler-service/app.Dockerfile
deadlycoconuts: Okay it works as usual. I've pushed a new commit without that line.

api/api/openapi.bundle.yaml
```diff
@@ -551,7 +551,7 @@ paths:
...
+                $ref: '#/components/schemas/IdObject_1'
```
ashwinath, line 554: This is automatically generated by the Makefile right?

ashwinath (approved):

api/api/openapi.bundle.yaml
deadlycoconuts: Yep that's right. I corrected the source openapi specs that was generated with `make bundle-openapi`

deadlycoconuts:
Thanks a lot for all the comments everyone! I'll rebase this on @leonlnj 's latest changes and it'll be good to go!

PR #171, opened 2022-02-28
Add UI/SDK support to pyfunc ensemblers

deadlycoconuts:
### Context
In PR #164 and PR #165, a real-time ensembler engine was introduced to process live ensembling requests, using an ensembler that is predefined in the `mlflow` pyfunc flavour. In PR #170, a process was introduced in Turing API to automatically orchestrate the image building process of a user specified pyfunc ensembler (identified by its given `project_id` and `ensembler_id`). Turing API then runs containers based on that generated image as an ensembler service as part of a Turing Router deployment. 

This PR introduces user interface changes in order to incorporate the minor tweaks associated with a Turing Router deployment with the inclusion of pyfunc ensemblers. 

Minor additions/fixes were also made to Turing SDK to incorporate the `pyfunc_config` object created from PRs #164, #165 and #170. 

Not to be merged before PR #170.

### Features
- Creation of a panel in the router creation/editing page to select a pyfunc ensembler to deploy from a list of created ensemblers stored in Turing:
Screenshot 2022-03-02 at 5 16 38 PM

- Creation of a panel on the router details page showing details of the pyfunc ensembler deployed, with the Ensembler ID and Ensembler Name shown:
Screenshot 2022-03-02 at 5 16 12 PM

- Changing pyfunc ensemblers in an existing Turing Router will be reflected as expected for all Turing Router configuration on the diff page:
Screenshot 2022-03-02 at 5 17 27 PM

### Main Modifications
- `ui/src/components/form/service_account_combo_box/ServiceAccountComboBox.js` - Renaming of `ServiceAccoutComboBox.js` to the more correctly spelled `ServiceAccountComboBox.js`
- `ui/src/router/components/configuration/components/EnsemblerConfigSection.js` - Addition of a panel to display pyfunc ensembler configuration
- `ui/src/router/components/configuration/components/pyfunc_config_section/*` - Addition of a panel to display pyfunc ensembler configuration
- `ui/src/router/components/form/components/ensembler_config/typeOptions.js` - Addition of pyfunc ensemblers to the list of available ensembler options
- `ui/src/router/components/form/components/pyfunc_config/*` - Addition of a new form to fill for pyfunc ensemblers
- `ui/src/router/components/form/validation/schema.js` - Update to schema validation for pyfunc ensemblers
- `ui/src/services/ensembler/PyFuncEnsembler.js` - Addition of new pyfunc ensembler service

### Minor Changes to Turing SDK
- `sdk/turing/generated/*` - Changes to OpenAPI auto-generated SDK classes after changes to OpenAPI specs in PR #170
- `sdk/turing/router/config/router_ensembler_config.py` - Addition of a new SDK class to support `pyfunc_config` 
- `sdk/turing/router/router.py` - Minor changes to functions after changes to OpenAPI specs in PR #170
- `sdk/turing/session.py` - Minor changes to functions after changes to OpenAPI specs in PR #170


sdk/tests/router/router_test.py
```diff
@@ -99,7 +99,7 @@ def test_create_router(turing_api, active_project, actual, expected, use_google_
     "actual,expected", [
         pytest.param(
             1,
-            turing.generated.models.InlineResponse200(id=1)
+            turing.generated.models.IdObject(id=1)
```
deadlycoconuts, line 102: These changes reflect the removal of the odd `InlineResponse` classes from the SDK code once the OpenAPI specs have been cleaned up in PR #170.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -85,24 +88,105 @@ def docker_config(self, docker_config: turing.generated.models.EnsemblerDockerCo
...
+            kwargs["pyfunc_config"] = self.pyfunc_config
 
         return turing.generated.models.RouterEnsemblerConfig(
             type=self.type,
             **kwargs
         )
 
 
+@dataclass
+class PyfuncRouterEnsemblerConfig(RouterEnsemblerConfig):
```
deadlycoconuts, line 125: Here's a new class to represent the new pyfunc ensemblers to be created.

sdk/tests/router/router_test.py
leonlnj: im bit confused, was this `idobject` named as such for any reasons?

image

leonlnj: im bit confused, was this `idobject` named as such for any reasons?

image

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -85,24 +88,105 @@ def docker_config(self, docker_config: turing.generated.models.EnsemblerDockerCo
         else:
             self._docker_config = docker_config
 
+    @property
+    def pyfunc_config(self) -> turing.generated.models.EnsemblerPyfuncConfig:
+        return self._pyfunc_config
+
+    @pyfunc_config.setter
+    def pyfunc_config(self, pyfunc_config: turing.generated.models.EnsemblerPyfuncConfig):
+        if isinstance(pyfunc_config, turing.generated.models.EnsemblerPyfuncConfig):
```
leonlnj, line 97: why not 

```python
if isinstance(pyfunc_config, dict):
            pyfunc_config['resource_request'] = \
                turing.generated.models.ResourceRequest(**pyfunc_config['resource_request'])
            self._pyfunc_config = turing.generated.models.EnsemblerPyfuncConfig(
                **pyfunc_config
            )
else
     self._pyfunc_config = pyfunc_config
```

sdk/tests/router/router_test.py
deadlycoconuts: `IdObject` is made to align with the return values of certain Turing API endpoints that return an object containing a field named `id`. 

`Id` on the other hand is made to align with the return values of *other* Turing API endpoints that return an integer representing some id.

If you're asking why Turing API was set up to return such integer values in a seemingly inconsistent way, we can perhaps bring this up again in some issue so that we can realign all the integer-related return values.

ui/src/router/components/form/components/pyfunc_config/PyFuncDeploymentPanel.js
```diff
@@ -0,0 +1,83 @@
...
+              label={
+                <FormLabelWithToolTip
+                  label="Pyfunc Ensembler *"
+                  content="Select the Pyfunc Ensembler to be used in your deployment"
+                />
+              }
+              // isInvalid={!!errors.pyfunc_config}
+              // error={errors}
+              display="row">
+              <EuiSuperSelect
```
leonlnj, line 52: Can we make it a combobox similarly to the[ docker ensembler](https://github.com/gojek/turing/blob/main/ui/src/router/components/form/components/docker_config/DockerDeploymentPanel.js#L47)? This component doesn't allow text search of ensembler name, user would need to scroll all the selections.

ui/src/router/components/form/components/pyfunc_config/PyFuncDeploymentPanel.js
```diff
@@ -0,0 +1,83 @@
...
+            <EuiFormRow
+              fullWidth
+              label={
+                <FormLabelWithToolTip
+                  label="Pyfunc Ensembler *"
+                  content="Select the Pyfunc Ensembler to be used in your deployment"
+                />
+              }
+              // isInvalid={!!errors.pyfunc_config}
+              // error={errors}
```
leonlnj, line 50: there should be error handling where Ensembler Type is set to pyfunc, but pyfunc ensembler is not selected. User shouldn't be able to navigate further -as per enricher/ensembler docker image type selected but docker image not populated

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: That's actually a really good point. I realised a lot of the setters in the SDK have been written in the same way so I'll go simplify those too.

ui/src/router/components/form/components/pyfunc_config/PyFuncDeploymentPanel.js
deadlycoconuts: That's a good suggestion, I've just pushed this under the new changes :)

ui/src/router/components/form/components/pyfunc_config/PyFuncDeploymentPanel.js
deadlycoconuts: Oops those were left out unintentionally, I've got those error handling and schema validation covered now with the new changes.

ui/src/router/components/configuration/components/pyfunc_config_section/PyFuncConfigViewGroup.js
```diff
@@ -0,0 +1,46 @@
...
+}) => {
+  const items = [
+    {
+      title: "Pyfunc Ensembler Details",
+      children: <PyFuncRefConfigTable config={pyfuncConfig} />,
+    },
+  ];
+
+  if (!!dockerConfig) {
+    items.push({
```
deadlycoconuts, line 22: This ensures that the router details screen is minimally able to display a Pyfunc ensembler's name and ID (but not the `dockerConfig` related stuff) should the image building job fail during router deployment.

leonlnj:
thanks, lgtm!

leonlnj (approved):

sdk/turing/generated/model/id_object1.py
```diff
@@ -0,0 +1,179 @@
...
+    convert_js_args_to_python_args,
+    date,
+    datetime,
+    file_type,
+    none_type,
+    validate_get_composed_info,
+)
+
+
+class IdObject1(ModelSimple):
```
krithika369, line 30: Can we get rid of the `IdObject_1` from the openAPI specs and give it a name too ?

krithika369, line 30: Can we get rid of the repeated `IdObject` definitions in the jobs.yaml / router.yaml files and directly reference the one in common.yaml? So we wont have multiple such classes with the numeric suffixes.

krithika369, line 30: Can we get rid of the repeated `IdObject` definitions in the jobs.yaml / router.yaml files and directly reference the one in common.yaml? So we wont have multiple such classes with the numeric suffixes.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -47,7 +50,7 @@ def type(self) -> str:
 
     @type.setter
     def type(self, type: str):
-        assert type in {'standard', 'docker'}
+        assert type in {'standard', 'docker', "pyfunc"}
```
krithika369, line 53: Super nit: We should be consistent the use of `''` vs `""` in python. I generally prefer the [black](https://github.com/psf/black) code formatter which picks double quotes over single but that's debatable.

krithika369, line 53: Super nit: We should be consistent the use of `''` vs `""` in python. I generally prefer the [black](https://github.com/psf/black) code formatter which picks double quotes over single but that's debatable.

sdk/turing/generated/model/id_object1.py
deadlycoconuts: I've actually tried doing that but it seems like it triggers a bug in the OpenAPI specs bundling, whereby it creates a reference for the `IdObject` in `openapi.bundle.yaml` to `'#/components/schemas/~1projects~1{project_id}~1routers~1{router_id}~1undeploy'`. This then causes the Python OpenAPI autogenerated classes to import from a module titled `projects_project_id_routers_router_id`, which doesn't exist. I'm not sure if this sounds similar to anything you're encountered before though.

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: Ahhah, we could bring `black` in as part of our linting process. But yes I've fixed the `''` and the `""` issue.

sdk/turing/generated/model/id_object1.py
deadlycoconuts: Oops okay I figured that that might not have been the right way to reference it; lemme find out if there's a neater way to do this

krithika369 (approved):
LGTM. Thanks for the comprehensive PR!

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -32,6 +32,17 @@ const typeOptions = [
       </Fragment>
     ),
   },
+  {
+    value: "pyfunc",
+    inputDisplay: "Pyfunc",
+    description: (
+      <Fragment>
+        Turing will deploy the selected pyfunc ensembler and will send to it
```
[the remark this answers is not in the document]
deadlycoconuts: That sounds helpful; I'll add that in :)

sdk/turing/generated/model/id_object1.py
deadlycoconuts: Okay I've taken a closer look at this, apparently according the [OpenAPI 3.0 standard](https://swagger.io/docs/specification/components/) (see the section on 'Externally Defined Components'), all external file paths are only used in references made in locally defined components. This is so that the latter act as 'local aliases' for the entire file (defining request bodies, responses, etc.). 

It seems like the OpenAPI generator also sticks to that convention and fails to generate the proper references if we try to make external references directly in other parts of the file (in this case it's the response definition for deleting routers) without first creating a local alias (a local component with the external reference). 

On the other hand if we were to try defining the `IdObject` in `router.yaml` (or `jobs.yaml`) with an external reference to the _other_ component instead, such that the definitions are linked up in a chain, i.e.:

In `router.yaml`:
```yaml
...
components:
  schemas:
    IdObject:
      $ref: "jobs.yaml#/components/schemas/IdObject"
...
```

In `jobs.yaml`:
```yaml
...
components:
  schemas:
    IdObject:
      $ref: "common.yaml#/components/schemas/IdObject"
...
```

This doesn't work either and still ends up generating 2 different `IdObject` classes. A similar issue with OpenAPI generator has been reported [here](https://github.com/OpenAPITools/openapi-generator/issues/2701), and it seems like it's still open from 2019 till today... :/ 

TL;DR: So all in all, I think we can't just erase the 2 `IdObject` definitions away from `router.yaml` or `jobs.yaml` :(

krithika369: I see.. Thanks for investigating this, @deadlycoconuts!

What if we use `RouterId` and `JobId` in the respective `routers.yaml` and `jobs.yaml` and files and no longer name the local component `IdObject` (only the item in `common.yaml` will be called `IdObject`) ?

deadlycoconuts: That works, but we'll have 3 classes instead of 2, but I guess that's way more elegant than having `IdObject` and `IdObject1` 😅 Thanks for the awesome solution; I'll put that in and merge the changes once I test it out again with the SDK.

krithika369: Cool! Thanks :)

PR #172, opened 2022-03-02
Set kafka message default to 10mb

leonlnj:
# Overview
This MR seek to increase the default producer message for kafka. Currently if the kafka broker is configured to accept larger size than default (1mb), it is limited on router's end as the size is not set. 

Although the producer config required on [kafka doc](https://kafka.apache.org/documentation.html#producerconfigs_max.request.size) is `max.request.size`, as we are using `confluent-kafka-go.v1/kafka` library which dependant on `librdkafka`.

In the [librdkafka doc](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md), the config field is `message.max.bytes`, hence this is used instead.

This is not the most elegant approach, ideally a set of kafka config can be passed as into to the router and be configured dynamically, which will be considered in the future.

# Modification
`engines/router/missionctl/log/resultlog/kafka.go` - added `"message.max.bytes": 10485760` which set the defaults as 10mb. 

# Testing
Tested with 1.5mb payload which throws `Result Logging Error: Broker: Message size too large`. After changing the config and running kafka with the changes in `engines/router/compose/kafka.yaml`, the message is produced successfully.


engines/router/missionctl/log/resultlog/kafka.go
```diff
@@ -53,7 +53,7 @@ func newKafkaLogger(cfg *config.KafkaConfig) (*KafkaLogger, error) {
 }
 
 func newKafkaProducer(cfg *config.KafkaConfig) (kafkaProducer, error) {
-	producer, err := kafka.NewProducer(&kafka.ConfigMap{"bootstrap.servers": cfg.Brokers})
+	producer, err := kafka.NewProducer(&kafka.ConfigMap{"bootstrap.servers": cfg.Brokers, "message.max.bytes": 10485760})
```
terryyylim, line 56: Considering this is a default value, should we make this configurable via `KafkaConfig`(i.e https://github.com/gojek/turing/blob/main/engines/router/missionctl/config/config.go#L94)?

terryyylim (approved):

engines/router/compose/kafka.yaml
```diff
@@ -23,6 +23,9 @@ services:
       - KAFKA_CFG_ADVERTISED_LISTENERS=CLIENT://kafka:9092,EXTERNAL://localhost:9093
       - KAFKA_CFG_AUTO_CREATE_TOPICS_ENABLE=true
       - KAFKA_INTER_BROKER_LISTENER_NAME=CLIENT
+      - KAFKA_CFG_MAX_REQUEST_SIZE=10485760
```
krithika369, line 26: Are all 3 values here required? Based on the changes to `engines/router/missionctl/log/resultlog/kafka.go`, I imagine we only need the second one. Can we clean up the extra ones?

PR #173, opened 2022-03-03
Set kafka message size and compression as configurable

leonlnj:
# Overview
This MR is a follow up from the [previous](https://github.com/gojek/turing/pull/172), it allows
`max.request.size` and `compression.type` of the Kafka producer configuration to be set in helm values when deploying Turing Server.

# Modification
`api/turing/api/request/request.go` - Appended `MaxMessageBytes` and `CompressionType` during building of new router version when logger type is kafka.
`api/turing/cluster/servicebuilder/router.go` - Creating router with env var `APP_KAFKA_MAX_MESSAGE_BYTES` and `APP_KAFKA_COMPRESSION_TYPE`
`engines/router/missionctl/log/resultlog/kafka.go` - Router's logger to init with config, which are read from env var during initialisation.

# Testing
Setting router default in Turing Server helms value
image

Router container env var set with the global values
image

### DIFF
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -275,6 +275,8 @@ type RouterDefaults struct {
 	// 	blue-exp-engine:
 	//	  Image: ghcr.io/myproject/blue-exp-engine-plugin:v0.0.1
 	ExperimentEnginePlugins map[string]*ExperimentEnginePluginConfig `validate:"dive"`
+	// Kafka Configuration. If result logging is using Kafka
+	KafkaConfig *KafkaConfig
 }
 
 // FluentdConfig captures the defaults used by the Turing Router when Fluentd is enabled
@@ -285,6 +287,14 @@ type FluentdConfig struct {
 	Tag string
 	// Flush interval seconds - value determined by load job frequency to BQ
 	FlushIntervalSeconds int
+}
+
+// KafkaConfig captures the defaults used by Turing Router when result logger is set to kafka
+type KafkaConfig struct {
+	// Producer Config - Max message byte to send to broker
+	MaxMessageBytes int
+	// Producer Config - Compression Type of message
+	CompressionType string
 }
 
 // AuthorizationConfig captures the config for auth using mlp authz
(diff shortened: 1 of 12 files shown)


api/turing/config/config.go
```diff
@@ -469,6 +479,8 @@ func setDefaultValues(v *viper.Viper) {
 	v.SetDefault("RouterDefaults::FluentdConfig::Tag", "turing-result.log")
 	v.SetDefault("RouterDefaults::FluentdConfig::FlushIntervalSeconds", "90")
 	v.SetDefault("RouterDefaults::Experiment", map[string]interface{}{})
+	v.SetDefault("RouterDefaults::KafkaConfig::MaxMessageBytes", "1048588")
+	v.SetDefault("RouterDefaults::KafkaConfig::CompressionType", "none")
```
terryyylim, line 483: Should this default value be `uncompressed` rather than `none`?

Ref: https://kafka.apache.org/documentation/#topicconfigs_compression.type

api/turing/models/log_config.go
```diff
@@ -52,6 +52,10 @@ type KafkaConfig struct {
 	Topic string `json:"topic"`
 	// Serialization Format used for the messages
 	SerializationFormat SerializationFormat `json:"serialization_format"`
+	// Producer Config - Max message byte to send to broker
+	MaxMessageBytes int `json:"max_message_bytes"`
+	// Producer Config - Compression Type of message
```
terryyylim, line 57: nit: `Compression type for the given Topic`

terryyylim (commented):
Thanks for the changes @leonlnj !

api/turing/config/config.go
leonlnj: Nope it is `none` because we are setting this for the producer config
https://kafka.apache.org/documentation/#producerconfigs_compression.type

terryyylim: Ah saw the wrong doc, cool thanks!

terryyylim (approved):

api/turing/models/log_config.go
```diff
@@ -52,6 +52,10 @@ type KafkaConfig struct {
 	Topic string `json:"topic"`
 	// Serialization Format used for the messages
 	SerializationFormat SerializationFormat `json:"serialization_format"`
+	// Producer Config - Max message byte to send to broker
+	MaxMessageBytes int `json:"max_message_bytes"`
```
krithika369, line 56: IMO, these values can just be applied at deploy time, like many other default values we have in `RouterDefaults` (since they are currently applicable to all routers and not user-configured). Any reason we are saving this to the DB?

PR #176, opened 2022-03-08
Add steps to merge original request and enricher response headers

deadlycoconuts:
This PR introduces a minor change to merge the headers of the original request sent to Turing routers with that of the response returned from an enricher.

This in effect parses the enricher response headers for downstream traffic rules and experiment input, but *does not* parse them for ensembling.

# Modifications
- `engines/router/missionctl/handlers/http_handler.go` - Addition of a few lines of code to merge the headers of the original request and enricher response

### DIFF
--- a/engines/router/missionctl/handlers/http_handler.go
+++ b/engines/router/missionctl/handlers/http_handler.go
@@ -91,6 +91,9 @@ func (h *httpHandler) getPrediction(
 
 	// Enrich
 	var resp mchttp.Response
+	// Creates a new map to represent the merged headers from the original request headers + enricher response headers
+	postEnrichmentResponseHeader := req.Header.Clone()
+
 	payload := requestBody
 	if h.IsEnricherEnabled() {
 		resp, httpErr := h.Enrich(ctx, req.Header, payload)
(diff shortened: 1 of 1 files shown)


engines/router/missionctl/handlers/http_handler.go
```diff
@@ -100,13 +103,17 @@ func (h *httpHandler) getPrediction(
 		if httpErr != nil {
 			return nil, httpErr
 		}
+		// Merge Enricher response headers with original request headers
+		for key, value := range resp.Header() {
+			postEnrichmentResponseHeader.Set(key, value[0])
```
krithika369, line 108: Instead, this can be:
```golang
for key := range resp.Header() {
    postEnrichmentResponseHeader.Set(key, resp.Header().Get(key))
}
```

`resp.Header().Get(key)` effectively does `value[0]`. It also handles 0 length value.

krithika369 (approved):
The change looks good. I think it would be useful to also log the enricher's response header in the outcome logs. This will require a change to the proto. We can tackle this as a separate item.

engines/router/missionctl/handlers/http_handler.go
deadlycoconuts: Alright I'll make that change :)

PR #177, opened 2022-03-10
Standardize labels for kube objects

terryyylim:
### Context

Some of the Kubernetes objects created via Turing API workflows do not contain labels. This PR standardizes the labels for all objects that are required, which would help in cost management and searching of resources in a cluster.

The standardization of labels touches 2 places:
- Turing's helm chart
- Turing's API workflows which creates Kubernetes objects

### DIFF
--- a/api/turing/cluster/controller.go
+++ b/api/turing/cluster/controller.go
@@ -82,9 +82,9 @@ type Controller interface {
 	CreateJob(namespace string, job Job) (*apibatchv1.Job, error)
 	GetJob(namespace string, jobName string) (*apibatchv1.Job, error)
 	DeleteJob(namespace string, jobName string) error
-	CreateServiceAccount(namespace, serviceAccountName string) (*apicorev1.ServiceAccount, error)
-	CreateRole(namespace string, roleName string, policyRules []apirbacv1.PolicyRule) (*apirbacv1.Role, error)
-	CreateRoleBinding(namespace, roleBindingName, serviceAccountName, roleName string) (*apirbacv1.RoleBinding, error)
+	CreateServiceAccount(namespace string, serviceAccount *ServiceAccount) (*apicorev1.ServiceAccount, error)
+	CreateRole(namespace string, role *Role) (*apirbacv1.Role, error)
+	CreateRoleBinding(namespace string, roleBinding *RoleBinding) (*apirbacv1.RoleBinding, error)
 	CreateSparkApplication(namespace string, request *CreateSparkRequest) (*apisparkv1beta2.SparkApplication, error)
 	GetSparkApplication(namespace, appName string) (*apisparkv1beta2.SparkApplication, error)
 	DeleteSparkApplication(namespace, appName string) error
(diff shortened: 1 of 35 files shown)


deadlycoconuts:
Thanks for this! Looks good to me! 🚀

deadlycoconuts (approved):

infra/charts/turing/templates/cluster-role-binding.yaml
```diff
@@ -3,6 +3,14 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ template "turing.serviceAccount.name" . }}-role-binding
+  labels:
+    app: {{ template "turing.fullname" . }}
+    chart: {{ template "turing.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
```
romanwozniak, line 10: It could make sense to standardize this under `_helpers.tpl` and then include it in the individual k8s resources.
Something like: 
```
labels:
  {{ toYaml (template "turing.labels" .) | indent 4 -}}
```

api/turing/cluster/controller.go
```diff
@@ -218,6 +218,7 @@ func (c *controller) ApplyConfigMap(namespace string, configMap *ConfigMap) erro
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      configMap.Name,
 			Namespace: namespace,
+			Labels:    configMap.Labels,
```
krithika369, line 221: I think there are more instances in this file (and perhaps across the codebase) where `ObjectMeta` from `k8s.io/apimachinery/pkg/apis/meta/v1` is used, where we are currently not setting the labels. Might be easy for search for this type and figure out what else needs to be updated, since it seems like all resources rely on the same underlying struct to store the basic info.

infra/charts/turing/templates/cluster-role-binding.yaml
```diff
@@ -3,6 +3,11 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: {{ template "turing.serviceAccount.name" . }}-role-binding
+  labels:
+{{ include "turing.labels" . | indent 4 }}
+{{- if .Values.turing.labels }}
```
krithika369, line 8: This if block can also be moved into the `_helpers.tpl` file, under the definition for "turing.labels", right?

krithika369 (commented):
PR description is outdated - doesn't reflect the changes to the API code.

api/turing/cluster/controller.go
terryyylim: Thanks for catching this, took a closer look and noticed I missed out `Secrets`, as well as `ServiceAccount`, `Role`, `RoleBinding` which are currently used by Batch Ensembling. I refactored the latter resources to follow the builder pattern instead of adding a new `labels` parameter to their respective create function calls.

api/turing/cluster/servicebuilder/fluentd.go
```diff
@@ -68,7 +70,7 @@ func (sb *clusterSvcBuilder) NewFluentdService(
 			LivenessHTTPGetPath:   fluentdHealthCheckPath,
 			ReadinessHTTPGetPath:  fluentdHealthCheckPath,
 			ProbeInitDelaySeconds: 10,
-			Labels:                buildLabels(project, envType, routerVersion.Router),
+			Labels:                buildLabels(project, routerVersion.Router),
```
krithika369, line 73: If `envType` is no longer required to be set from this place (which makes sense to me), should we also remove it from the `NewFluentdService` function signature now?

api/turing/cluster/servicebuilder/fluentd_test.go
```diff
@@ -75,6 +75,13 @@ func TestNewFluentdService(t *testing.T) {
 				Namespace:   project.Name,
 				AccessModes: []string{"ReadWriteOnce"},
 				Size:        volSize,
+				Labels: map[string]string{
+					"app":          "test-svc",
+					"environment":  "",
```
krithika369, line 80: Could you leave a comment here on why the environment label value is empty ? Will be useful in the future.

api/turing/cluster/servicebuilder/plugins_server.go
```diff
@@ -41,7 +41,7 @@ func (sb *clusterSvcBuilder) NewPluginsServerService(
 			Name:                  GetComponentName(routerVersion, ComponentTypes.PluginsServer),
 			Namespace:             project.Name,
 			Image:                 nginxImage,
-			Labels:                buildLabels(project, envType, routerVersion.Router),
+			Labels:                buildLabels(project, routerVersion.Router),
```
krithika369, line 44: Think we can remove `envType` from this method's signature as well.

api/turing/cluster/servicebuilder/router.go
```diff
@@ -98,7 +98,7 @@ func (sb *clusterSvcBuilder) NewRouterService(
 	// Namespace is the name of the project
 	namespace := GetNamespace(project)
 
-	configMap, err := buildFiberConfigMap(routerVersion, project, experimentConfig)
+	configMap, err := buildFiberConfigMap(routerVersion, project, experimentConfig, envType)
```
krithika369, line 101: Why is `envType` being introduced here? Seems like it's not required downstream anyway, since all resources will use the `labeller` package's methods.

api/turing/service/router_deployment_service.go
```diff
@@ -254,7 +255,7 @@ func (ds *deploymentService) UndeployRouterVersion(
 
 	// Delete secret
 	eventsCh.Write(models.NewInfoEvent(models.EventStageDeletingDependencies, "deleting secrets"))
-	secret := ds.svcBuilder.NewSecret(routerVersion, project, "", "", "")
+	secret := ds.svcBuilder.NewSecret(routerVersion, project, ds.environmentType, "", "", "")
```
krithika369, line 258: Similar comment as above: why are we sending in the envType here ? Looks like the methods in `api/turing/cluster/servicebuilder/service_builder.go` no longer need it.

krithika369 (approved):
LGTM, left some small questions / comments. Thanks for the refactoring!

PR #178, opened 2022-03-10
Remove system kafka config from db and api

leonlnj:
# Overview
This MR seeks to patch the kafka configuration that was created [previously](https://github.com/leonlnj/turing/commit/a3d788d63059d68a9cab727bbb3110824b44bde2). 

# Modification

- Removing all `MaxMessageBytes` and `CompressionType` from `/api/*` as its not required to persist in db nor return in api response. From the api server perspective it only needs to create the env var in  `api/turing/cluster/servicebuilder/router.go`-`buildRouterEnvs` which will be picked up by the router.
- Passing`routerDefault` instead of individual default as arguments
- api/turing/service/alert_service_test.go - updated error message for clarity


api/turing/cluster/servicebuilder/router.go
```diff
@@ -85,8 +86,7 @@ func (sb *clusterSvcBuilder) NewRouterService(
 	envType string,
 	secretName string,
 	experimentConfig json.RawMessage,
-	fluentdTag string,
-	jaegerCollectorEndpoint string,
+	routerDefault *config.RouterDefaults,
```
krithika369, line 89: Nit: `routerDefaults` as plural?

api/turing/service/router_deployment_service.go
```diff
@@ -51,10 +51,9 @@ type deploymentService struct {
 	environmentType           string
 
 	// Router configs
-	fluentdConfig           *config.FluentdConfig
-	jaegerCollectorEndpoint string
-	sentryEnabled           bool
-	sentryDSN               string
+	sentryEnabled bool
+	sentryDSN     string
+	routerDefault *config.RouterDefaults
```
krithika369, line 56: Same comment as earlier - `routerDefaults` would be more consistent with the name of the struct.

krithika369 (approved):
LGTM.

PR #179, opened 2022-03-15
Add create router version feature

deadlycoconuts:
### Context
When updating a router with a new configuration, Turing API is currently only able to create a new router version and then deploy it automatically as a single atomic step. This may not be ideal for users who simply intend to create new versions without deploying them, since there isn't an endpoint that support such an operation.

This PR thus introduces a new API endpoint method for Turing API to create a new router version without deploying it automatically, and includes updates to Turing UI and Turing SDK to support this new functionality.

### Features
- New endpoint to support the creation of new router versions without deploying them automatically (`POST` operation with the existing `/projects/{project_id}/routers/{router_id}/versions` endpoint)
- Introduction of new buttons at the 'Edit Router` diff and confirmation page to reflect the new option to create new router versions without deploying them
Screenshot 2022-03-18 at 4 59 25 PM
Screenshot 2022-03-18 at 5 01 36 PM

- Introduction of a new SDK method to create new router versions without deploying them
```python
new_router_config = RouterConfig(
    routes=routes,
    rules=rules,
    default_route_id="test",
    experiment_engine=experiment_config,
    resource_request=resource_request,
    timeout="100ms",
    log_config=log_config,
    enricher=None,
    ensembler=None
)

turing.router.config.router_version.RouterVersion.create(new_router_config, my_router_id)
```
### Main Modifications
- `api/turing/api/router_versions_api.go` - Addition of a new request handler for the new API endpoint method
- `api/api/*` - Changes to the OpenAPI specs from the addition of the new API endpoint method
- `ui/src/router/edit/components/VersionComparisonView.js` - Addition of new buttons and onclick handlers to the diff page
- `ui/src/router/edit/EditRouterView.js` - Addition of additional effects and API calls upon submission
- `ui/src/router/components/form/components/VersionCreationSummary.js` - Creation of a new confirmation page for router version creation
- `sdk/turing/router/router.py` - Addition of a method to the `Router` SDK class to support router version creation
- `sdk/turing/session.py` - Addition of a support method to support router version creation
- `sdk/turing/generated/api/router_api.py` - Changes to OpenAPI autogenerated classes from addition of the new API endpoint method

### Tiny Fixes
- `sdk/requirements.txt` - Pinning of `cloudpickle==1.2.2` to the requirements list to ensure compatibility with the Mlflow dependency of the pyfunc ensembler service engine


romanwozniak:
Minor feedback: can we use the `secondary` color from the elastic EUI color schema (it [was removed](https://github.com/elastic/eui/pull/5323) in the newer version of EUI, but it can be replaced with a similar color) for the `Save` button, instead of this mustard-yellow color? 

Also, what do you all think about simplifying the button labels to `Save` and `Deploy` (from the proposed `Create without Deploying` and `Create and Deploy`) and instead, add the explanation about the action consequences into the confirmation pop-up window?

deadlycoconuts:
> Minor feedback: can we use the `secondary` color from the elastic EUI color schema (it [was removed](https://github.com/elastic/eui/pull/5323) in the newer version of EUI, but it can be replaced with a similar color) for the `Save` button, instead of this mustard-yellow color?

Ahah, I thought yellow looked interesting 😅 The `secondary` colour is actually green, which also happens to be the same colour as the 'Deployed' symbol, which may be a little confusing:
Screenshot 2022-03-15 at 11 44 35 AM
vs 
Screenshot 2022-03-15 at 11 52 28 AM



There's also `text` which doesn't look too bad:
Screenshot 2022-03-15 at 11 44 22 AM

Otherwise `success` ([successor](https://github.com/elastic/eui/pull/4874) to the `secondary` colour), `accent`, `ghost` all have a similar shade to the button below, which I think fades a little too much into the background: 
Screenshot 2022-03-15 at 11 51 26 AM

> Also, what do you all think about simplifying the button labels to `Save` and `Deploy` (from the proposed `Create without Deploying` and `Create and Deploy`) and instead, add the explanation about the action consequences into the confirmation pop-up window?
I think that sounds like a good idea, it frees up space and makes the UI look more minimalist.

romanwozniak:
> 

You can also consider somethinbg like this:
158302733-7a85774c-e5c7-485b-b462-1444f7c926dc

Or, use the different shades of blue for Save and Deploy buttons:

krithika369:
> Ahah, I thought yellow looked interesting 😅 The secondary colour is actually green, which also happens to be the same colour as the 'Deployed' symbol, which may be a little confusing:

I agree with this. But the gray color alternative is also the same as "undeployed". :)

Of all the options being discussed, my personal preference is different shades of blue for Save vs Deploy:
Screenshot 2022-03-15 at 12 05 24 PM

deadlycoconuts:
Alright, lemme get those colours up :D

krithika369:
> Also, what do you all think about simplifying the button labels to Save and Deploy (from the proposed Create without Deploying and Create and Deploy) and instead, add the explanation about the action consequences into the confirmation pop-up window?

Sounds great! The original specs say "Update", "Update and Deploy". I think "Save" and "Deploy" is concise and still conveys the intention.

@deadlycoconuts I haven't reviewed the MR yet but I'd like to confirm - this change is only for the "Edit Router" workflow, right? (The "Create" on the buttons got me wondering.)

deadlycoconuts:
> @deadlycoconuts I haven't reviewed the MR yet but I'd like to confirm - this change is only for the "Edit Router" workflow, right? (The "Create" on the buttons got me wondering.)

Ahah yes, I essentially just added an extra button in the existing "Edit Router" workflow (to... fork the existing workflow? if that's the right way to say it?). And oh I used "Create" because they were supposed to hint at "Create (Version) and (Deploy)" and "Create (Version) without Deploying" 😅

deadlycoconuts:
Would this colour scheme be okay? It's one that uses the default colour palette for the buttons; there doesn't seem to be a way to easily get the lighter-shade blue (because it isn't in the default palette) and I think we'll have to manually pass CSS properties in order to do that:
Screenshot 2022-03-15 at 1 02 37 PM
^ Furthermore the 'Previous' button now corresponds with the 'Cancel' button from the previous page.

romanwozniak:
> Would this colour scheme be okay? It's one that uses the default colour palette for the buttons; there doesn't seem to be a way to easily get the lighter-shade blue (because it isn't in the default palette) and I think we'll have to manually pass CSS properties in order to do that: Screenshot 2022-03-15 at 1 02 37 PM ^ Furthermore the 'Previous' button now corresponds with the 'Cancel' button from the previous page.
> 
> Edit: On second thought, we can just leave the customised values for this button without changing the palette (since this alternate colour isn't really used elsewhere): Screenshot 2022-03-15 at 2 10 26 PM

I like the first option more, especially the part, that 'Previous' is `EuiLinkButton`

api/turing/api/router_versions_api.go
```diff
@@ -33,6 +34,46 @@ func (c RouterVersionsController) ListRouterVersions(
 	return Ok(routerVersions)
 }
 
+// CreateRouterVersion creates a router version from the provided configuration. If no router exists
+// within the provided project with the provided id, this method will throw an error.
+// If the update is valid, a new RouterVersion will be created but NOT deployed.
+func (c RouterVersionsController) CreateRouterVersion(r *http.Request, vars RequestVars, body interface{}) *Response {
```
romanwozniak, line 40: I expect that more of existing code can be re-used here. Current implementation of the `Edit Router` feature, CREATES a new router version and DEPLOYS it immediately. This new `CreateRouterVersion` does the same except that it doesn't deploy the new version after the new version is saved into the DB. Which makes me thinking that that for both usecases, the same controller method can be used, with extra `?deploy=true|false` parameter passed to distinguish between these two scenarios.

ui/src/router/components/form/components/VersionCreationSummary.js
```diff
@@ -0,0 +1,16 @@
+import React, { Fragment } from "react";
+import { EuiSpacer } from "@elastic/eui";
+
+export const VersionCreationSummary = ({ router }) => {
+  return (
+    <Fragment>
+      <p>
+        You're about to create a new version for the Turing router{" "}
+        <b>{router.name}</b> in the <b>{router.environment_name}</b> environment
+        without deploying it.
```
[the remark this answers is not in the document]
deadlycoconuts: Hmm that's true. I think I based that on what was written for the deployment confirmation (in `DeploymentSummary.js`):
```
You're about to deploy a new Turing router <b>{router.name}</b> into{" "}
<b>{router.environment_name}</b> environment.
```
which I guess, going by the same logic, shouldn't mention the environment either?

But yeah I'll update it to make it clearer and less... excessive in providing random information.

api/api/specs/routers.yaml
```diff
@@ -250,6 +250,40 @@ paths:
           description: "Invalid project_id or router_id"
         404:
           description: "No router versions found"
+    put:
```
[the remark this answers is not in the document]
krithika369: +1. For the routers API, the `PUT` to update a router makes sense. For the `/versions` API, creating a new version should be `POST`.

api/turing/api/router_versions_api.go
krithika369: I have a similar but somewhat different view here. At the moment, the corresponding services for these entities is mostly a data access layer. I feel more of the business logic should be moved there and the controller layer kept lean to only handle parsing the request payload, calling the service methods and returning the final response in the appropriate format (i.e., all the presentation details).

So, update router API will call:
1. routerService.CreateRouter
2. routerVersionService.CreateRouterVersion
3. deploymentService.DeployVersion // handles updating router states, current version of router entity

Create router version API will only need to call service methods 1 & 2.

This still allows reusability but keeping the controllers very intentional. On the downside, this is going to take some refactoring. We did this exercise for experiments after the first cut (when considering the possibility of versioning the APIs) and I think the code structure is cleaner.  If we agree to take this route for Turing as well, we can add it to the backlog (and leave this controller method as is for now).

sdk/tests/router/router_test.py
```diff
@@ -152,7 +152,7 @@ def test_update_router(turing_api, active_project, actual, expected, use_google_
     turing.set_project(active_project.name)
 
     base_router = turing.Router(
-        id=1,
+        id=191,
```
krithika369, line 155: Why is this changed here? :)

ui/src/router/edit/EditRouterView.js
```diff
@@ -8,12 +8,16 @@ import { VersionComparisonView } from "./components/VersionComparisonView";
...
 import { ConfirmationModal } from "../../components/confirmation_modal/ConfirmationModal";
+import { VersionCreationSummary } from "../components/form/components/VersionCreationSummary";
 
 const EditRouterView = ({ projectId, currentRouter, ...props }) => {
   const { data: updatedRouter } = useContext(FormContext);
   const [showDiffView, toggleDiffView] = useToggle();
 
-  const [submissionResponse, submitForm] = useTuringApi(
+  const [
+    submissionCreateVersionWithDeploymentResponse,
```
krithika369, line 18: nit: For brevity, I'd prefer to distinguish them as `updateRouterResponse` vs `createRouterVersionResponse`. Implicitly, it's understood that update router will update the deployment as well and create version only creates a version.

krithika369 (commented):
Thanks for the comprehensive PR and updating the SDK as well. As a side note, we also need to think about updating the user docs (which are currently duplicated in the internal repo for Confluence but we have public docs in this repo too) when the UI implementation changes.

api/turing/api/router_versions_api.go
krithika369: @romanwozniak @deadlycoconuts thoughts?

deadlycoconuts: Hmm that sounds possible but like you said it's going to require quite a big amount of refactoring to reduce the controller layer to something of a request handler layer. I don't really have an opinion on an option that I prefer because I don't really have the context that the ones who decided on the relative responsibilities the controller/service layers should have, so I'm open to anything.

romanwozniak: The proposal makes total sense for me, but I don't think that it should be in the scope of this PR as it requires major refactoring of the other controller/service parts. So [this](https://github.com/gojek/turing/pull/179#discussion_r826757089) change is still preferred by me for now

krithika369: Alright. I don't think the refactoring should be in this PR either, will add it to the backlog.

ui/src/router/edit/EditRouterView.js
deadlycoconuts: Alright, that makes sense - internally it aligns with the `create` and `update` terms that we'll be using in the backend too.

sdk/tests/router/router_test.py
deadlycoconuts: Ahah, it's because I just wanted to align it with the `id` of the router in the mocked response [here](https://github.com/gojek/turing/blob/335efc9464f5ed99bcf9d65973205f1f36ea019f/sdk/tests/testdata/api_responses/create_router_0000.json#L2). The way the test is set-up doesn't require `id` to be checked so the tiny change here doesn't really do anything, it's more of a self-nitpick 😅

api/api/specs/routers.yaml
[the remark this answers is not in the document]
deadlycoconuts: Oops that's true, `POST` is indeed the better option. I'll be making that change.

ui/src/router/components/form/components/VersionCreationSummary.js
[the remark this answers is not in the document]
deadlycoconuts: Okay, I guess because `DeploymentSummary.js` is currently being used to display messages for both Router creation and update, `router.environment_name` is being mentioned in the confirmation summary for both use cases. I guess I can sacrifice a bit of code reuse here by splitting this into 2 files, `CreationSummary.js`, which is going to carry the existing message, and `UpdateSummary.js`, which will remove mention of the environment, and emphasise the deployment of the new version.

api/api/openapi.bundle.yaml
```diff
@@ -608,6 +608,12 @@ paths:
         schema:
           format: int32
           type: integer
+      - description: boolean indicating whether to deploy the new version once updated
+        in: query
+        name: deploy
+        required: true
+        schema:
+          type: boolean
```
deadlycoconuts, line 616: Added a new flag to be included in the query string `"?deploy=(true|false)"` for the endpoint `"/projects/{project_id}/routers/{router_id}"`, which is only required for `PUT` operations, i.e. updating routers with/without deployment

api/api/openapi.bundle.yaml
```diff
@@ -620,7 +626,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/RouterDetails'
+                $ref: '#/components/schemas/RouterVersion'
```
deadlycoconuts, line 629: To standardise the return values of the pre-existing endpoint with either `true` or `false` as the deploy flag for the `PUT` operation, I've made the operation return `RouterVersion` in both cases. Originally, updating a router returns `RouterDetails`, but since the same endpoint is now used for updating/saving a new router version, it would be odd to return 2 different objects. Furthermore, it'll be nice to return the new `RouterVersion` that's created even during an update, as it'll allow users to immediately retrieve the version number of the latest version (while it's being deployed).

api/turing/api/routers_api.go
```diff
@@ -155,24 +160,32 @@ func (c RoutersController) UpdateRouter(r *http.Request, vars RequestVars, body
...
-		if err != nil {
-			log.Errorf("Error deploying router %s:%s:%d: %v",
-				project.Name, router.Name, routerVersion.Version, err)
-		}
-	}()
-
-	return Ok(router)
+	if toDeploy {
+		go func() {
+			err := c.deployOrRollbackRouter(project, router, routerVersion)
```
deadlycoconuts, line 180: Creating a new router version now uses the same controller method used to update a router. The `toDeploy` flag, which is passed to the API as a URL query string, now determines which parts of the method to run, most notably the `deployOrRollbackRouter` call.

sdk/turing/router/router.py
```diff
@@ -130,20 +130,17 @@ def get(cls, router_id: int) -> 'Router':
...
-        self._config = config
-        updated_router = Router.from_open_api(
-            turing.active_session.update_router(router_id=self.id, router_config=config.to_open_api())
-        )
-        self.__dict__ = updated_router.__dict__
-        return self
+        version = turing.active_session.update_router(router_id=self.id,
+                                                      router_config=config.to_open_api(),
+                                                      deploy=True)
+        return RouterVersion(environment_name=self.environment_name, name=self.name, **version.to_dict())
```
deadlycoconuts, line 143: The SDK now reflects the change in the response when updating a router (with deployment). It now expects a `RouterVersion` instead of a updated `Router`.

ui/src/router/edit/EditRouterView.js
```diff
@@ -26,28 +28,36 @@ const EditRouterView = ({ projectId, currentRouter, ...props }) => {
   useEffect(() => {
     if (submissionResponse.isLoaded && !submissionResponse.error) {
       addToast({
-        id: "submit-success-create",
-        title: "Router configuration is updated!",
+        id: withDeployment ? "update-success" : "save-success",
+        title: withDeployment
+          ? "Router configuration is updated!"
+          : "Router configuration is saved (but not deployed)!",
```
deadlycoconuts, line 34: Not sure if this message in the pop-up dialog is sufficient for the user, but in any case, the user most certainly would've been warned in the confirmation window.

api/api/openapi.bundle.yaml
romanwozniak: It's not your fault, but this still doesn't satisfy the criteria of `PUT` being an idempotent operation. So I would rather standardize this `create/create and deploy` operation under `POST /projects/{project_id}/routers/{router_id}/versions`

deadlycoconuts: Hmm, yeah, it's a little awkward for the same endpoint to have somewhat different operations depending on the query string; makes it hard to place it in either group.

But just to clarify, if we're moving to both `create/create and deploy` operations to `POST /projects/{project_id}/routers/{router_id}/versions`, it means we're letting the `router_versions` API handle them, as opposed to the `routers` API previously?

api/api/specs/routers.yaml
```diff
@@ -13,7 +13,10 @@ info:
 .timeout: &timeout
   type: "string"
   pattern: '^[0-9]+(ms|s|m|h)$'
-  
+
+.deployFlag: &deployFlag
+  type: boolean
```
romanwozniak, line 18: This doesn't make sense. Such YAML anchors are only added for the re-usability (meaning, for schemas, that are used more than once). This can be just defined on the operation itself

api/api/specs/routers.yaml
```diff
@@ -124,7 +133,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/RouterDetails"
+                $ref: "#/components/schemas/RouterVersion"
```
romanwozniak, line 136: This makes me uncomfortable. The operation is on `router` entity, so the response is expected to be the `RouterDetails` and not `RouterVersion`. 

The `RouterVersion` response would be expected, if the operation is done on `router_versions` (i.e. `POST /projects/{project_id}/routers/{router_id}/versions`)

api/api/specs/routers.yaml
```diff
@@ -111,6 +114,12 @@ paths:
           schema:
             <<: *id
           required: true
+        - in: "query"
+          name: "deploy"
+          description: "boolean indicating whether to deploy the new version once updated"
+          schema:
+            <<: *deployFlag
+          required: true
```
romanwozniak, line 122: Shall we set it as `dafault=true`?

api/turing/api/base_controller.go
```diff
@@ -85,3 +85,11 @@ func (c BaseController) getRouterVersionFromRequestVars(
 	}
 	return routerVersion, nil
 }
+
+func (c BaseController) getDeployFlagFromRequestVars(vars RequestVars) (deployFlag bool, error *Response) {
```
romanwozniak, line 89: This is weird. This deploy flag, unlike project/router, is only meaningful in the context of the `reouters_api` controller. Also, I'm not sure if it makes sense to define a separate function just for this, because this parameter is only extracted in a single controller's method (moving it into a function makes sense for the re-usability)

romanwozniak (commented):
My motivation for using `?deploy=true|false` flag was primarily defined by my understanding of the `Edit Router` call to be in fact:
- `Create New Router Version`
- (if deploy=true) Then `Promote this New Version to serve Router's traffic`

So, in this context adding this `?deploy` flag makes sense (to me). However, as @krithika369 mentioned before, if you see this call as `Edit router configuration with the new version`, then this `deploy` flag is confusing.

In the longer term, it doesn't really matter because after [the API controllers are refactored](https://github.com/gojek/turing/pull/179#discussion_r828725503), then most of the business logic will be moved out from the controllers and there will be less of code duplication.

Currently these multiple conditional checks 
```go
if toDeploy {
  // ...
}
```
are not contributing to the code clarity, so maybe we should separate these two operations altogether. So in addition to existing 
`PUT /proejects/{project_id}/routers/{router_id}`
We can just add 
`POST /projects/{project_id}/routers/{router_id}/versions` which will create a new version without promoting it.

wdyt @krithika369

api/api/specs/routers.yaml
```diff
@@ -250,6 +250,40 @@ paths:
           description: "Invalid project_id or router_id"
         404:
           description: "No router versions found"
+    post:
```
deadlycoconuts, line 253: Turned this into a `POST` operation as suggested.

api/turing/api/router_versions_api.go
```diff
@@ -139,6 +180,12 @@ func (c RouterVersionsController) Routes() []Route {
 			path:    "/projects/{project_id}/routers/{router_id}/versions",
 			handler: c.ListRouterVersions,
 		},
+		{
+			method:  http.MethodPost,
+			path:    "/projects/{project_id}/routers/{router_id}/versions",
+			body:    request.CreateOrUpdateRouterRequest{},
+			handler: c.CreateRouterVersion,
+		},
```
deadlycoconuts, line 188: Pulled this back out from the recycle bin given the latest proposed changes.

deadlycoconuts:
Okay thanks everyone for the clarification! I've just reverted the changes made to this PR to make it align with the latest proposal to keep the 2 endpoints for the 2 functionalities separate:

- `PUT /projects/{project_id}/routers/{router_id}`
- `POST /projects/{project_id}/routers/{router_id}/versions`

Lemme know if there's anything you'd like to add, @romanwozniak and @krithika369

sdk/tests/router/router_test.py
```diff
@@ -268,6 +268,56 @@ def test_list_versions(turing_api, active_project, generic_router, generic_route
...
+
+    responses.add(
+        method="POST",
+        url=f"/v1/projects/{active_project.id}/routers/{base_router.id}/versions",
+        body=json.dumps(generic_router_version, default=tests.json_serializer),
+        status=200,
+        content_type="application/json"
+    )
+
+    actual_response = base_router.create_version(generic_router_config)
```
krithika369, line 293: Why are we creating a router here for the comparison? Why can we not simply compare using a `RouterVersion` entity?

sdk/turing/router/router.py
```diff
@@ -171,6 +171,16 @@ def list_versions(self) -> List['RouterVersion']:
         return [RouterVersion(environment_name=self.environment_name, name=self.name, **ver.to_dict())
                 for ver in response]
 
+    def create_version(self, config: RouterConfig) -> RouterVersion:
```
krithika369, line 174: Can this simply be a `create` method on the RouterVersion class? That's consistent with the API. Although the UI is designed to be Save vs Create router, I don't think the SDK needs to follow the same convention and I feel the `RouterVersion.create` would be more standard.

ui/src/router/components/form/components/UpdateSummary.js
```diff
@@ -0,0 +1,15 @@
+import React, { Fragment } from "react";
+import { EuiSpacer } from "@elastic/eui";
+
+export const UpdateSummary = ({ router }) => {
```
krithika369, line 4: Can this be `RouterUpdateSummary`, to match `RouterCreationSummary` ?

ui/src/router/edit/EditRouterView.js
```diff
@@ -1,19 +1,20 @@
...
 import { ConfirmationModal } from "../../components/confirmation_modal/ConfirmationModal";
+import { VersionCreationSummary } from "../components/form/components/VersionCreationSummary";
 
 const EditRouterView = ({ projectId, currentRouter, ...props }) => {
-  const { data: updatedRouter } = useContext(FormContext);
+  const { data: routerConfig } = useContext(FormContext);
   const [showDiffView, toggleDiffView] = useToggle();
 
-  const [submissionResponse, submitForm] = useTuringApi(
+  const [updateRouterResponse, updateRouterForm] = useTuringApi(
```
krithika369, line 17: Nit: `updateRouterForm` sounds like the form component rather than an actionable function name. Can it be called `updateRouterCallback` or `submitUpdateRouter` ? 

Same for the create version.

ui/src/router/edit/EditRouterView.js
```diff
@@ -57,13 +108,17 @@ const EditRouterView = ({ projectId, currentRouter, ...props }) => {
...
-            updatedRouter={updatedRouter}
+            updatedRouter={routerConfig}
             onPrevious={toggleDiffView}
             onSubmit={onSubmit}
-            isSubmitting={submissionResponse.isLoading}
+            isSubmitting={
+              updateRouterResponse.isLoading ||
+              createRouterVersionResponse.isLoading
+            }
+            setWithDeployment={setWithDeployment}
```
krithika369, line 118: Minor suggestion: I think rather than `setWithDeployment` prop (which sounds like a detail internal to the `EditRouterView` component, we can create 2 onsubmit props on the `VersionComparisonView` like:

```javascript
<VersionComparisonView
...
onSave={() => {setWithDeployment(false); onSubmit();}
onDeploy={() => {setWithDeployment(true); onSubmit();}
```

sdk/samples/router/edit_a_router.py
```diff
@@ -0,0 +1,195 @@
...
+from turing.router.config.resource_request import ResourceRequest
+from turing.router.config.log_config import LogConfig, ResultLoggerType
+from turing.router.config.traffic_rule import TrafficRule, HeaderTrafficRuleCondition
+from turing.router.config.enricher import Enricher
+from turing.router.config.router_ensembler_config import DockerRouterEnsemblerConfig
+from turing.router.config.common.env_var import EnvVar
+from turing.router.config.experiment_config import ExperimentConfig
+
+
+def main(turing_api: str, project: str):
```
krithika369, line 17: Since we show all the steps in detail in the create router sample, I wonder if this one can be kept concise.
1. Get an existing router (in reality, this needn't exist but this is only a sample and the user should put in whatever router if they want)
2. Update some property > Deploy it
3. Update some other property  > Save it (i.e., create version only)

api/api/specs/routers.yaml
```diff
@@ -250,6 +250,40 @@ paths:
...
+      tags: *tags
+      summary: "Create router version without deploying it"
+      parameters:
+        - in: "path"
+          name: "project_id"
+          description: "project id of the project of the router"
```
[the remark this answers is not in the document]
deadlycoconuts: Okay I've also taken the chance to throw out `project id` from the other endpoint descriptions that have the same awkward phrasing.

api/api/specs/routers.yaml
```diff
@@ -250,6 +250,40 @@ paths:
...
+          schema:
+            <<: *id
+          required: true
+      requestBody:
+        description: "router configuration to save"
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/RouterConfig"
```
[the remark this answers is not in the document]
deadlycoconuts: Yep, that would make the entire request body more logical, and reduces unnecessary checks (as you've pointed out below in the API function that updates routers/creates new router versions). It would also make updating routers a little less cumbersome with the SDK as users would no longer need to handle the `environment_name` and `(router_)name` attributes.

Would calling the new struct `RouterUpdateConfig` be a little clearer? I was thinking that the name would reflect its intended use to be in 'requests only'. 

And if we're introducing the new struct right now, I think we should probably change the API/UI/SDK all at once too, in order to avoid breaking anything. Or, we could do everything in a separate task. But I'm okay with doing everything right now.

sdk/tests/router/router_test.py
deadlycoconuts: I think this is linked to your observation in the comment below that `create_version` is a method of the `Router` class; creating a version requires a base router object to exist so that's why the test required creating a router. But of course, if we were to move `create_version` to a method of `RouterVersion` then there'll be no need to create a router here.

sdk/turing/router/router.py
deadlycoconuts: Yep that sounds logical. I'll shift the `create_version` to be a `RouterVersion` class method then. Maybe I'll rename it to `create` too so that `RouterVersion.create` would mirror `Router.create` in the SDK.

krithika369: > However, this would also now mean that we'll have to explicitly specify the router_id of the router that we'll like to create this new version for as an explicit argument in the RouterVersion.create method, since RouterConfig instances don't contain any ids.

That sounds logical to me.

api/api/specs/routers.yaml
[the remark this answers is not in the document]
krithika369: I thought we could deal with the update router API (request body change from `RouterConfig` -> `RouterVersionConfig`) later on. In this case, the suggested change should only affect the SDK (as the classes are auto-generated) and can be carried out in this MR.

As you've noted, the update router API request body change will affect parts of the API too. And in general, there are other APIs about which the same argument can be made (eg: ensembler name and type), i.e., should the update request expect fields that are not allowed to be changed. So we can capture this in the backlog and revisit at a later time.

api/turing/api/router_versions_api.go
```diff
@@ -33,6 +34,47 @@ func (c RouterVersionsController) ListRouterVersions(
...
+	}
+	if router, errResp = c.getRouterFromRequestVars(vars); errResp != nil {
+		return errResp
+	}
+
+	requestConfig := body.(*request.RouterConfig)
+
+	// This is used temporarily in order to reuse the BuildRouterVersion method
+	dummyRouterRequest := request.CreateOrUpdateRouterRequest{
+		Environment: "",
```
deadlycoconuts, line 55: Even though we're using `RouterConfig` without `EnvironmentName` and (Router) `Name` as the request body, there's a need to construct this dummy `request.CreateOrUpdateRouterRequest` object because the existing `BuildRouterVersion` struct method is defined only for the said struct.

In the future when the refactoring of all the controller/service methods are done, this dummy object can be removed.

sdk/requirements.txt
```diff
@@ -7,3 +7,4 @@ pandas==1.2.2
 python-dateutil
 requests==2.25.1
 urllib3 >= 1.25.3
+cloudpickle==1.2.2
```
deadlycoconuts, line 10: This has now been pinned here as a fix for the pyfunc ensembler engine compatibility issues.

sdk/samples/router/edit_a_router.py
```diff
@@ -0,0 +1,61 @@
...
+    versions = router_1.list_versions()
+    for v in versions:
+        print(v)
+
+    # 2. When you CREATE a router version, Turing will save the new version, but not deploy it.
+    new_router_config_to_save = router_1.config
+    new_router_config_to_save.resource_request.min_replica = 0
+
+    new_undeployed_version = turing.router.config.router_version.RouterVersion.create(new_router_config_to_deploy,
+                                                                                      router_id=1)
```
deadlycoconuts, line 51: This sample now reflects how a router version is to be created using `RouterVersion.create` instead of using a `Router` class method as performed previously.

sdk/tests/conftest.py
```diff
@@ -444,7 +444,10 @@ def generic_router_version(
         id=2,
         created_at=datetime.now() + timedelta(seconds=20),
         updated_at=datetime.now() + timedelta(seconds=20),
-        router=None,
+        router=turing.generated.models.Router(
+            environment_name="test_env",
+            name="test_router"
+        ),
```
deadlycoconuts, line 450: This is done to accommodate changes to how the `RouterVersion` SDK class stores a copy of the `environment_name` and (router) `name`. It keeps these information so that users can easily extract a `RouterConfig` SDK (which as of present contains these fields) that can be used to 1) create new routers, 2) update routers or 3) create new router versions. 

In the future when (2) gets refactored and no longer requires the `environment_name` and (router) `name` anymore, these fields will only serve the creation of new routers (1).

sdk/turing/router/config/router_config.py
```diff
@@ -49,8 +49,8 @@ class RouterConfig:
     ensembler: Union[RouterEnsemblerConfig, Dict] = None
 
     def __init__(self,
-                 environment_name: str,
-                 name: str,
+                 environment_name: str = None,
+                 name: str = None,
```
deadlycoconuts, line 53: Introducing the default `None` argument values so users can now construct `RouterConfig` SDK classes to create new router versions (which do not require these values). In a way, `RouterConfig` is a single common interface for 3 main use cases, 1) creating new routers, 2) updating routers, and 3) creating new router versions.

deadlycoconuts:
I've refactored the create router version API method to use the downsized `RouterVersionConfig`, and also for the UI and SDK to reflect this change. In the future, when refactoring `UpdateRouter` to use the same struct in the request body, we should be able to cut down and remove some redundant code in both the UI and the SDK which is currently supporting the slightly bulkier `RouterConfig` (with the `EnvironmentName` and (Router) `Name`) for update router operations.

sdk/turing/router/config/router_version.py
```diff
@@ -53,3 +56,22 @@ def get_config(self) -> RouterConfig:
...
+        """
+        Creates a new router version for the router with the given router_id WITHOUT deploying it
+
+        :param config: configuration of router
+        :param router_id: router id of the router for which this router version will be created
+        :return: the new router version
+        """
+        version = turing.active_session.create_router_version(
+            router_id=router_id,
+            router_version_config=config.to_open_api(use_generic_config=True)
```
krithika369, line 71: We can construct the `RouterVersionConfig` here manually for now (in line with the comment above to remove `use_generic_config`).

krithika369 (approved):
Left some small suggestions. Thanks for completing the MR with the sample and e2e tests too. LGTM :)

api/api/specs/routers.yaml
```diff
@@ -603,6 +637,45 @@ components:
             ensembler:
               $ref: "#/components/schemas/RouterEnsemblerConfig"
 
+    RouterVersionConfig:
```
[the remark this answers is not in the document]
deadlycoconuts: Oh yeah you're right. I was initially concerned about some backward compatibility issues with the SDK, which was why I intentionally left that out, but I think those concerns are unfounded. I'll be updating the specs and the SDK with the change you're suggested.

sdk/turing/router/config/router_version.py
deadlycoconuts: Okay I'll leave a comment there to say that it'll be removed once the refactoring is complete.

deadlycoconuts: I've simplified this further by just extracting the `RouterVersionConfig` directly from `RouterConfig`, so there's now no need for this temporary thing.

sdk/turing/session.py
```diff
@@ -273,6 +273,14 @@ def list_router_versions(self, router_id: int) -> List[RouterVersion]:
             router_id=router_id
         )
 
+    @require_active_project
```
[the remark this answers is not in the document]
deadlycoconuts: Hmm I guess it doesn't hurt to have one, but if we're doing it, shouldn't it be an instance method of `Router` because we could simply use an `Router` instance's `id` attribute to call the  `create_router_version` method you've quoted, or simply the `RouterVersion.create` class method, both of which require a `router_id` as an argument.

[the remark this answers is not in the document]
krithika369: Apologies. I did mean to say "instance method".

deadlycoconuts:
Thanks @krithika369 for all the responses and suggestions. I've refactored the `CreateOrUpdateRouterRequest` to move `BuildExperimentEngineConfig` and `BuildRouterVersion` to the `request.RouterConfig struct`. There aren't many changes beyond that (except for the some refactoring of the unit tests). 

I'll be merging this branch if nothing goes wrong with the CI pipeline!

PR #181, opened 2022-03-15
Add env vars support to Pyfunc ensembler

deadlycoconuts:
In PR #164, PR #165, PR #170, PR #171, and PR #174 an automated pyfunc ensembler building feature was introduced for Turing. However, the existing implementation lacked support for passing environment variables specified by the user to the running pyfunc ensembler service in its container.

This PR introduces support for the aforementioned feature, by adding an extra `env` field to the `EnsemblerPyfuncConfig`/`pyfuncConfig`/`pyfunc_config` struct/model/schema used in Turing API/UI/SDK. Turing API/UI/SDK has also been correspondingly updated to allow users to specify the new field when deploying/viewing routers with pyfunc ensemblers.

## Additions
- `api/api*` - Changes to the OpenAPI specs from the addition of the new `env` field
- `api/turing/models/ensembler.go` - Addition of the new `env` field to `EnsemblerPyfuncConfig`
- `api/turing/service/router_deployment_service.go` - Slight changes to the `buildEnsemblerServiceImage` to store `env` as part of an `EnsemblerDockerConfig`
- `ui/src/router/components/configuration/components/pyfunc_config_section/PyFuncConfigViewGroup.js` - Addition of a new field to display environment variables in the configuration details panel
- `ui/src/router/components/form/components/pyfunc_config/PyFuncConfigFormGroup.js` - Addition of a new field for users to input environment variables when configuring Pyfunc ensemblers as part of a router deployment
- `ui/src/services/ensembler/PyFuncEnsembler.js` - Addition of the new `env` field to `PyFuncEnsembler`
- `sdk/turing/generated/model/ensembler_pyfunc_config.py` - Changes to the OpenAPI autogenerated class corresponding to the pyfunc ensembler configuration
- `sdk/turing/router/config/router_ensembler_config.py` - Addition of the new `env` field to `pyfunc_config`


api/db-migrations/000011_add_pyfunc_config_env.up.sql
```diff
@@ -0,0 +1,10 @@
+-- Migrate pyfunc_config data from old schema to new schema (with new 'env' field set as an empty array by default)
+UPDATE ensembler_configs
+SET pyfunc_config = (SELECT json_build_object(
+        'env', '[]'::jsonb,
```
deadlycoconuts, line 4: Not sure if this is really necessary since the only pyfunc ensemblers that have been created with this field are the ones in dev, but for consistency's sake I thought having a migration script for this would make things neater. Alternatively, we could consider implementing some lazy migration of the saved pyfunc ensemblers in the API?

ui/src/router/components/configuration/components/pyfunc_config_section/PyFuncConfigTable.js
```diff
@@ -2,7 +2,7 @@ import React, { useContext } from "react";
 import { EuiDescriptionList } from "@elastic/eui";
 import EnsemblersContext from "../../../../../providers/ensemblers/context";
 
-export const PyFuncRefConfigTable = ({ config: { ensembler_id } }) => {
+export const PyFuncConfigTable = ({ config: { ensembler_id } }) => {
```
deadlycoconuts, line 5: Fixed this for consistency

api/db-migrations/000011_add_pyfunc_config_env.down.sql
```diff
@@ -0,0 +1,9 @@
+-- Migrate pyfunc_config data from new schema to old schema (remove the env column)
```
krithika369, line 1: Tip: this can be simplified using the `-` operator like: https://stackoverflow.com/a/35283553

api/db-migrations/000011_add_pyfunc_config_env.up.sql
krithika369: Actually I feel it's better to not have this in the migration, just to keep the migration history lean.

This feature is still under development. :) We don't really need to worry since we haven't cut any release with migration file # 10 (and as a consequence, we also haven't released this feature to our internal users, as you've noted). We can run this data migration manually on the internal dev DB.

krithika369 (approved):
LGTM! 🚀

deadlycoconuts:
Thanks for the comments @krithika369 ! FYI, I'm also going to bring in the tiny little refactoring commit from the SDK minor release into this PR :) I'll be merging this if the pipeline passes 🚀

PR #182, opened 2022-03-16
Reorder undeployment service which skips secret deletion

deadlycoconuts:
### Context
The existing router deployment (`DeployRouterVersion`) and undeployment  (`UndeployRouterVersion`) functions do not build and teardown components in the same order. While in most cases, this should not pose any issues, in instances where `UndeployRouterVersion` is used as a subroutine to teardown a router version that has been partially deployed (but has subsequently encountered an error) using `DeployRouterVersion`, a potential problem can occur whereby artifacts created during the deployment fail to get deleted.

One such occurrence is with the implementation of `deployOrRollbackRouter`, which calls `deployRouterVersion` and subsequently `DeployRouterVersion` to deploy a router version, but triggers `UndeployRouterVersion` if it encounters an error during deployment. 

Of particular concern are the secret creation/deletion and service object creation steps (this occurs in both deployment and undeployment). Previously during undeployment, `createServices` gets called before secret deletion occurs. A non-`nil` error returned would cause `UndeployRouterVersion` to terminate prematurely before any secret artifacts get deleted: 

```go
// UndeployRouterVersion removes the deployed router, if exists. Else, an error is returned.
func (ds *deploymentService) UndeployRouterVersion(...) error {
	...
	// Construct service objects for each of the components to be deleted
	services, err := ds.createServices(
		routerVersion, project, ds.environmentType, "", nil,
		ds.routerDefaults, ds.sentryEnabled, ds.sentryDSN,
		ds.knativeServiceConfig.TargetConcurrency, ds.knativeServiceConfig.QueueProxyResourcePercentage,
		ds.knativeServiceConfig.UserContainerLimitRequestFactor,
	)
	if err != nil {
		return err
	}
	var errs []string

	// Delete secret
	eventsCh.Write(models.NewInfoEvent(models.EventStageDeletingDependencies, "deleting secrets"))
	secret := ds.svcBuilder.NewSecret(routerVersion, project, "", "", "")
	err = deleteSecret(controller, secret)
	if err != nil {
		errs = append(errs, err.Error())
	}
	...
}
```

Hence, this PR simply proposes a rearrangement of the different substeps performed in both `DeployRouterVersion` and `UndeployRouterVersion`, to ensure that any router deployment teardown by the latter function proceeds up till the same point of failure in the former function, meaning that any artifacts created during deployment up until failure will get removed.

cc @leonlnj

### DIFF
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -150,6 +150,17 @@ func (ds *deploymentService) DeployRouterVersion(
 	}
 	secretName := secret.Name
 
+	// Construct service objects for each of the components and deploy
+	services, err := ds.createServices(
+		routerVersion, project, ds.environmentType, secretName, experimentConfig,
+		ds.routerDefaults, ds.sentryEnabled, ds.sentryDSN,
+		ds.knativeServiceConfig.TargetConcurrency, ds.knativeServiceConfig.QueueProxyResourcePercentage,
+		ds.knativeServiceConfig.UserContainerLimitRequestFactor,
+	)
+	if err != nil {
+		return endpoint, err
+	}
+
 	// Deploy fluentd if enabled
 	if routerVersion.LogConfig.ResultLoggerType == models.BigQueryLogger {
 		fluentdService := ds.svcBuilder.NewFluentdService(routerVersion, project,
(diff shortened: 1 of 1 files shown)


leonlnj:
LGTM

leonlnj (approved):

PR #183, opened 2022-03-21
Upgrade Go version/Knative/Spark Operator/Kubernetes Client/Docker Compose for dev environment.

ashwinath:
In lieu of keeping all our libraries and dependencies up to date, in this PR, we are updating everything to allow compatibility with the following:

- Kubernetes 1.22
- Istio 12.5
- Knative 1.0
- Spark Operator v1beta2-1.3.3-3.1.1

I would not suggest using this in our clusters at the moment as the version difference is very large. For instance, [kubectl supports only maximum of 1 skew version](https://kubernetes.io/releases/version-skew-policy/#kubectl). Merlin is also taking the same approach here: https://github.com/gojek/merlin/pull/221. Should we merge this into a separate branch instead of the main branch?

This also requires the cluster to have the networking/v1beta1 API installed for the Istio APIs. The other API versions remains unchanged; Spark and Knative components are using the same version.

Also, Go version has been upgraded to 1.16 to prevent compilation issues with the newer libraries. I have kept it the same as what [Knative](https://github.com/knative/serving/blob/177aec6ffed895ee1d470e7e2de0810ae93659c0/go.mod#L3) and [Istio](https://github.com/istio/istio/blob/6332f0901f96ca97cf114d57b466d4bcd055b08c/go.mod#L3) uses.

Trying to upgrade beyond 1.16 yields the following error which I'm not sure how to solve:
```
github.com/gojek/turing/api/turing/service imports
        golang.org/x/oauth2/google imports
        cloud.google.com/go/compute/metadata: ambiguous import: found package cloud.google.com/go/compute/metadata in multiple modules:
        cloud.google.com/go v0.97.0 (${HOME}/go/pkg/mod/cloud.google.com/go@v0.97.0/compute/metadata)
        cloud.google.com/go/compute v1.1.0 (${HOME}/go/pkg/mod/cloud.google.com/go/compute@v1.1.0/metadata)
```

EDIT:

Found a way to upgrade to Go 1.18. Turing API, Experiment and Router has been upgraded to Go 1.18. The golangci-linter has been upgraded to the latest version to support the newer language features.

I have also changed the CI to allow the Github Actions to run when a PR targets any branch (previously only main).


romanwozniak:
> Should we merge this into a separate branch instead of the main branch?

Yes, this makes sense.

> I would not suggest using this in our clusters at the moment as the version difference is very large

Could you elaborate on what's the proposed plan then? 
1) Branch from main and prepare the new release that is guaranteed to work with the upgraded knative/istio
2) Upgrade cluster dependencies
3) Deploy the new release into the upgraded cluster
4) Merge the branch with `gojek:main`

Is this right? 

> Trying to upgrade beyond 1.16 yields the following error which I'm not sure how to solve

What have you tried so far? Any analysis on the root cause?

ashwinath:
> Could you elaborate on what's the proposed plan then?

Originally I was thinking along the lines of these:


For our current deployments with the old clusters, things should remain the same. But for the new deployments, a separate pipeline should mimic the old pipeline, with the difference of the target Kubernetes clusters and different tagged versions. We should tag the new branch and build separately from the main branch. In summary it should look something like this:


Old pipeline:
1. Tag new version on main branch.
2. Deploy to current Kubernetes clusters.


New pipeline:
1. Tag new version on new branch.
2. Deploy to new Kubernetes cluster.
3. Rebase any new changes on main and test again (repeat this step as many times as needed).


Reconciliation:
1. Merge to main
2. Tag new version
3. Target new clusters.


However @krithika369 was suggesting merging to main directly after the 1.4 release (correct me if I am wrong here). Any updates required should be merged to the 1.4 branch and tagged again. It will still require a new pipeline for testing on the new cluster. In summary it should look something like this:


Old pipeline:
1. Branch out release 1.4 and tag it.
2. Deploy to current Kubernetes clusters.


New pipeline:
1. Tag new version on main.
2. Deploy to new Kubernetes cluster.
3. Repeat steps 1 - 2 as many times as needed

No reconciliation needed because releases will be done periodically with the new branching method. I prefer @krithika369's method.

> What have you tried so far? Any analysis on the root cause?

I will attend to this later as I am currently testing the batch ensembling part as of now.

ashwinath:
Some clarifications to the above comment:

This branch will only be merged into main after the 1.4 release has been cut; i.e. this PR goes into the 1.5 release. Any urgent fixes to 1.4 will be merged into the 1.4 release and main branches.

.github/workflows/turing.yaml
```diff
@@ -374,10 +374,10 @@ jobs:
         with:
           name: cluster-init.${{ env.CLUSTER_INIT_VERSION }}.tar
 
-      - name: Set up Go 1.14
+      - name: Set up Go 1.16
         uses: actions/setup-go@v2
         with:
-          go-version: "1.14"
+          go-version: "1.16"
```
romanwozniak, line 380: There are two `Set up Go` steps in this job (probably it's an artefact of the merge op). This one can be removed

api/e2e/test/01_create_router_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
```
romanwozniak, line 2: The old style build tags (`+build ...`) can be removed now

ashwinath: Unfortunately still required: https://github.com/gojek/turing/runs/5749127170?check_suite_focus=true#step:17:24

https://go.dev/doc/go1.16

> The new [go/build/constraint](https://go.dev/pkg/go/build/constraint/) package parses build constraint lines, both the original // +build syntax and the //go:build syntax that will be introduced in Go 1.17. This package exists so that tools built with Go 1.16 will be able to process Go 1.17 source code. See [https://golang.org/design/draft-gobuild](https://go.dev/design/draft-gobuild) for details about the build constraint syntaxes and the planned transition to the //go:build syntax. Note that //go:build lines are not supported in Go 1.16 and should not be introduced into Go programs yet.

But the linter fails if I do not put the //go:build syntax. But let me try upgrading to go 1.18, I think I found a way to run Turing API on Go 1.18 and it might change this outcome.

ashwinath: Removed as upgraded to 1.18.

ashwinath:
> > Trying to upgrade beyond 1.16 yields the following error which I'm not sure how to solve
> 
> What have you tried so far? Any analysis on the root cause?


Managed to find a way to use Go 1.18. Someone had an issue with upgrading from 1.16 as well: https://github.com/aws-observability/aws-otel-collector/issues/926. I upgraded experiments and router to Go 1.18 as well.


The other upgrade was the golangci-linter as well, upgraded to the latest version 1.45.2 to support Go 1.18.

.github/workflows/cluster-init.yaml
```diff
@@ -14,8 +14,6 @@ on:
 
   # Automatically run CI on branches, that have active PR opened
   pull_request:
-    branches:
-      - main
```
romanwozniak, line 18: Would it be better to instead add the `upgrade-knative` into the list and TODO about why is it there and when should it be removed?

.github/workflows/turing.yaml
```diff
@@ -211,10 +209,10 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
-      - name: Set up Go 1.14
+      - name: Set up Go 1.18
         uses: actions/setup-go@v2
         with:
-          go-version: "1.14"
+          go-version: "1.18"
```
romanwozniak, line 215: Curious, is it possible to use env var to control go-version everywhere?

.github/workflows/turing.yaml
```diff
@@ -433,7 +426,9 @@ jobs:
           VAULT_TOKEN: root
         run: |
           echo "::group::Wait for Vault to become available"
-          timeout --foreground 45 bash -c \
+          # TODO: remove
+          curl -v -H "X-Vault-Token: ${{ env.VAULT_TOKEN }}" http://vault.127.0.0.1.nip.io/v1/sys/mounts
```
romanwozniak, line 430: Is it still required?

.github/workflows/turing.yaml
```diff
@@ -446,6 +441,8 @@ jobs:
           }
           EOF
 
+          echo credentials.json
```
romanwozniak, line 444: Needs to be cleaned up?

api/go.mod
```diff
@@ -63,27 +185,31 @@ replace (
...
-	k8s.io/kube-controller-manager => k8s.io/kube-controller-manager v0.16.6
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.16.6
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.16.6
-	k8s.io/kubectl => k8s.io/kubectl v0.16.4
-	k8s.io/kubelet => k8s.io/kubelet v0.16.6
-	k8s.io/kubernetes => k8s.io/kubernetes v1.16.4
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.16.6
-	k8s.io/metrics => k8s.io/metrics v0.16.6
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.16.6
+	k8s.io/apiextensions-apiserver => k8s.io/apiextensions-apiserver v0.22.7
```
romanwozniak, line 190: Do we still need these replacements after the upgrade of all of the dependencies?

api/turing/cluster/controller_test.go
```diff
@@ -99,6 +100,7 @@ func TestDeployKnativeService(t *testing.T) {
 	// Define tests
 	cs := knservingclientset.NewSimpleClientset()
 	tests := map[string][]reactor{
+		// TODO: problematic for new_service
```
romanwozniak, line 103: What does it mean?

romanwozniak (commented):
Thanks, it lgtm. Development documentation should also be updated to reflect the changes to the Golang version

api/go.mod
ashwinath: Yes, these 0.16.6 corresponds to k8s 1.16, so 0.22 corresponds to 1.22. But if you are asking about the aliasing then yes it is needed still.

api/turing/cluster/controller_test.go
ashwinath: Oh this was supposed to be removed, I made a note that this test was failing at first but I fixed it.

.github/workflows/turing.yaml
ashwinath: Removing this.

.github/workflows/turing.yaml
ashwinath: Removing this.

.github/workflows/cluster-init.yaml
ashwinath: I was thinking of this as well but I'm not sure which is better. On one hand you could have other target branches to merge into if we are supporting a different release branch. But since we aren't doing that we can go with this approach of targeting the upgrade-knative branch.

.github/workflows/turing.yaml
ashwinath: Let me change it to use an envvar instead.

ashwinath: Additionally i changed it for the Go linter as well.

.github/workflows/cluster-init.yaml
ashwinath: I changed this to main + upgrade-knative.

ashwinath:
> Thanks, it lgtm. Development documentation should also be updated to reflect the changes to the Golang version

I have updated the Go versions as well as all the dependency versions such as Knative/Istio/Spark etc.

romanwozniak (approved):

PR #184, opened 2022-03-21
[Documentation]: Experiment Engine plugins

romanwozniak:
# Context:
[Turing v1.3.0](https://github.com/gojek/turing/releases/tag/v1.3.0) introduced support of the runtime experiment engine plugins. This (overdue) PR adds an example implementation of the experiment engine plugin, updates Turing CI pipeline to use this example plugin in the E2E test suite, and provides developers documentation on how to build a plugin from scratch. 

# Details:
This PR includes a ["serverless" experiment engine plugin implementation](https://github.com/romanwozniak/turing/tree/exp_engine_plugins_docs/engines/experiment/examples/plugins/hardcoded). This plugin reads a hardcoded list of experiments from the plugin configuration (provided at deployment time), and then assigns available treatments to request based on the traffic split configuration (traffic split is done on the experiment unit (`customer_id`) by using `mod` operator on the segmentation unit hash).


.github/workflows/turing.yaml
```diff
@@ -102,6 +102,37 @@ jobs:
...
+        run: |
+          set -o pipefail
+          make build-image | tee output.log
+          echo "::set-output name=test-experiment-engine-plugin-version::${{ env.VERSION }}"
+
+      - name: Save Docker image
+        run: |
+          docker image save \
+            --output test-experiment-engine-plugin.${{ steps.build-image.outputs.test-experiment-engine-plugin-version }}.tar \
+            plugin-example-engine-plugin:${{ steps.build-image.outputs.test-experiment-engine-plugin-version }}
```
terryyylim, line 127: CMIIW, since output `test-experiment-engine-plugin-version` is essentially using `env.VERSION`, is there any reason why we don't set the environment variable `VERSION` at the job level instead, since it's used by all steps except the checkout stage?

engines/experiment/docs/rpc_plugins.md
```diff
@@ -0,0 +1,398 @@
...
+
+type FooBarExperimentRunner struct {}
+
+func (*FooBarExperimentRunner) Configure(cfg json.RawMessage) error {
+	log.Infof("Configuration: %s", string(cfg))
+	return nil
+}
+```
+[`hashicorp/go-plugin`](https://github.com/hashicorp/go-plugin) uses [`hclog`](https://github.com/hashicorp/go-hclog) 
+library for logging, and it expects that the plugin's log stream will be written into the standard output as `hlog`
```
terryyylim, line 285: Is it `hlog` or `hclog`?

terryyylim (commented):
Thanks for the PR, docs look good! Left some small comments.

.github/workflows/turing.yaml
romanwozniak: Yup, this makes sense. I probably just followed the structure of the other `build` jobs, but it's not reasonable for this temporary artifact. Updated

engines/experiment/docs/rpc_plugins.md
```diff
@@ -0,0 +1,398 @@
...
+Which is composed of the `ExperimentManager` interface of:
+```go
+type ExperimentManager interface {
+    // GetEngineInfo returns the configuration of the experiment engine
+    GetEngineInfo() (Engine, error)
+    
+    // ValidateExperimentConfig validates the given Turing experiment config for the expected data and format
+    ValidateExperimentConfig(cfg json.RawMessage) error
+    
+    // GetExperimentRunnerConfig converts the given config (as retrieved from the DB) into a format suitable
```
terryyylim, line 109: Another question here just to understand this more deeply, if `GetExperimentRunnerConfig` is only called at the time of router deployment, would the router require a re-deployment if the underlying experiment variable for the Router changes (e.g addition of a new segmenter)? If not, how would the update be propagated?

romanwozniak: To understand your question better, let me ask if this is what you mean:

1. There are some experiments configured in the experiment engine
2. Turing router, that uses this experiment engine is deployed
3. Experiments in the experiment engine are updated (i.e experiment variables, segmenters etc.)

4. ???

If this is the scenario you are talking about, then yes – it's expected that the new version of the router must be created and the configuration on how to extract the new variable from the request should be added into this version.

terryyylim: Yes that's what I meant - thanks for clarifying! Would this above workflow description be of value to add to the documentation?

romanwozniak: This documentation is expected to cover primarily the development process of plugins. The general information of the experiment engine and its contract was expected to be in [developer_guide.md](https://github.com/gojek/turing/blob/main/engines/experiment/docs/developer_guide.md). I don't mind adding it into either of the two docs, but it feels like an isolated chunk of the documentation, which isn't very useful without the overhauling of the [developer_guide.md](https://github.com/gojek/turing/blob/main/engines/experiment/docs/developer_guide.md)

terryyylim: 👌🏻  We can park that for future dev_guide documentation improvement then.

romanwozniak: cc @krithika369

krithika369: I somehow feel this may be better in the user workflow doc than dev (the user needs to remember to update the router deployment if critical experiment config that affects the run experiment call, no matter the engine, changes). If the engine used is the in-house Turing Experiments, we can probably help them by having some advanced cross-app checks and balances (may be in the long term roadmap) and alert the user. But I think such changes may be infrequent and kind of obvious to the user. So I'm not too concerned.

leonlnj (approved):

PR #186, opened 2022-03-23
Add wait for status update

deadlycoconuts:
## Context
This update to the Turing SDK adds commits to the existing release of Turing SDK (`v0.2.0`), which serve to introduce a `version` attribute in the `Router` SDK. This `version` attribute (as well as a router's other attributes) gets updated when `wait_for_status` gets called on a `Router` instance. When `wait_for_status` returns successfully, the `version` attribute would be expected to contain the current router version number. 

This update is mainly targeted at scenarios when one updates a router - right after calling `update`, the existing router version that is deployed remains the same while the new version enters the `pending` state. Instead of having to query the status of the Router actively in order to retrieve the new router version id when it has finally been deployed successfully, the user can simply be assured that his/her `Router` instance has a newly updated version id when the waiting function has returned succesfully.

### DIFF
--- a/sdk/turing/router/config/experiment_config.py
+++ b/sdk/turing/router/config/experiment_config.py
@@ -29,9 +29,9 @@ class ExperimentConfig:
 
     @config.setter
     def config(self, config: Dict):
-        if config is not None and 'project_id' in config:
-            config['project_id'] = int(config['project_id'])
         self._config = config
+        if self._config is not None and 'project_id' in self._config:
+            self.config['project_id'] = int(self._config['project_id'])
 
     def to_open_api(self) -> OpenApiModel:
         if self.config is None:
--- a/sdk/turing/router/config/router_ensembler_config.py
+++ b/sdk/turing/router/config/router_ensembler_config.py
@@ -59,11 +59,12 @@ class RouterEnsemblerConfig:
         if isinstance(standard_config, turing.generated.models.EnsemblerStandardConfig):
             self._standard_config = standard_config
         elif isinstance(standard_config, dict):
-            standard_config['experiment_mappings'] = [
+            openapi_standard_config = standard_config.copy()
+            openapi_standard_config['experiment_mappings'] = [
                 turing.generated.models.EnsemblerStandardConfigExperimentMappings(**mapping)
                 for mapping in standard_config['experiment_mappings']
             ]
-            self._standard_config = turing.generated.models.EnsemblerStandardConfig(**standard_config)
+            self._standard_config = turing.generated.models.EnsemblerStandardConfig(**openapi_standard_config)
         else:
             self._standard_config = standard_config
 
@@ -76,11 +77,12 @@ class RouterEnsemblerConfig:
         if isinstance(docker_config, turing.generated.models.EnsemblerDockerConfig):
             self._docker_config = docker_config
         elif isinstance(docker_config, dict):
-            docker_config['resource_request'] = \
-                turing.generated.models.ResourceRequest(**docker_config['resource_request'])
-            docker_config['env'] = [turing.generated.models.EnvVar(**env_var) for env_var in docker_config['env']]
+            openapi_docker_config = docker_config.copy()
+            openapi_docker_config['resource_request'] = \
+                turing.generated.models.ResourceRequest(**openapi_docker_config['resource_request'])
+            openapi_docker_config['env'] = [turing.generated.models.EnvVar(**env_var) for env_var in openapi_docker_config['env']]
             self._docker_config = turing.generated.models.EnsemblerDockerConfig(
-                **docker_config
+                **openapi_docker_config
             )
         else:
             self._docker_config = docker_config
--- a/sdk/turing/router/router.py
+++ b/sdk/turing/router/router.py
@@ -33,7 +33,6 @@ class Router(ApiObject):
                  environment_name: str,
                  monitoring_url: str,
                  status: str,
-                 version: int = None,
                  config: Dict = None,
                  endpoint: str = None,
                  **kwargs):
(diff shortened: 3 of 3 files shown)


sdk/turing/router/router.py
```diff
@@ -47,8 +48,10 @@ def __init__(self,
 
         if config is not None:
             self._config = RouterConfig(name=name, environment_name=environment_name, **config)
+            self._version = config.get('version')
```
krithika369, line 51: I think I probably missed it in the [original PR](https://github.com/gojek/turing/pull/171).

Why can't the implementation just be:

```python
    @property
    def version(self) -> int:
        return self._config.get('version') if self._config or None
```

Why do we need the `version` argument on init?

krithika369 (approved):
Minor comment: In the future, may be clearer to name the branch on the origin like `sdk_v0.2.0-branch` (just because we have differences in the app vs sdk numbering and it also follows the tag name format to prefix with 'sdk'). Anyway, we can nuke these branches when they've served their purpose.

sdk/turing/router/router.py
```diff
@@ -128,7 +135,8 @@ def update(self, config: RouterConfig) -> 'Router':
         Update the current router with a new set of configs specified in the RouterConfig argument
 
         :param config: configuration of router
-        :return: instance of router updated (self)
+        :return: instance of router (self); this router contains details of the currently deployed router; it contains
+            the details of the currently DEPLOYED router version
```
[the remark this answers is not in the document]
deadlycoconuts: Hmmm okay now that I'm seeing this again, I'm starting to think what I wrote sounds a little clunky... mmm, what about, `instance of router (self); this router contains details of the currently deployed router and router version`?

sdk/turing/router/router.py
deadlycoconuts: You're right, there's no need for version to be up there in the constructor.

On the other hand, we can't just remove the second line in this `if` conditional and replace it with the getter as you've suggested, because `self._config`, which is a `RouterConfig` instance, does not contain a `version` attribute:
```python
 if config is not None:
            self._config = RouterConfig(name=name, environment_name=environment_name, **config)
            self._version = config.get('version')
```
Instead, we need to extract the `version` from the `config` argument, which is only available when the `Router` instance is being constructed.

deadlycoconuts: Actually, now that I think of it, I might just refactor the `config` getter this way:
```python
 @property
    def config(self) -> `RouterConfig`:
        return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)
```
Doing so will allow us to keep the `version` getter that you've suggested, as well as remove the awkward if conditional in the constructor.

deadlycoconuts: In short, we'll get something like:

```python
...
   def __init__(self,
                 id: int,
                 name: str,
                 project_id: int,
                 environment_name: str,
                 monitoring_url: str,
                 status: str,
                 version: int = None,
                 config: Dict = None,
                 endpoint: str = None,
                 **kwargs):
        super(Router, self).__init__(**kwargs)
        self._id = id
        self._name = name
        self._project_id = project_id
        self._environment_name = environment_name
        self._endpoint = endpoint
        self._monitoring_url = monitoring_url
        self._status = RouterStatus(status)
        self._config = config
...
 @property
    def config(self) -> `RouterConfig`:
        return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)

@property
def version(self) -> int:
    return self._config.get('version') if self._config else None
```

krithika369: Ah right. Makes sense to me.

sdk/turing/router/router.py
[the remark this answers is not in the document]
krithika369: How about `instance of router (self); this router contains details of the router and its currently deployed version`? :)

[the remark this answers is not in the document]
deadlycoconuts: Sounds perfect :)

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -76,11 +77,12 @@ def docker_config(self, docker_config: turing.generated.models.EnsemblerDockerCo
         if isinstance(docker_config, turing.generated.models.EnsemblerDockerConfig):
             self._docker_config = docker_config
         elif isinstance(docker_config, dict):
-            docker_config['resource_request'] = \
-                turing.generated.models.ResourceRequest(**docker_config['resource_request'])
-            docker_config['env'] = [turing.generated.models.EnvVar(**env_var) for env_var in docker_config['env']]
+            openapi_docker_config = docker_config.copy()
+            openapi_docker_config['resource_request'] = \
+                turing.generated.models.ResourceRequest(**openapi_docker_config['resource_request'])
+            openapi_docker_config['env'] = [turing.generated.models.EnvVar(**env_var) for env_var in openapi_docker_config['env']]
```
deadlycoconuts, line 83: I've made some additional changes - they look pretty minor but they were causing some bugs in the unit tests. 

Essentially, the previous implementation of the setter methods were mutating the dictionaries that were being passed to it, which is dangerous and might lead to unintended effects. These new changes simply create new working dictionaries that are only used within the scope of the setter.

cc @krithika369

sdk/turing/router/config/experiment_config.py
```diff
@@ -29,9 +29,9 @@ def config(self) -> Dict:
 
     @config.setter
     def config(self, config: Dict):
-        if config is not None and 'project_id' in config:
-            config['project_id'] = int(config['project_id'])
         self._config = config
+        if self._config is not None and 'project_id' in self._config:
+            self.config['project_id'] = int(self._config['project_id'])
```
deadlycoconuts, line 34: Same issue as the one I just commented on, though the fix here was to simply reorder the statements.

sdk/turing/router/config/router_ensembler_config.py
krithika369: Sounds good. Thanks for addressing them. We need to propagate these changes to main later as well.

deadlycoconuts:
Okay thanks for your comments! I'll be merging this now.

PR #188, opened 2022-03-25
Fix pyfunc bugs for release

deadlycoconuts:
### Context

This PR aims to include a couple of minor bug fixes and refactoring before the upcoming minor release of Turing:
- Renaming of the `features` keyword argument of the abstract `ensemble` method in `EnsemblerBase` of the SDK to `input` for less ambiguity when implementing Pyfunc ensemblers
- Transformation of the `list` of route responses in real-time ensembler requests to a `dict` of route responses indexed by route names in the SDK `PyFunc` interface, so as to improve usability and consistency of the `predictions` keyword argument of the aforementioned `ensemble` method
- Addition of a name search query to the Turing API endpoint for listing registered ensemblers, so as to resolve the non-functioning ensembler search bar in the Turing UI
- Fixing of e2e batch prediction tests due to the non-deterministic behaviour of the experiment engine
- (Temporary) downgrading of support for the Turing SDK, ensembler job engine and ensembler service engine to **_only_** Python `3.7.*`. This is to align Turing with Merlin's support for Python of that particular minor version.

### DIFF
--- a/api/api/openapi.bundle.yaml
+++ b/api/api/openapi.bundle.yaml
@@ -48,7 +48,8 @@ paths:
         schema:
           default: 10
           type: integer
-      - in: query
+      - description: Searches ensembler name for a partial match of the search text
+        in: query
         name: search
         schema:
           type: string
@@ -176,7 +177,9 @@ paths:
         schema:
           format: int32
           type: integer
-      - in: query
+      - description: Searches ensembler name used for the ensembling job for a partial
+          match of the search text
+        in: query
         name: search
         schema:
           type: string
--- a/api/api/openapi.bundle.yaml
+++ b/api/api/openapi.bundle.yaml
@@ -49,6 +49,10 @@ paths:
           default: 10
           type: integer
       - in: query
+        name: search
+        schema:
+          type: string
+      - in: query
         name: type
         schema:
           $ref: '#/components/schemas/EnsemblerType'
--- a/api/api/openapi.bundle.yaml
+++ b/api/api/openapi.bundle.yaml
@@ -177,8 +177,8 @@ paths:
         schema:
           format: int32
           type: integer
-      - description: Searches ensembler name used for the ensembling job for a partial
-          match of the search text
+      - description: Searches ensembling job name used for a partial match of the
+          search text
         in: query
         name: search
         schema:
--- a/api/api/specs/ensemblers.yaml
+++ b/api/api/specs/ensemblers.yaml
@@ -33,6 +33,10 @@ paths:
             type: integer
             default: 10
         - in: query
+          name: search
+          schema:
+            type: string
+        - in: query
           name: type
           schema:
             $ref: "#/components/schemas/EnsemblerType"
--- a/api/api/specs/ensemblers.yaml
+++ b/api/api/specs/ensemblers.yaml
@@ -34,6 +34,7 @@ paths:
             default: 10
         - in: query
           name: search
+          description: Searches ensembler name for a partial match of the search text
           schema:
             type: string
         - in: query
--- a/api/turing/service/ensembler_service.go
+++ b/api/turing/service/ensembler_service.go
@@ -1,6 +1,8 @@
 package service
 
 import (
+	"fmt"
+
 	"github.com/gojek/turing/api/turing/models"
 	"github.com/jinzhu/gorm"
 )
@@ -23,6 +25,7 @@ type EnsemblersFindByIDOptions struct {
 type EnsemblersListOptions struct {
 	PaginationOptions
 	ProjectID     *models.ID            `schema:"project_id" validate:"required"`
+	Search        *string               `schema:"search"`
 	EnsemblerType *models.EnsemblerType `schema:"type" validate:"omitempty,oneof=pyfunc docker"`
 }
 
@@ -66,6 +69,10 @@ func (service *ensemblersService) List(options EnsemblersListOptions) (*Paginate
 	query := service.db
 	if options.ProjectID != nil {
 		query = query.Where("project_id = ?", options.ProjectID)
+	}
+
+	if options.Search != nil && len(*options.Search) > 0 {
+		query = query.Where("name like ?", fmt.Sprintf("%%%s%%", *options.Search))
 	}
 
 	if options.EnsemblerType != nil {
--- a/api/turing/service/ensembler_service.go
+++ b/api/turing/service/ensembler_service.go
@@ -72,7 +72,7 @@ func (service *ensemblersService) List(options EnsemblersListOptions) (*Paginate
 	}
 
 	if options.Search != nil && len(*options.Search) > 0 {
-		query = query.Where("name like ?", fmt.Sprintf("%%%s%%", *options.Search))
+		query = query.Where("name ilike ?", fmt.Sprintf("%%%s%%", *options.Search))
 	}
 
 	if options.EnsemblerType != nil {
--- a/engines/pyfunc-ensembler-job/environment.yaml
+++ b/engines/pyfunc-ensembler-job/environment.yaml
@@ -1,6 +1,6 @@
 name: pyfunc-ensembler-job
 dependencies:
-  - python=3.8
+  - python=3.7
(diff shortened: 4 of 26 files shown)


api/api/specs/ensemblers.yaml
```diff
@@ -32,6 +32,10 @@ paths:
           schema:
             type: integer
             default: 10
+        - in: query
+          name: search
+          schema:
+            type: string
```
deadlycoconuts, line 38: Added this to the API endpoint so the ensembler search bar should now work.

engines/pyfunc-ensembler-service/tests/conftest.py
```diff
@@ -11,25 +11,13 @@ def initialize(self, artifacts: dict):
...
-    @staticmethod
-    def get_route_name_to_id_mapping(predictions):
-        """
-        Helper function to look through the predictions returned from the various routes and to map their names to
-        their id numbers (the order in which they are found in the payload.
-        """
-        route_name_to_id = {}
-        for i, pred in enumerate(predictions):
-            route_name_to_id[pred['route']] = i
-        return route_name_to_id
```
deadlycoconuts, line 32: Removed this unnecessary helper method since the `list` of route responses -> `dict` of route responses transformation is now all done behind the scenes by the `PyFunc` interface.

sdk/turing/version.py
```diff
@@ -1 +1 @@
-VERSION = "0.1.0"
+VERSION = "0.3.0"
```
deadlycoconuts, line 1: I think I forgot to update this for the release of Turing `0.2.0` but it's now going to updated as usual.

infra/e2e/turing.values.yaml
```diff
@@ -34,11 +34,11 @@ turing:
             - name: treatment-1
           variants_configuration:
             control:
-              traffic: 0.85
+              traffic: 1.0
               treatment_configuration:
                 foo: bar
             treatment-1:
-              traffic: 0.15
+              traffic: 0.0
```
deadlycoconuts, line 41: @romanwozniak , I didn't manage to pick these changes up in your previous PR, but were you intending to demonstrate some treatment assignment behaviour when you introduced these configs and swapped `nop` for `proprietary` here? https://github.com/gojek/turing/blob/d12014c897a815d3526a41d390dc360e53a3a1a7/api/e2e/test/testdata/create_router_nop_logger_proprietary_exp.json.tmpl#L15
I think it's causing some odd undeterministic behaviour in the e2e tests in the CI pipeline like here:
https://github.com/gojek/turing/runs/5673171262?check_suite_focus=true#step:18:74

api/turing/service/ensembler_service.go
```diff
@@ -68,6 +71,10 @@ func (service *ensemblersService) List(options EnsemblersListOptions) (*Paginate
 		query = query.Where("project_id = ?", options.ProjectID)
 	}
 
+	if options.Search != nil && len(*options.Search) > 0 {
+		query = query.Where("name like ?", fmt.Sprintf("%%%s%%", *options.Search))
```
krithika369, line 75: `ilike` may be better, for case-insensitive search.

sdk/turing/version.py
krithika369: How often do we manually update it? And what would be the benefit of doing so? I wonder if this can just be left at `0.0.0` always, since the `make version` command overwrites this file with the correct version during build anyway.

api/api/specs/ensemblers.yaml
krithika369: Would be good to include a description for this field as it's not immediately clear from the name what we are searching. For example, something like:`Searches ensembler name for a partial match of the search text`

krithika369 (approved):
Left some small comments. LGTM, pending the discussion about the e2e test case. Thank you!

sdk/turing/version.py
deadlycoconuts: > How often do we manually update it?

I think judging from... how the original string was `0.1.0`, and that's our first release... we've probably never updated it at all 😅

>And what would be the benefit of doing so? I wonder if this can just be left at 0.0.0 always, since the make version command overwrites this file with the correct version during build anyway.

That is true. Maybe I'll just roll it back to `0.0.0` like you suggested, and leave a comment to say that's no need to update that field manually.

infra/e2e/turing.values.yaml
romanwozniak: @deadlycoconuts https://github.com/gojek/turing/pull/189

romanwozniak: Don't forget to revert this before merging @deadlycoconuts

api/api/openapi.bundle.yaml
```diff
@@ -172,7 +177,9 @@ paths:
         schema:
           format: int32
           type: integer
-      - in: query
+      - description: Searches ensembler name used for the ensembling job for a partial
```
krithika369, line 180: Isn't it just ensembling job name? 😄 The job name is generated from the name of the ensembler, with some suffix. But ultimately, the search is carried out on the ensembling job's name.

deadlycoconuts: Oh that's true too xD Haha alright I'll simplify the name back to what it should plainly be 😅

engines/pyfunc-ensembler-job/environment.yaml
```diff
@@ -1,6 +1,6 @@
 name: pyfunc-ensembler-job
 dependencies:
-  - python=3.8
+  - python=3.7
```
krithika369, line 3: I think we need to update these for the engines too: https://github.com/gojek/turing/blob/main/engines/pyfunc-ensembler-job/setup.py#L25

deadlycoconuts:
@romanwozniak https://github.com/gojek/turing/runs/5715592714?check_suite_focus=true#step:18:78 🥲  I'll go look at it again and see what went wrong.

romanwozniak:
> @romanwozniak https://github.com/gojek/turing/runs/5715592714?check_suite_focus=true#step:18:78 🥲 I'll go look at it again and see what went wrong.

I have only two assumptions:
 * `"hash/fnv"` hashing is non-deterministic (I doubt that)
 * There is something wrong with the `/batch_predict` endpoint – i.e. response from one prediction request in the batch is being assigned to the wrong request before marshaling
 This is the raw response, so the issue has nothing to do with the JSON comparison

```
01_create_router_test.go:84: Response Payload:
        [{"code":200,"data":{"request":{"client":{"id":4}},"response":{"route_responses":[{"route":"control","data":{"version":"control"},"is_default":true}],"experiment":{"configuration":{"foo":"bar"}}}}},{"code":200,"data":{"request":{"client":{"id":7}},"response":{"route_responses":[{"route":"control","data":{"version":"control"},"is_default":true}],"experiment":{"configuration":{"foo":"bar"}}}}}]
```

You probably can merge this MR and keep the fix to the E2E test out of scope for now

romanwozniak:
@deadlycoconuts actually, it's neither nor. I can reproduce the issue by running the unit test ( which I added recently) multiple times. However, the hashing is consistent. It could be related to the floating-point precision or something similar. Let  me try to figure it out

deadlycoconuts:
Thanks a lot for your help for the time being @romanwozniak !

romanwozniak:
> Thanks a lot for your help for the time being @romanwozniak !

TIL: `range` operation over `map` is not deterministic

romanwozniak:
> Thanks a lot for your help for the time being @romanwozniak !

@deadlycoconuts It's fixed now. Sorry, for hijacking your PR. The diff is here: 
https://github.com/gojek/turing/pull/188/files/481c6933929dbea18854ad9ebab07a1b3ba6e612..808818c4285983ef243ab6612545b4fff91e51a0

deadlycoconuts:
Haha no worries, thanks a lot for digging into and solving the mystery of this really elusive bug! 😂

deadlycoconuts:
Okay I'll be merging this if there aren't anything outstanding! Thanks a lot everyone for your input! 🚀

PR #189, opened 2022-03-25
[BUG] Fix End-to-End test suite 

romanwozniak:
# Context:
https://github.com/gojek/turing/pull/184 contained the updates to the e2e tests to cover the deployment of a Turing Router with the experiment engine plugin. It turned out, that one of the tests started failing intermittently because of the non-deterministic behavior (https://github.com/gojek/turing/pull/188#discussion_r835065627).

The problem occurs only with the test, that asserts the `/batch_predict` endpoint response. Previously, the request payload to the `/batch_predict` contained two equal prediction request payloads (the entire request was `[{}, {}]`), however, https://github.com/gojek/turing/pull/184 has changed this to verify that the experiment engine assigns treatments correctly to different requests. 

I'm not able to verify, that the experiment engine assigns treatments non-deterministically, so I assume that the problem was observed because of how the expected and actual responses were compared.

This PR adds a unit test for the test experiment engine's `ExperimentRunner. GetTreatmentForRequest` implementation and updates the e2e test part of the `/batch_predict` response assertion.

### DIFF
--- a/api/e2e/test/01_create_router_test.go
+++ b/api/e2e/test/01_create_router_test.go
@@ -111,7 +111,7 @@ func TestCreateRouter(t *testing.T) {
 						}
 					},
 					{
-						"code": 200
+						"code": 200,
 						"data": {
 							"request": {
 								"client": {
--- a/api/e2e/test/01_create_router_test.go
+++ b/api/e2e/test/01_create_router_test.go
@@ -87,7 +87,7 @@ func TestCreateRouter(t *testing.T) {
 						"code": 200,
 						"data": {
 							"request": {
-								"customer": {
+								"client": {
 									"id": 4
 								}
 							},
@@ -114,7 +114,7 @@ func TestCreateRouter(t *testing.T) {
 						"code": 200
 						"data": {
 							"request": {
-								"customer": {
+								"client": {
 									"id": 7
 								}
 							},
--- a/api/e2e/test/01_create_router_test.go
+++ b/api/e2e/test/01_create_router_test.go
@@ -84,6 +84,7 @@ func TestCreateRouter(t *testing.T) {
 					t.Logf("Response Payload:\n%s", string(responsePayload))
 					expectedResponse := `[
 					{
+						"code": 200,
 						"data": {
 							"request": {
 								"customer": {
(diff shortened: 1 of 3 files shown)


engines/experiment/examples/plugins/hardcoded/runner_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"testing"
+
+	"github.com/gojek/turing/engines/experiment/manager"
+	"github.com/gojek/turing/engines/experiment/pkg/request"
+	"github.com/gojek/turing/engines/experiment/runner"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestExperimentRunner_GetTreatmentForRequest(t *testing.T) {
```
deadlycoconuts, line 15: Thanks for adding this test as an initial check to see if it's on the sample experiment's engine works :)

deadlycoconuts:
Thanks a lot for coming up with this! 

>I'm not able to verify, that the experiment engine assigns treatments non-deterministically, so I assume that the problem was observed because of how the expected and actual responses were compared.

I went to take a closer look at how the example experiment engine and the batch prediction engine works but nothing seems to suggest non-deterministic behaviour as you suggested. Maybe it's as you have mentioned, just due to the original way the responses were compared, which is interesting how it throws out results that were seemingly random.

Once again, thanks a lot for the fix!

api/e2e/test/01_create_router_test.go
```diff
@@ -81,40 +81,62 @@ func TestCreateRouter(t *testing.T) {
 					assert.Equal(t, http.StatusOK, response.StatusCode,
 						"Unexpected response (code %d): %s",
 						response.StatusCode, string(responsePayload))
-					actualResponse := gjson.GetBytes(responsePayload, "#.data.response").String()
```
deadlycoconuts, line 84: Somehow when you rewrote the entire expected and actual responses explicitly, the tests seem to work perfectly fine but I'm still unsure of the original cause of the non-deterministic test outcomes initially. One possible (but still unlikely I feel) cause might've been the way the responses were extracted from the original JSON body with the `gjson` package in that the response order might not be preserved, but I can't confirm this 👀

deadlycoconuts (approved):

api/e2e/test/01_create_router_test.go
romanwozniak: Yes, may assumptions are along these lines too

PR #190, opened 2022-03-29
Update workflows to use Python 3.7 for ensembler engines and sdk

deadlycoconuts:
### Context
With the PR #188, the Python versions for the Turing SDK as well as both ensembler engines have been fixed at `3.7.*`. While these changes have been made on these components, the workflows for these components have not been updated yet.

This PR serves to update these workflows to utilise Python 3.7 accordingly.

### Additional Fixes
- Tiny fix to the Turing SDK to handle cases whereby `RouterConfig` is not returned as part of a response body containing Routers
- Tiny fix to the Turing UI to allow for text wrapping when the docker image name is too long

### DIFF
--- a/sdk/turing/router/router.py
+++ b/sdk/turing/router/router.py
@@ -1,7 +1,7 @@
 import time
 import logging
 
-from typing import List, Dict
+from typing import List, Dict, Optional
 
 import turing.generated.models
 from turing._base_types import ApiObject, ApiObjectSpec
@@ -75,8 +75,11 @@ class Router(ApiObject):
         return self._status
 
     @property
-    def config(self) -> 'RouterConfig':
-        return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)
+    def config(self) -> Optional['RouterConfig']:
+        if self._config is not None:
+            return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)
+        else:
+            return None
 
     @property
     def version(self) -> int:
--- a/ui/src/components/config_section/index.scss
+++ b/ui/src/components/config_section/index.scss
@@ -11,7 +11,7 @@
 .euiPanel--configSection {
   .euiDescriptionList.euiDescriptionList--responsiveColumn {
     > * {
-      @include euiTextBreakWord;
+      @include euiTextTruncate;
 
       line-height: 2.1 !important;
       margin-top: 2px !important;
--- a/ui/src/components/config_section/index.scss
+++ b/ui/src/components/config_section/index.scss
@@ -11,7 +11,7 @@
 .euiPanel--configSection {
   .euiDescriptionList.euiDescriptionList--responsiveColumn {
     > * {
-      @include euiTextTruncate;
+      @include euiTextBreakWord;
 
       line-height: 2.1 !important;
       margin-top: 2px !important;
--- a/.github/workflows/pyfunc-ensembler-job.yaml
+++ b/.github/workflows/pyfunc-ensembler-job.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.7
 
       - name: Setup Java
         uses: actions/setup-java@v2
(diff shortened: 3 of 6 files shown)


romanwozniak (approved):

sdk/turing/router/router.py
```diff
@@ -75,8 +75,11 @@ def status(self) -> RouterStatus:
         return self._status
 
     @property
-    def config(self) -> 'RouterConfig':
-        return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)
+    def config(self) -> Optional['RouterConfig']:
+        if self._config is not None:
+            return RouterConfig(name=self.name, environment_name=self.environment_name, **self._config)
+        else:
+            return None
```
deadlycoconuts, line 82: Added this `if-else` check to handle cases whereby querying the API for a certain router returns an object without a `RouterConfig`. This occurs notably when a new Turing Router is pending (being deployed).

ui/src/components/config_section/index.scss
```diff
@@ -11,7 +11,7 @@
 .euiPanel--configSection {
   .euiDescriptionList.euiDescriptionList--responsiveColumn {
     > * {
-      @include euiTextTruncate;
+      @include euiTextBreakWord;
```
deadlycoconuts, line 14: @krithika369, I can't seem to override this mixin that truncates quite a number of text descriptions in other UI components - I tried creating custom style rules for the `Container` table but they don't seem to be able override this mixin either :/

krithika369: Oh we shouldn't change it globally for all rows.

> I tried creating custom style rules for the Container table

Could you share the example here? What was tried and where it was placed?

deadlycoconuts: So I tried creating this custom style class and using it for the `EuiDecriptionList` of a `ContainerConfigTable`, but it doesn't overwrite any of the truncation styles:

```scss
.euiDescriptionList--responsiveColumn.containerConfig {
  .euiDescriptionList__description {
    @include euiTextBreakWord;
  }
}
```

Explicitly specifying the style of the `euiTextBreakWord` mixin  as found [here](https://github.com/elastic/eui/blob/491840ed37e6b0bba0f731ef848fe45326e41e6b/src/global_styling/mixins/_typography.scss#L127) doesn't seem to work either:

```scss
.euiDescriptionList--responsiveColumn.containerConfig {
  .euiDescriptionList__description {
    overflow-wrap: break-word !important;
    word-wrap: break-word !important;
    word-break: break-word;
  }
}
```

deadlycoconuts: Another thing I've tried to try was to add `<div className="eui-textBreakWord">` tags to the `description` of the item corresponding to the image name within the `items` list of the `ContainerConfigTable` as:

```js
{
  title: "Image",
  description: (
      <div className="eui-textBreakWord">
        thisisaverylargexcepetionnamewithalotofwordsthatatsomepointaregoingtobetruncated
      </div>
    )
}
```
Although this doesn't 'truncate' the line and replace the missing characters with an ellipses, it doesn't break the line either... generating something like this instead: 
Screenshot 2022-03-30 at 10 59 51 AM

krithika369: I see.. I was going to suggest something along the lines of your second example to override the properties for the single row's description. But adding the `div` is probably messing with some other flex properties. I will try to have a stab at it to understand better.

deadlycoconuts: Thanks a lot! I'll see what I can try out from my end!

krithika369: Using `span` in combination with inline style overrides seems to do it:
```javascript
const items = [
    {
      title: "Image",
      description: (<span style={{
        wordBreak: "break-word",
        textOverflow: "unset",
        whiteSpace: "break-spaces"
      }} >
        {image}
      </span >),
    },
```

We're having to override some existing `!important` properties which doesn't seem easy with classes.

deadlycoconuts: Wow thanks for discovering that! I'll include your code snippet in!

deadlycoconuts:
Thanks everyone for your comments/approval! Merging this! 🚀

PR #191, opened 2022-04-01
Include response headers in logs

deadlycoconuts:
## Context
In PR #176, a change was introduced to merge the headers of the original request sent to Turing routers with that of the response returned from an enricher. The change in this PR complements the change previously made by introducing logging of the response headers of each response received by a Turing Router. 

## Modifications
- `engines/router/missionctl/handlers/logutils.go` - Addition of a step to turn response headers into `string`s to be logged
- `engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.pb.go` - Changes to the generated proto classes 
- `engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto` - Changes to the proto to include response headers
- `engines/router/missionctl/log/resultlog/resultlog.go` - Changes to the `AddResponse` method to include response headers as an argument


engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto
```diff
@@ -16,11 +16,15 @@ message Request {
 }
 
 message Response {
-    // The JSON response from a Turing component (Enricher / Experiment Engine / Router / Ensembler), UTF-8-encoded.
-    string response = 1;
+    // The JSON response body from a Turing component (Enricher / Experiment Engine / Router / Ensembler),
+    // UTF-8-encoded.
+    string response_body = 1;
```
deadlycoconuts, line 21: Renamed `response` to `response_body` because the addition of the new field `response_header` would create some ambiguity as to what `response` would really mean.

engines/router/missionctl/handlers/logutils.go
```diff
@@ -133,3 +142,8 @@ func uncompressHTTPBody(header http.Header, body []byte) ([]byte, error) {
 	}
 	return result, nil
 }
+
+func stringifyResponseHeader(header http.Header) (string, error) {
+	headerJSON, err := json.Marshal(header)
+	return string(headerJSON), err
+}
```
deadlycoconuts, line 149: A response header would basically be stored as a string representation of a JSON-ified response header.

engines/router/missionctl/log/resultlog/proto/turing/TuringResultLog.proto
```diff
@@ -16,11 +16,16 @@ message Request {
 }
 
 message Response {
-    // The JSON response from a Turing component (Enricher / Experiment Engine / Router / Ensembler), UTF-8-encoded.
-    string response = 1;
+    // The JSON response body from a Turing component (Enricher / Experiment Engine / Router / Ensembler),
+    // UTF-8-encoded.
+    string body = 1;
```
deadlycoconuts, line 21: Renamed `response` as `body` because the addition of the new field `header` would create some ambiguity as to what `response` would really mean.

krithika369: That's a valid point. But unfortunately, this will be a breaking change, not for the proto decoding itself, but rather, the downstream workflows that depend on it. For example, in most internal use cases, the Kafka messages are forwarded to BQ and the field name change would result in a change in the table as well.

Similarly, our own BQ logger uses the schema generated from this proto. This means, the next time users update their routers, they have to carry out a schema migration on the existing table as the only thing we handle automatically is addition of new columns ([ref](https://github.com/gojek/turing/blob/main/engines/router/missionctl/log/resultlog/bigquery.go#L238)).

We can look into a schema migration if required (which we have done once in the past) but it would require the users to take actions to migrate their existing data / scripts as well. So we should avoid it unless required.

Considering all of the above, I think `header`, `response` is not too outlandish and we can live with it.

engines/router/missionctl/log/resultlog/kafka_test.go
```diff
@@ -124,15 +126,22 @@ func TestNewProtobufKafkaLogEntry(t *testing.T) {
 	_, turingLogEntry := makeTestTuringResultLogEntry(t)
 	// Overwrite the turing request id value
 	turingLogEntry.TuringReqId = "testID"
+	// Manually remove the key-value pair corresponding to "Content-Type" in the router and enricher header to prevent
+	// non-deterministic behaviour when the hashmap in the log gets serialised
+	delete(turingLogEntry.Enricher.Header, "Content-Encoding")
+	delete(turingLogEntry.Router.Header, "Content-Encoding")
 	// Run newProtobufKafkaLogEntry and validate
 	key, message, err := newProtobufKafkaLogEntry(turingLogEntry)
+	fmt.Println(message)
```
krithika369, line 135: Is this meant to be a debug statement?

engines/router/missionctl/log/resultlog/kafka_test.go
```diff
@@ -124,15 +126,22 @@ func TestNewProtobufKafkaLogEntry(t *testing.T) {
 	_, turingLogEntry := makeTestTuringResultLogEntry(t)
 	// Overwrite the turing request id value
 	turingLogEntry.TuringReqId = "testID"
+	// Manually remove the key-value pair corresponding to "Content-Type" in the router and enricher header to prevent
```
krithika369, line 129: Could you elaborate more on this non-deterministic behavior? What changes? And why specifically `Content-Encoding` ?

engines/router/missionctl/handlers/logutils.go
```diff
@@ -44,16 +45,21 @@ func logTuringRouterRequestSummary(
...
 			uncompressedData, err := uncompressHTTPBody(resp.header, resp.body)
 			if err != nil {
 				logger.Errorf("Error occurred when reading %s response body: %s",
 					resp.key, err.Error())
-				logEntry.AddResponse(resp.key, nil, err.Error())
+				logEntry.AddResponse(resp.key, nil, nil, err.Error())
 			} else {
-				logEntry.AddResponse(resp.key, uncompressedData, "")
+				// Format the response header
+				responseHeader := map[string]string{}
```
krithika369, line 58: Minor: The `NewTuringResultLogEntry` function in the `resultlog` package has the same logic for formatting headers. Can we extract that into a separate function that can also be reused here?

krithika369 (commented):
Thanks for this PR. It is marked as Draft but I took a quick look anyway and left some comments.

The changes themselves look good. The biggest factor to consider is, how they would play with the existing logging set up - both BQ and Kafka. For example, if someone has a router deployed that's logging to BQ and they update the router, is the new field added to each of the components' records and are they written correctly? These are unit-tested but would be nice to carry out some e2e test as well before merging this PR, locally.

engines/router/missionctl/log/resultlog/kafka_test.go
deadlycoconuts: It's essentially when the `Header` field of `turingLogEntry`, which is an (unordered) map of strings, get encoded, the order in which the items in `Header` get serialised is not guaranteed to always be the same, leading to some unit tests failing and not others. Since the default test log entry `Header` has 2 keys, one for `Content-Encoding` and the other for `Content-Type`, I simply removed removed one of them to ensure that the output's always the same. 

Another option is to create another test log entry for this kafka test without using `makeTestTuringResultLogEntry`, or to make `makeTestTuringResultLogEntry` use only 1 item for the `Header` field, which would also simplify the other tests using it.

krithika369: I see.. In this case, what do you think about decoding the message using the proto and then using `assert.JSONEq` (as we do in `TestNewJSONKafkaLogEntry`) to compare?

engines/router/missionctl/handlers/logutils.go
deadlycoconuts: Okay that's done :)

engines/router/missionctl/log/resultlog/kafka_test.go
deadlycoconuts: Oops, it's gone now.

engines/router/missionctl/log/resultlog/kafka_test.go
deadlycoconuts: Hmm that's possible, but I'm not sure if the intention of the `TestNewProtobufKafkaLogEntry` test was to test the byte encoding of the kafka log entry, as opposed to `TestNewJSONKafkaLogEntry` which seems to be testing for the JSON  output.

engines/router/missionctl/log/resultlog/bigquery.go
```diff
@@ -42,22 +42,53 @@ func (e *bqLogEntry) Save() (map[string]bigquery.Value, string, error) {
...
 	// It seems protobq.Marshal will be adding support for map[string]string that would help simplify the
 	// implementation of Save().
+	kvPairs["request"] = bigquery.Value(map[string]interface{}{
+		"header": formatBQLogEntryHeader(e.TuringResultLogEntry.Request.Header),
+		"body":   e.TuringResultLogEntry.Request.Body,
+	})
+	kvPairs["experiment"] = formatBQLogEntryResponse(e.TuringResultLogEntry.Experiment)
+	kvPairs["enricher"] = formatBQLogEntryResponse(e.TuringResultLogEntry.Enricher)
+	kvPairs["router"] = formatBQLogEntryResponse(e.TuringResultLogEntry.Router)
+	kvPairs["ensembler"] = formatBQLogEntryResponse(e.TuringResultLogEntry.Ensembler)
```
deadlycoconuts, line 56: I've added these additional lines after realising that the headers in the response need to be specially handled before being pushed to BigQuery tables. There's a similar step done for the request headers above.

krithika369 (approved):
LGTM! 🎉

deadlycoconuts:
Thanks for the comments! Merging it now! 🚀

PR #192, opened 2022-04-20
Nop Ensembler Config

krithika369:
### Background

The `default_route_id` configuration on the Turing routers has proven to be confusing to some of our users. It also only plays a direct role when the ensembler chosen is of type "none" or "standard" - it acts as the final response and fallback route in these 2 scenarios, respectively. When the custom ensembler (in the form of "docker" or "pyfunc" configuration) is used, this property of the route is passed down to the ensembler but does not play a critical role, since the user is free to set up the ensembling logic in any manner (eg: simply compare the route name instead of the default route property). Thus, in this PR and in the future ones, the concept of the default route will gradually be phased out to the end user (in the SDK / UI).

**Note:** As of this PR, the Default Route configuration still remains but will be overridden by the Final Response Route chosen for Nop Ensemblers.

#### UI
When the user selects "None" ensembler, a dropdown to select the final response route will be displayed.

Screenshot 2022-04-20 at 8 56 06 AM

Screenshot 2022-04-20 at 8 53 21 AM

Screenshot 2022-04-20 at 8 54 18 AM

To facilitate this, the `Router` and `RouterVersion` data used in the Turing components will have to always be parsed and constructed from the raw API response data (to copy the default route as the Nop ensembler's final response route). Similarly, when the data is posted to the API server, the reverse must be done (i.e., copy the Nop ensembler's final response route as the router's default route). This is required because the API does not introduce any changes for the configuration of the final response route and we continue to rely on the router's default route to achieve the intended behavior. (See `ui/src/services/router/TuringRouter.js`.)

Main Changes:
* `ui/src/services/ensembler/NopEnsembler.js` - Add Nop Ensembler type
* `ui/src/services/router/TuringRouter.js`, `ui/src/services/version/RouterVersion.js` - Handle the conversion between the Nop ensembler's final response route and the router's default route.
* `ui/src/router/components/form/components/nop_config/NopConfigFormGroup.js` - Nop Ensembler form panel

#### SDK

The same logic described for the UI above is also implemented on the Python SDK, so the user may provide the Nop ensembler configuration like below. If provided, it will override the default route.
```python
    ensembler = NopRouterEnsemblerConfig(
        final_response_route_id="test-route"
    )
```

Main changes:
* `sdk/turing/router/config/router_ensembler_config.py` - Introduction of `NopRouterEnsemblerConfig` as a child class of `RouterEnsemblerConfig`
* `sdk/turing/router/config/router_config.py` - Handle the conversion between the Nop ensembler's final response route and the router's default route.

More details on the changes are added as comments.


sdk/tests/conftest.py
```diff
@@ -454,8 +453,8 @@ def generic_router_version(
         error="NONE",
         image="test.io/just-a-test/turing-router:0.0.0-build.0",
         routes=[generic_route for _ in range(2)],
-        default_route="http://models.internal/default",
-        default_route_id="control",
+        default_route=generic_route.endpoint,
```
krithika369, line 456: These tests were originally using a default route id that wasn't a part of the routes themselves. Now, the `_verify_default_route_exists` method added to `sdk/turing/router/config/router_config.py` will prevent that, so the route must be valid.

sdk/turing/router/config/router_config.py
```diff
@@ -216,20 +220,40 @@ def to_open_api(self) -> OpenApiModel:
             kwargs['enricher'] = self.enricher.to_open_api()
         if self.ensembler is not None:
             kwargs['ensembler'] = self.ensembler.to_open_api()
+            if kwargs['ensembler'] is None:
+                # Nop ensembler, delete
+                del kwargs['ensembler']
```
krithika369, line 225: The Turing API does not handle an ensembler type "nop" - it must be left unset.

ui/src/router/activity_log/RouterActivityLogView.js
```diff
@@ -32,21 +32,16 @@ export const RouterActivityLogView = ({ projectId, routerId, router }) => {
...
-    [router.status]
-  );
-
   useEffect(() => {
     fetchEventsOnce();
   }, [fetchEventsOnce]);
 
   useEffect(() => {
-    if (routerStatus === Status.PENDING) {
+    if (router.status === Status.PENDING) {
```
krithika369, line 40: Such changes are required here and in other places because, the router / version objects are now the parsed objects rather than the raw response. So, for example, the status property of the router here is not merely a string, but an Enum object ([ref](https://github.com/gojek/turing/blob/main/ui/src/services/router/TuringRouter.js#L61)).

ui/src/router/details/RouterDetailsView.js
```diff
@@ -42,12 +45,19 @@ export const RouterDetailsView = ({ projectId, routerId, ...props }) => {
...
   useEffect(() => {
-    if (router.status === Status.PENDING.toString()) {
+    if (router.status?.toString() === Status.PENDING.toString()) {
       const interval = setInterval(fetchRouterDetails, 5000);
       return () => clearInterval(interval);
     }
   }, [router.status, fetchRouterDetails]);
 
+  useEffect(() => {
+    // Parse router details
```
krithika369, line 55: Every time the router details change, parse the data for use and store in the local state.

sdk/tests/conftest.py
deadlycoconuts: Whoops, thanks a lot for adding this so more upfront checks are performed before sending any requests to the API!

sdk/turing/router/config/router_config.py
```diff
@@ -216,20 +220,40 @@ def to_open_api(self) -> OpenApiModel:
...
 
         return turing.generated.models.RouterConfig(
             environment_name=self.environment_name,
             name=self.name,
             config=turing.generated.models.RouterVersionConfig(
                 routes=[route.to_open_api() for route in self.routes],
-                default_route_id=self.default_route_id,
```
deadlycoconuts, line 225: Might be asking something obvious, but why is the `default_route_id` removed from the `RouterVersionConfig`'s init parameters? I was wondering if this would cause any issue with the autogenerated class `RouterVersionConfig` class and on Turing API's end when the OpenAPI specs seem to specify that field as required: https://github.com/krithika369/turing/blob/f8b443975a61d8abaed145970ea2f08aa8d37279/api/api/specs/routers.yaml#L608

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -1,37 +1,57 @@
...
+        return None
+
 
 @dataclass
 class RouterEnsemblerConfig:
     """
     Class to create a new RouterEnsemblerConfig
 
-    :param type: type of the ensembler; must be one of {'standard', 'docker'}
+    :param type: type of the ensembler; must be one of {'nop', 'standard', 'docker', 'pyfunc'}
```
deadlycoconuts, line 32: Thanks for spotting that `pyfunc` had been left out!

sdk/turing/router/config/router_config.py
deadlycoconuts: Oops my bad; I totally forgot you did this here: https://github.com/gojek/turing/pull/192/files/f8b443975a61d8abaed145970ea2f08aa8d37279#diff-49fa6d9d67eac9ae9ac62b762e4db2c2b72de2d407f4a5aa1930e94a0f187a97R213

sdk/README.md
```diff
@@ -19,14 +19,14 @@ Check out [samples](./samples) for examples on how to use Turing SDK.
 
 #### Prerequisites
 
-* Python >= 3.8
+* Python 3.7.*
```
deadlycoconuts, line 22: Thanks for catching this! (as well as the other non-updated makefile commands)

sdk/tests/router/config/router_config_test.py
```diff
@@ -18,3 +18,17 @@ def test_set_router_config_with_invalid_routes(actual, new_routes, expected, req
...
+
+@pytest.mark.parametrize(
+    "actual,default_route_id,expected", [
+        pytest.param(
+            "generic_router_config",
+            "test-route-not-exists",
+            InvalidRouteException
+        )
+    ])
+def test_set_router_config_with_invalid_default_route(actual, default_route_id, expected, request):
```
deadlycoconuts, line 30: Extreme nit: can we rename `default_route_id` to something along the lines of `invalid_route_id` or `not_the_default_route_id` to make it a little clearer that the route in that variable is actually invalid?

sdk/turing/router/config/router_config.py
terryyylim: Would it be clearer to add this comment into the code file itself too so that we don't need to dig PRs to figure out why the value needs to be deleted?

sdk/turing/router/config/router_config.py
```diff
@@ -216,20 +220,40 @@ def to_open_api(self) -> OpenApiModel:
...
+            self.ensembler.nop_config.final_response_route_id is not None):
+            default_route_id = self.ensembler.nop_config.final_response_route_id
+        self._verify_default_route_exists(default_route_id)
+        return default_route_id
+
+    def _verify_default_route_exists(self, default_route_id):
+        for route in self.routes:
+            if route.id == default_route_id:
+                return
+        raise turing.router.config.route.InvalidRouteException(
```
terryyylim, line 253: Since we're already importing `Route` via `from turing.router.config.route import Route`, perhaps we could just import `InvalidRouteException` and use it here too?

sdk/turing/router/config/router_config.py
```diff
@@ -216,20 +220,40 @@ def to_open_api(self) -> OpenApiModel:
...
+        default_route_id = self.default_route_id
+        # If nop config is set, use the final_response_route_id as the default
+        if (self.ensembler.type == "nop" and
+            self.ensembler.nop_config is not None and
+            self.ensembler.nop_config.final_response_route_id is not None):
+            default_route_id = self.ensembler.nop_config.final_response_route_id
+        self._verify_default_route_exists(default_route_id)
+        return default_route_id
+
+    def _verify_default_route_exists(self, default_route_id):
```
terryyylim, line 249: nit: Missing typing for `default_route_id` parameter

ui/src/router/components/configuration/components/EnricherConfigSection.js
```diff
@@ -3,7 +3,7 @@ import { EuiPanel } from "@elastic/eui";
 import { DockerConfigViewGroup } from "./docker_config_section/DockerConfigViewGroup";
 
 export const EnricherConfigSection = ({ config: { enricher } }) => {
-  return !enricher ? (
+  return !enricher || enricher.type === "nop" ? (
```
terryyylim, line 6: Out of curiosity, is this something that should have been done previously or?

terryyylim (approved):

ui/src/router/components/configuration/components/EnricherConfigSection.js
krithika369: Good question - it's needed with the new changes for constructing the Turing router object from the API response, before it is used in the details view. We have [this](https://github.com/gojek/turing/blob/main/ui/src/services/router/TuringRouter.js#L29) to capture the default values for the Docker config form, in the enricher definition. So the enricher config is not empty and we need to explicitly check its type now.

sdk/turing/router/config/router_config.py
krithika369: Thanks, fixed.

sdk/turing/router/config/router_config.py
krithika369: Done.

krithika369:
Thanks for the review and the comments, all! I'll be merging this PR once the pipeline succeeds.

PR #193, opened 2022-04-20
Minor fixes for experiment engine configs in the helm chart

krithika369:
This PR makes the following 2 minor fixes to the helm chart:
* `infra/charts/turing/values.yaml` - The default value for the `experimentEngines` should be a map rather than a list, as it is processed so. Having `{}` as the default would cause the following warnings during a helm install:
```
coalesce.go:200: warning: cannot overwrite table with non table for experimentEngines (map[])
```
* `infra/charts/turing/templates/_helpers.tpl` - When an experiment engine is registered but the `options` are unset (either the key does not exist or the value is an empty map `{}`), the `{{ toYaml $expEngine.options | indent 4 }}` causes problems with the indentation of other lines:
```
Error: 'error converting YAML to JSON: yaml: line 32: mapping values are not allowed
      in this context'
```
This is simply handled by adding a check for empty options. 

The chart version has been bumped up.

### DIFF
--- a/infra/charts/turing/Chart.yaml
+++ b/infra/charts/turing/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.2.7
+version: 0.2.8
 appVersion: v1.0.0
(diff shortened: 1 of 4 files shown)


romanwozniak (approved):
Thanks! lgtm

PR #195, opened 2022-04-26
Increased timeouts for Knative deployments and merge main into upgrade-knative branch to keep in sync

ashwinath:
I have increased the timeouts for Knative deployments from 30s to 600s as it is easier to debug clusters more easily, as well as handling slow Kubernetes clusters.

I have also merged main into this branch so that it is easier to synchronise and merge into main easily in the future.


krithika369:
> I have also merged main into this branch

I assume most of the changes in this PR are harmless. Any pointers on what to review? Were there any merge conflicts or any other parts of the code that needs attention?

api/turing/cluster/knative_service.go
```diff
@@ -23,8 +23,9 @@ var knativeSvcDefaults = struct {
 	// to requests
 	RequestTimeoutSeconds int
 }{
-	AutoscalingClass:      "kpa.autoscaling.knative.dev",
-	RequestTimeoutSeconds: 30,
+	AutoscalingClass: "kpa.autoscaling.knative.dev",
+	// TODO: was 30s
+	RequestTimeoutSeconds: 600,
```
krithika369, line 28: As opposed to doing this, can we increase the configurable value [here](https://github.com/gojek/turing/blob/main/api/turing/config/config.go#L206) which controls the overall deployment timeout? Then, we can simply increase it as a config override for [e2e tests](https://github.com/gojek/turing/blob/main/infra/e2e/turing.values.yaml).

krithika369: That being said, increasing the timeout has the potential to make the e2e tests themselves slow (where we are testing a bad config). So I'm wary of increasing this permanently unless there is a strong need.

ashwinath: Yes, let me do some more investigation before we determine this number. But I'll pull this into a configuration.

ashwinath:
> > I have also merged main into this branch
> 
> I assume most of the changes in this PR are harmless. Any pointers on what to review? Were there any merge conflicts or any other parts of the code that needs attention?

Yes, anything that involved the cluster controller have been part of a merge conflict due to the adding of context parameter.

api/turing/cluster/knative_service.go
krithika369: > But I'll pull this into a configuration.

Cool. Just to clarify - if we are intending to tweak  the Knative `RequestTimeoutSeconds` here, we may need to make a new config for this. Also, in this case, it would be important to determine and note down why this needs to be modified to begin with (i.e., why tweaking the overall deployment timeout is not sufficient).

If the existing Deployment timeout config can achieve this, that's preferable.

api/turing/service/router_deployment_service.go
```diff
@@ -480,8 +482,23 @@ func deleteK8sService(
 	controller cluster.Controller,
 	service *cluster.KubernetesService,
 	timeout time.Duration,
+	isCleanUp bool,
 ) error {
-	return controller.DeleteKubernetesService(context.Background(), service.Name, service.Namespace, timeout)
+	ctx := context.Background()
+	var err error
+	if isCleanUp {
```
krithika369, line 489: **Nit:** This "if-else" block can be re-written concisely as:

```golang
controller.DeleteKubernetesDeployment(ctx, service.Name, service.Namespace, timeout, isCleanUp)
```

Same comment for the block at the end of this method too. It can be:
```golang
return controller.DeleteKubernetesService(ctx, service.Name, service.Namespace, timeout, isCleanup)
```

There are a couple other such instances - something that was overlooked in the original MR. Anyway, I'll include this minor change in my next MR to Turing.

api/turing/cluster/knative_service.go
krithika369: The rest of the changes in this PR LGTM.

krithika369:
Closing this in favor of https://github.com/gojek/turing/pull/203 with more recent updates.

PR #196, opened 2022-04-27
Bugfix: Turing API should process experiment engine passkey only if client selection enabled

krithika369:
The Turing API does special handling of the experiment engine passkey (namely, copying existing passkey if unset and encrypting it) when a router is created / edited. While the concept of the passkey is only applicable to those experiment managers that allow client selection, the code was previously attempting to process the passkey for all _standard_ experiment engines. This would lead to a 500 error: `Passkey must be configured` when attempting to make a create / update router request.

The fix corrects the check:
* `api/turing/service/experiment_service.go` - Add `IsClientSelectionEnabled` as a convenience method, for the request module to use
* `api/turing/api/request/request.go` - Call experiment service's `IsClientSelectionEnabled` instead of `IsStandardExperimentManager`
* `api/turing/api/deployment_controller.go` - Call decrypt on the passkey based on `IsClientSelectionEnabled` instead of `IsStandardExperimentManager`

In addition, some small changes are made to the hardcoded experiment engine RPC plugin example, to complete its configuration from the API.
* `engines/experiment/examples/plugins/hardcoded/manager.go` - Implement `ListVariablesForExperiments` so the variables expected by the runner ([here](https://github.com/gojek/turing/blob/main/engines/experiment/examples/plugins/hardcoded/manager.go#L52) and [here](https://github.com/gojek/turing/blob/main/engines/experiment/examples/plugins/hardcoded/runner.go#L49)) can be configured on the router.
* `engines/experiment/examples/plugins/hardcoded/configs/plugin_config_example.yaml` - Update the example plugin config to include the unit variable.
* `engines/experiment/docs/rpc_plugins.md` - Update doc to show the example plugin being configured from the UI

Screenshot 2022-04-27 at 8 14 33 AM


api/e2e/test/helpers_api_test.go
```diff
@@ -41,7 +41,7 @@ func getRouter(
 
 	err = json.Unmarshal(respBytes, &router)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Could not unmarshal: %v\n %s", err, string(respBytes))
```
krithika369, line 44: Added this for debugging e2e test failure, but I think it's ok to keep it.

Screenshot 2022-04-27 at 11 27 43 AM

api/turing/service/experiment_service_test.go
```diff
@@ -8,18 +8,141 @@ import (
...
+		engineInfoErr string
+		expectedErr   string
+		expected      bool
+	}{
+		"standard | no client selection": {
+			engineInfo: manager.Engine{
+				Name: "standard-engine-1",
+				Type: manager.StandardExperimentManagerType,
+				StandardExperimentManagerConfig: &manager.StandardExperimentManagerConfig{
+					ClientSelectionEnabled: false,
```
terryyylim, line 88: nit: Any value in adding these testcases too?
- `&manager.StandardExperimentManagerConfig{}` for `engineInfo.StandardExperimentManagerConfig.ClientSelectionEnabled` condition
- `StandardExperimentManagerConfig: nil` for `engineInfo.StandardExperimentManagerConfig != nil` condition

terryyylim (approved):
Just a nit comment, overall LGTM! 🚀

api/turing/service/experiment_service_test.go
krithika369: * `&manager.StandardExperimentManagerConfig{}` is the same as `&manager.StandardExperimentManagerConfig{ClientSelectionEnabled: false}` - so I don't see any use for this.
* `StandardExperimentManagerConfig: nil` - I initially decided not to bother about this because it's only precautionary (`NewExperimentsService` will ensure that a `nil` value is not possible for this field if the type if determined to be standard). But I guess it's good to cover nonetheless - I'll add it in. Thanks for the feedback!

PR #197, opened 2022-05-04
Add Fallback Response Route Config for Standard Ensemblers

krithika369:
This MR adds similar changes to the UI and SDK as https://github.com/gojek/turing/pull/192, for standard ensemblers, to allow the user to select one of the configured routes as a fallback route. The fallback route will be activated in the following scenarios:
* Call to the experiment engine failed
* Experiment engine returned an experiment and/or treatment that is not configured in the treatment <> route mapping on the standard ensembler

In a scenario where the call to the chosen route fails, the fallback route will not be activated and the error response will be returned.

Under the hood, the fallback route still relied on the "default route" configuration in the Turing API and router.

Screenshot 2022-05-03 at 9 19 00 AM

 ### Modifications

#### Router
* `engines/router/missionctl/fiberapi/routing_strategy.go` - Experiment engine failure should use fallback route (used to return an error response). Selected route failure should be returned as is (used to choose default route).

#### UI
* `ui/src/services/router/TuringRouter.js`, `ui/src/services/version/RouterVersion.js` - Handle standard ensembler's fallback route property translation with the API
* `ui/src/router/components/form/components/ensembler_config/standard_ensembler/StandardEnsemblerFormGroup.js` - Form changes for configuring fallback route
* `ui/src/router/components/configuration/components/EnsemblerConfigSection.js` - View changes for fallback route
* `ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js` - Common component for configuring Nop final response route / standard ensembler fallback route
* `ui/src/router/components/configuration/components/route_config_section/RouteConfigSection.js` - Common component for displaying Nop final response route / standard ensembler fallback route

#### SDK
* `sdk/turing/router/config/router_ensembler_config.py` - Add wrapper data class `EnsemblerStandardConfig` to encapsulate the fallback response route info, along with other standard config expected by the API
* `sdk/turing/router/config/router_config.py` - Configure Nop/Standard ensembler's final response / fallback properties via the `ensembler` property setter rather than on init (to handle update correctly).

Updated the docs and unit tests where appropriate.


docs/how-to/create-a-router/configure-ensembler.md
```diff
@@ -13,6 +13,14 @@ The router will return a response from the route configured to act as the final
 
 ![](../../.gitbook/assets/nop_ensembler_config.png)
 
+## Standard Ensembler
+
+For experiment engines configured to work with standard ensemblers (i.e., Standard Experiment Engines with experiment selection enabled), the router will return a response from from one of the routes based on the configured mapping between routes and experiment treatments. At run time, the treatment returned by the engine will be used to select the corresponding route’s response.
```
terryyylim, line 18: typo: `from one of the routes` (extra from)

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -23,6 +23,33 @@ def final_response_route_id(self, final_response_route_id: str):
     def to_open_api(self) -> OpenApiModel:
         return None
 
+@dataclass
+class EnsemblerStandardConfig:
```
deadlycoconuts, line 27: Maybe I'm not catching the motivation here, but what's the reason behind creating intermediary classes like `EnsemblerStandardConfig` and `EnsemblerNopConfig`?

docs/how-to/create-a-router/configure-ensembler.md
```diff
@@ -13,6 +13,14 @@ The router will return a response from the route configured to act as the final
...
+
+In addition, a fallback route may be configured whose results will be used at runtime when the call to the experiment engine fails or if there does not exist a route mapping for the treatment generated by the experiment engine.
```
terryyylim, line 20: nit: `or if a route mapping for the treatment generated by the experiment engine does not exist`

engines/router/missionctl/fiberapi/routing_strategy.go
```diff
@@ -64,23 +64,20 @@ func (r *DefaultTuringRoutingStrategy) SelectRoute(
 		close(expResultCh)
 	}
 
-	// If error, return it
+	// If error, log it and return the fallback(s)
 	if expErr != nil {
 		log.WithContext(ctx).Errorf(expErr.Error())
-		return nil, fallbacks, createFiberError(expErr)
+		return nil, fallbacks, nil
```
terryyylim, line 70: Considering this change, I don't think we're returning any errors anymore? So perhaps the 3rd response object could be removed.

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
```diff
@@ -0,0 +1,70 @@
...
+      isInvalid={props.isInvalid}
+      options={routeOptions}
+      value={value || ""}
+      onChange={onChange}
+      isDisabled={routeOptions.length <= 1}
+    />
+  );
+};
+
+export const RouteSelectionPanel = ({
```
deadlycoconuts, line 35: Thanks for refactoring this by moving this to a separate file!

terryyylim (approved):

deadlycoconuts (commented):
Looks good to me overall!

docs/how-to/create-a-router/configure-ensembler.md
krithika369: My way sounds fancier. :P But I'll take your suggestion.

engines/router/missionctl/fiberapi/routing_strategy.go
krithika369: Good question. This method implements an interface exposed by Fiber ([here](https://github.com/gojek/fiber/blob/4f0f8284dc84cdab289715eecd70b88213e5133c/routing_strategy.go#L11)). We could certainly consider removing it but we have to examine the usage more closely before making that change (i.e., even if it has 0 benefits within the Turing router, should Fiber, as a standalone library, stop supporting it?). I think we will cover this when we get to the error handling improvements in Turing.

sdk/tests/router/config/router_ensembler_config_test.py
```diff
@@ -454,8 +467,8 @@ def test_create_nop_router_ensembler_config(
     nop_config,
     expected):
     ensembler = NopRouterEnsemblerConfig(final_response_route_id=final_response_route_id)
-    assert ensembler.nop_config == nop_config
     assert ensembler.to_open_api() == expected
+    assert ensembler.nop_config == nop_config
```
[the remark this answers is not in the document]
krithika369: Haha it's actually less for consistency. Now, the `nop_config` is only populated when the `to_open_api()` is run, so we have to check it after that. But I guess this behavior is consistent with the previous implementations (eg: `PyfuncRouterEnsemblerConfig`).

Earlier, I was also populating the `nop_config` property on the base class `RouterEnsemblerConfig` on init of the child class `NopRouterEnsemblerConfig` but figured it doesn't help much (i.e., you can keep changing the properties exposed by `NopRouterEnsemblerConfig` and it wont be reflected into the `nop_config` unless `to_open_api` is run anyway).

sdk/turing/router/config/router_ensembler_config.py
krithika369: I'm happy you asked this. It's a bit of a question I was having myself - whether these classes were really needed. I have come to the conclusion we can do away with them, with some refactoring.

Why they exist today is, when the router config is parsed out of the API response (such as Get Router), we initialise a `RouterEnsemblerConfig` object as opposed to the child class objects. With the base class object, you need to rely on the `.nop_config` or `.standard_config` to get all the properties, which is what these intermediate classes achieve (the autogenerated Turing classes do not have the `final_response_route_id` or `fallback_response_route_id` properties and these are initialised on the intermediate data classes).

Ideally, we can init the proper child classes using some creator factory in which case the child classes alone will suffice to hold these new properties as well as encompass the autogenerated Turing classes for the standard config, pyfunc config, etc. I haven't dug into this yet but will either include it in my next PR which closes the loop on this feature or add it to the backlog.

Does it make sense?

deadlycoconuts: Yeah that makes perfect sense; the thing about the router config needing to be parsed out of the API response and having to handle the `final_response_route_id` or `fallback_response_route_id` properties completely slipped my mind.

>Ideally, we can init the proper child classes using some creator factory in which case the child classes alone will suffice to hold these new properties as well as encompass the autogenerated Turing classes for the standard config, pyfunc config, etc.

Yeah this would be perfect since it'll remove the potentially confusing naming of all the ensembler config classes (that are as of present only distinguishable from each other by the order in which words are being stuck together 😅); I kinda left the SDK in an awkward spot whereby the users get to create ensembler configs using the child classes when building a router config from scratch, but are forced to only manage the parent `RouterEnsemblerConfig` object when the router config conversely parsed out of the API response. But yeah this approach will definitely take a longer time to design/implement so it's definitely out of the scope of this PR. 

Thanks a lot for the current implementation and for the clarification!

deadlycoconuts (approved):

krithika369:
Thanks, all, for the review and the comments. I'll be merging this shortly.

PR #198, opened 2022-05-17
Remove the default route configuration

krithika369:
This is a follow up from https://github.com/gojek/turing/pull/192 and https://github.com/gojek/turing/pull/197 and this PR removes the concept of the default route from the Turing UI and deprecates the `default_route_id` property in the SDK. The API and Router now treat the `default_route_id` property as optional (the API will require that the value is set when using Standard / Nop ensemblers only). The changes are described in detail below.

### API
* `api/api/specs/routers.yaml` - Make `default_route_id` in router config optional
* `api/e2e/**` - Update e2e to stop setting the `default_route_id` in the test cases as they are no longer required for docker ensemblers.
* `api/turing/cluster/servicebuilder/router.go` - Apply default route id to the router only if it is set in the DB
* `api/turing/validation/validator.go` - Check that default route is set only if ensembler type is Standard or Nop and not set otherwise.

### Router
* `engines/router/missionctl/fiberapi/routing_policies.go` - The `routeSelectionPolicy` is used in the Custom ensembler's [EnsemblingFanIn](https://github.com/gojek/turing/blob/main/engines/router/missionctl/fiberapi/fan_in.go#L27) implementation as well as the [DefaultTuringRoutingStrategy](https://github.com/gojek/turing/blob/main/engines/router/missionctl/fiberapi/routing_strategy.go#L17) used in the Nop / Standard ensembler. Now that the Default Route config is not always required, it is checked in the parent struct.
* `engines/router/missionctl/fiberapi/routing_strategy.go` - Check that default route property is set

### SDK
* `sdk/turing/router/config/router_config.py` - Add deprecation warning for the `default_route_id` property. Process it only if set.
* Update unit tests and samples.

### UI
The main changes in the workflow here are:
* There is no "default route" selection in the routes config anymore
* Traffic rules can be configured on all routes
* Final Response route / Fallback response route cannot be a route that has any traffic rule on it (so, in essence, the check is flipped from _"default route cannot have traffic rule on it"_ to _"route with traffic rule cannot be a final/fallback route"_). This in line with the sequence of steps (routes and traffic rules are configured first).

Other details on the UI changes are captured inline.

Screenshot 2022-05-17 at 2 35 18 PM


api/api/specs/routers.yaml
```diff
@@ -512,8 +512,6 @@ components:
           type: "array"
           items:
             $ref: "#/components/schemas/Route"
-        default_route:
```
krithika369, line 515: I don't think this was ever meant to be there, it should have only been `default_route_id`. Removing it.

api/turing/validation/validator_test.go
```diff
@@ -212,27 +213,29 @@ func (tt routerConfigTestCase) RouterConfig() *request.RouterConfig {
...
 			ResultLoggerType: "nop",
 		},
+		Ensembler: tt.ensembler,
 	}
 }
 
 func TestValidateTrafficRules(t *testing.T) {
+	// Common variables used by suite tests
+	routeAID, routeBID := "route-a", "route-b"
+	routeA, routeB := &models.Route{
```
krithika369, line 223: Minor clean up of the hardcoded route info in the tests, to use the variables here.

engines/router/configs/custom_ensembling_router.yaml
```diff
@@ -3,7 +3,6 @@ id: combiner
 fan_in:
   type: fiber.EnsemblingFanIn
   properties:
-    default_route_id: route_id_1
```
krithika369, line 6: No harm having this in the router example but this value will not be set by the API anymore, for custom ensemblers.

ui/src/router/components/configuration/components/router_config_section/RoutesConfigTable.js
```diff
@@ -59,6 +59,7 @@ export const RoutesConfigTable = ({ routes, rules = [], defaultRouteId }) => {
       return acc;
     }, {});
 
+    // Keep the default route (used in Nop / Std ensemblers), if any, on top
```
krithika369, line 62: The previous blue cell highlight for the default route (shown below) is now removed, but I left the default route to continue to be on sorted on the top.

Screenshot 2022-05-17 at 2 32 55 PM

ui/src/router/components/form/components/RouteDropDownOption.js
```diff
@@ -1,18 +1,22 @@
...
   EuiFlexGroup,
   EuiFlexItem,
-  EuiText,
   EuiTextColor,
+  EuiText,
   EuiToolTip,
 } from "@elastic/eui";
 
-export const RuleCardRouteDropDownOption = ({ id, endpoint, isDefault }) => {
+export const RouteDropDownOption = ({
```
krithika369, line 10: This has been refactored to also let the Fallback Route / Final Response Route panel use the same format for the options (the badge + url) as shown in the PR description.

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
```diff
@@ -1,48 +1,78 @@
...
+import { EuiFlexItem, EuiForm, EuiFormRow, EuiSuperSelect } from "@elastic/eui";
 
+import { RouteDropDownOption } from "../RouteDropDownOption";
 import { Panel } from "../Panel";
-import { EuiComboBoxSelect } from "../../../../../components/form/combo_box/EuiComboBoxSelect";
 
 const SelectRouteIdCombobox = ({ value, routeOptions, onChange, ...props }) => {
-  // Note: routeNames changes on every render, which we need to capture name changes to routes
-  const routeNames = routeOptions.map((e) => (!!e.value ? e.value : e.label));
+  const noneValue = "_none_";
```
krithika369, line 8: As captured in the inline code comments below, this special value (can be anything) is needed as a default, to select a placeholder-like option since the `EuiSuperSelect` doesn't have a placeholder property yet. The same mechanism is used in the traffic rules panel.

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
```diff
@@ -1,48 +1,78 @@
...
+      value: noneValue,
+      inputDisplay: props.placeholder,
+      disabled: true,
+    },
+    ...routeOptions,
+  ];
 
   return (
-    <EuiComboBoxSelect
+    <EuiSuperSelect
```
krithika369, line 35: Changing to `EuiSuperSelect` to have a richer dropdown, similar to the traffic rules panel.

ui/src/router/components/form/components/router_config/RoutesPanel.js
```diff
@@ -34,27 +19,18 @@ export const RoutesPanel = ({
...
         {routes.map((route, idx) => (
           <EuiFlexItem key={`route-${idx}`}>
             <RouteCard
               route={route}
-              isDefault={idx === defaultRouteIndex}
               endpointOptions={endpoints}
               onChange={onChange(`routes.${idx}`)}
-              onSelect={() => setDefaultRouteIndex(idx)}
-              onDelete={onDeleteRoute(idx)}
+              onDelete={routes.length > 1 ? onDeleteRoute(idx) : null}
```
krithika369, line 33: Previously, we weren't allowing the default route to be deleted. Now that this concept is removed, we only try to ensure there is always at least one route.

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -34,7 +28,7 @@ export const RulesPanel = ({
   };
 
   const addRuleButton = (
-    <EuiButton fullWidth onClick={onAddRule} isDisabled={routes.length < 2}>
```
krithika369, line 37: Same change as above - previously, we always had a default route and no traffic rule could be created on it.

ui/src/router/components/form/validation/schema.js
```diff
@@ -64,6 +64,14 @@ const routeSchema = yup.object().shape({
   timeout: timeoutSchema.required("Timeout is required"),
 });
 
+const validRouteSchema = yup
+  .mixed()
+  .test("valid-route", "Valid route is required", function (value) {
+    const configSchema = this.from.slice(-1).pop();
```
krithika369, line 70: The last item in the validation context has the entire router config (hierarchically), so using that to define this schema. This way, the schema can be used from multiple steps (fallback/final response validation and traffic rules).

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -34,7 +28,7 @@ export const RulesPanel = ({
...
+    <EuiButton fullWidth onClick={onAddRule} isDisabled={!routes.length}>
```
deadlycoconuts, line 31: Tiny nit: since `routes.length == 0` can never happen (due to the changes you've made [here](https://github.com/gojek/turing/pull/198/files?short_path=4f18af2#r874426063)), should we just remove `isDisabled` since there no way the 'Add Rule' button will ever be disabled?

ui/src/router/components/form/validation/schema.js
deadlycoconuts: Thanks for refactoring this!

deadlycoconuts (approved):
Thanks a lot for the changes to the UI/API/SDK! Looks good to me except for one tiny comment!

terryyylim (approved):
Overall LGTM, very neat PR, thank you @krithika369 !

ui/src/router/components/form/components/router_config/RulesPanel.js
krithika369: Thanks for catching - at first, I didn't enforce the presence of at least one route from the routes panel and I had modified this condition (since all routes could be deleted). I introduced that check later but failed to update this. But thinking about it now, it makes sense to keep the old check here (have at least 2 routes to be able to create a rule) - so I put it back.

krithika369:
Thanks for the reviews! I've addressed the comments. I'll merge this in once the CI succeeds.

PR #200, opened 2022-05-27
Bugfix: Clear default route Id from custom ensembler configs

krithika369:
This MR fixes a few minor bugs on the UI / SDK.

### UI

* `ui/src/services/router/TuringRouter.js` - When a router with No Ensembler / Standard Ensembler is edited and updated to use Custom Ensembler (docker / pyfunc), the `default_route_id` property from the previous configs should be dropped before calling the API. (We could drop this as soon as the type of the Ensembler is changed from the UI but we don't do this for other components too, because with that approach, if the user switched back to the old type in the same edit session, we would have lost the values.)
* `ui/src/services/ensembler/PyFuncEnsembler.js` - Fix regression around missing container config display for Pyfunc ensemblers. Since the detail views now use the parsed data, including the `docker_config` when parsing.

### SDK

Under the hood, the Standard Ensembler's `fallback_response_route_id` and the No Ensembler's `final_response_route_id` both map to the `default_route_id` which will eventually be removed from the SDK. While it is being supported, we were previously not accounting for the idea that the user could directly update the `default_route_id` when Nop / Standard Ensemblers are in use (i.e., we were copying the ensembler configs to `default_route_id` on `to_open_api()` but would have lost the direct changes to the `default_route_id`). This is now overcome by letting the setter for `default_route_id` also update the ensembler configs. Then, upon `to_open_api()`, we only need to look at the ensembler's configs.

To facilitate this fix and to handle potential inconsistencies in the presence of `RouterEnsemblerConfig` vs one of the child classes, the setter for the ensembler now forces the configs to be of the child class types. This is the first step towards addressing this: https://github.com/gojek/turing/pull/197#discussion_r864489973

### DIFF
--- a/sdk/turing/router/config/router_config.py
+++ b/sdk/turing/router/config/router_config.py
@@ -11,7 +11,13 @@ from turing.router.config.traffic_rule import TrafficRule
 from turing.router.config.resource_request import ResourceRequest
 from turing.router.config.log_config import LogConfig, ResultLoggerType
 from turing.router.config.enricher import Enricher
-from turing.router.config.router_ensembler_config import RouterEnsemblerConfig, NopRouterEnsemblerConfig, StandardRouterEnsemblerConfig
+from turing.router.config.router_ensembler_config import (
+    DockerRouterEnsemblerConfig,
+    NopRouterEnsemblerConfig,
+    PyfuncRouterEnsemblerConfig,
+    RouterEnsemblerConfig,
+    StandardRouterEnsemblerConfig,
+)
 from turing.router.config.experiment_config import ExperimentConfig
 
 
@@ -136,6 +142,13 @@ class RouterConfig:
         details="Please use the ensembler properties to configure the final / fallback route.")
     def default_route_id(self, default_route_id: str):
         self._default_route_id = default_route_id
+        # User may directly modify the default_route_id property while it is deprecated.
+        # So, copy to the nop / standard ensembler if set.
+        if hasattr(self, "ensembler"):
+            if isinstance(self.ensembler, NopRouterEnsemblerConfig):
+                self.ensembler.final_response_route_id = default_route_id
+            elif isinstance(self.ensembler, StandardRouterEnsemblerConfig):
+                self.ensembler.fallback_response_route_id = default_route_id
 
     @property
     def experiment_engine(self) -> ExperimentConfig:
@@ -206,8 +219,6 @@ class RouterConfig:
         if ensembler is None:
             # Init nop ensembler config if ensembler is not set
             self._ensembler = NopRouterEnsemblerConfig(final_response_route_id=self.default_route_id)
-        elif isinstance(ensembler, RouterEnsemblerConfig):
-            self._ensembler = ensembler
         elif isinstance(ensembler, dict):
             # Set fallback_response_route_id into standard ensembler config
             if ensembler["type"] == "standard" and "fallback_response_route_id" not in ensembler["standard_config"]:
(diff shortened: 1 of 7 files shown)


sdk/turing/router/config/router_config.py
```diff
@@ -250,7 +271,7 @@ def to_open_api(self) -> OpenApiModel:
         )
 
     def _get_default_route_id(self):
-        default_route_id = self.default_route_id
+        default_route_id = None
```
krithika369, line 274: By setting `None` here, we will be clearing this value for routers with custom ensemblers.

sdk/turing/router/config/router_config.py
```diff
@@ -206,16 +219,24 @@ def ensembler(self, ensembler: Union[RouterEnsemblerConfig, Dict]):
...
                 ensembler["standard_config"]["fallback_response_route_id"] = self.default_route_id
             self._ensembler = RouterEnsemblerConfig(**ensembler)
         else:
             self._ensembler = ensembler
-
+        # Init child class types
+        if isinstance(self._ensembler, RouterEnsemblerConfig):
+            if self._ensembler.type == "nop" and not isinstance(self._ensembler, NopRouterEnsemblerConfig):
+                self._ensembler = NopRouterEnsemblerConfig.from_config(self._ensembler.nop_config)
+            if self._ensembler.type == "standard" and not isinstance(self._ensembler, StandardRouterEnsemblerConfig):
```
terryyylim, line 233: Not sure if I'm missing something here, why is it 2 `if` conditions and then followed by `elif` here?

sdk/turing/router/config/router_config.py
```diff
@@ -136,6 +142,13 @@ def default_route_id(self) -> str:
         details="Please use the ensembler properties to configure the final / fallback route.")
     def default_route_id(self, default_route_id: str):
         self._default_route_id = default_route_id
+        # User may directly modify the default_route_id property while it is deprecated.
+        # So, copy to the nop / standard ensembler if set.
+        if hasattr(self, "ensembler"):
```
terryyylim, line 147: Just curious, wouldn't `if self.ensembler` work? 🤔

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -350,6 +363,22 @@ def service_account(self) -> str:
     def service_account(self, service_account: str):
         self._service_account = service_account
 
+    @staticmethod
+    def from_config(config: turing.generated.models.EnsemblerDockerConfig):
```
terryyylim, line 367: CMIIW, this method is not being overwritten anywhere else since this class is never used by another subclass, is there a reason why this is not a `classmethod` (since we're initializing the class through this call)?

sdk/turing/router/config/router_config.py
krithika369: Thanks, I missed updating the second check to an `elif`.

sdk/turing/router/config/router_config.py
krithika369: Not on `__init__` which is why I added this. During init, the default_route_id attribute is created first (followed by the `ensembler` which always reads the value from the `default_route_id` if its own value is not an Ensembler instance.

sdk/turing/router/config/router_ensembler_config.py
krithika369: I chose to use `staticmethod` because we don't need access to the class properties. But, as you're saying, we are in fact initializing the class through the call so it makes more sense for it to be a class method. I'll update this. Thanks!

krithika369:
@terryyylim I fixed another small UI bug and updated the PR with tests. Could you take another look? Thanks!

ui/src/services/router/TuringRouter.js
```diff
@@ -109,19 +109,24 @@ export class TuringRouter {
...
+      delete obj.config["ensembler"];
+    } else if (obj.config.ensembler.type === "standard") {
+      // Copy the fallback response route id to the top level, as the default route
+      obj.config.default_route_id =
+        obj.config["ensembler"].standard_config["fallback_response_route_id"];
+      delete obj.config["ensembler"].standard_config[
+        "fallback_response_route_id"
+      ];
+    } else {
+      // Docker or Pyfunc ensembler, clear the default_route_id
```
terryyylim, line 125: Would there be a need to clear `pyfunc_config` if the Ensembler type is Docker?

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -350,6 +364,22 @@ def service_account(self) -> str:
     def service_account(self, service_account: str):
         self._service_account = service_account
 
+    @classmethod
+    def from_config(cls, config: turing.generated.models.EnsemblerDockerConfig):
```
terryyylim, line 368: nit: Missing return types for `classmethod`s.

ui/src/services/router/TuringRouter.js
krithika369: Nope. Pyfunc ensemblers may also have docker configs (if it has been built), not the other way.

sdk/turing/router/config/router_ensembler_config.py
krithika369: Cool, let fix that.

terryyylim (approved):

PR #202, opened 2022-05-31
Add dynamic loading of Experiment Engine config

terryyylim:
**What this PR does / why we need it:**

This PR is required to dynamically load any global variables from configured Experiment Engine UI via script tag in Turing's index.html file. This can be done through providing the url of the config, similar to how the url of the UI is provided currently.

**Which issue(s) this PR fixes:**

`NONE`

PR #205, opened 2022-06-07
Fix dynamic loading of Experiment Engine config

terryyylim:
A previous related [PR](https://github.com/gojek/turing/pull/202) caused a bug which crashes Routers Details view. This view uses `ExperimentConfigSection` component which requires configured Experiment Engine information retrieved via global variables configured on the Experiment Engine UI. This resulted in runtime configs being NOT loaded into the index.html headers and hence, not available for use.

Logs show the following information:
```
TypeError: Cannot read properties of undefined (reading 'experimentEngineApiUrl')
...
```

This PR addresses the above issue for both `ExperimentConfigSection` and `StandardExperimentConfigGroup` components which load information from the remote Experiment Engine UI.

### DIFF
--- a/ui/src/components/remote_component/DynamicHookComponent.js
+++ b/ui/src/components/remote_component/DynamicHookComponent.js
@@ -1,0 +1,61 @@
+import React, { useEffect, useState } from "react";
+
+import useDynamicScript from "../../hooks/useDynamicScript";
+
+// Renderless component wrapper
+const LoadDynamicScript = ({ url, setReady, setFailed }) => {
+  const { ready, failed } = useDynamicScript({
+    url: url,
+  });
+
+  useEffect(() => {
+    setReady(ready);
+    setFailed(failed);
+  }, [setReady, setFailed, ready, failed]);
+
+  return null;
+};
+
+// Dynamic Script Loading component wrapper
+export const DynamicHookComponent = ({
+  FallbackView,
+  experimentEngine,
+  children,
+}) => {
+  const [urlReady, setUrlReady] = useState(false);
+  const [urlFailed, setUrlFailed] = useState(false);
+  const [configReady, setConfigReady] = useState(false);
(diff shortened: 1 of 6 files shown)


ui/src/router/components/configuration/components/ExperimentConfigSection.js
```diff
@@ -25,6 +27,9 @@ const FallbackView = ({ text }) => (
 );
 
 const CustomExperimentConfigView = ({ projectId, remoteUi, config }) => {
+  const [configReady, setConfigReady] = useState(false);
+  const [configFailed, setConfigFailed] = useState(false);
+
   // Retrieve script from host dynamically
   const { ready, failed } = useDynamicScript({
```
krithika369, line 34: Shall we replace this with `LoadDynamicScript` too instead of directly calling the hook?

Also, since we have similar logic in multiple places now, there is likely some scope for refactoring this further to reduce some code duplication by wrapping all the loading into a single component.

ui/src/components/remote_component/DynamicHookComponent.js
```diff
@@ -0,0 +1,61 @@
...
+  useEffect(() => {
+    setReady(ready);
+    setFailed(failed);
+  }, [setReady, setFailed, ready, failed]);
+
+  return null;
+};
+
+// Dynamic Script Loading component wrapper
+export const DynamicHookComponent = ({
```
krithika369, line 20: The changes are great! My only comment for improvement is, `DynamicHookComponent` is named and placed (under `ui/src/components/remote_component`) as though it were a generic component. It's weird that it worries about the details of `url` vs `config` and talks about "Experiment Engine". Can we move this component to  `ui/src/components/experiments` and rename it to `ExperimentEngineLoaderComponent` or something like that?

krithika369 (approved):
LGTM! Thanks for the fix and the refactoring!

PR #208, opened 2022-06-10
Update postgres dependency in Helm charts

krithika369:
The Helm release action fails with the following error message ([see](https://github.com/gojek/turing/runs/6822344706?check_suite_focus=true)):
```
Error: can't get a valid version for repositories postgresql. Try changing the version constraint in Chart.yaml
```

The chart version (8.9.8) can be listed using:
```
helm search repo bitnami/postgresql --versions
```
and downloaded using:
```
helm pull bitnami/postgresql --version 8.9.8 --untar
```

However, `helm dependency update ./infra/charts/turing` fails with the above message and it can be confirmed that the chart is no longer available at `https://artifacthub.io/packages/helm/bitnami/postgresql/8.9.8`.

As suggested in the error remediation, this PR updates the dependency chart version to another which is the latest one supporting Postgres `11.X`:
```
Chart version 8.9.8 (Postgres 11.8.0) -> Chart version 10.16.2 (Postgres 11.14.0)  
```

Accordingly, some of the default values for Postgres and the corresponding docs have also been updated.


krithika369:
Verified publishing the chart @ https://github.com/krithika369/charts

leonlnj (approved):
LGTM

PR #209, opened 2022-06-10
Update deprecated kube state metrics from the alerts

krithika369:
This PR updates some deprecated kube-state metrics, which are removed in more recent versions (see https://github.com/kubernetes/kube-state-metrics/pull/1004), as below:

```
kube_pod_container_resource_requests_cpu_cores ->  kube_pod_container_resource_requests{resource="cpu"}
```

```
kube_pod_container_resource_requests_memory_bytes ->  kube_pod_container_resource_requests{resource="memory"}
```


leonlnj (approved):
LGTM

PR #210, opened 2022-06-10
Bugfix: Show Configure Alerts button when router is deployed

krithika369:
This PR is a small fix for a regression introduced by https://github.com/gojek/turing/pull/192, which introduced conversion of the API router / router version response to `TuringRouter` and `RouterVersion` objects, respectively, before passing down to the views. Thus, the conversion of the status string to the `Status` object would have also happened ahead of time and there is no need to do `Status.fromValue` within the `RouterAlertDetails` component. This was missed in the original PR.

Screenshot 2022-06-10 at 11 05 20 AM


leonlnj (approved):
LGTM

PR #216, opened 2022-07-19
Use background context rather than cancelled context

terryyylim:
### Bug
Turing's dependency on Kubernetes API was updated in https://github.com/gojek/turing/pull/183 and most Kubernetes APIs now require context to be passed which caused a regression. When [ctx.Done()](https://github.com/gojek/turing/blob/main/api/turing/cluster/controller.go#L721) is called, the context was cancelled and ctx.Err() returns a non-nil error. Thus, when passed to [ListPods API](https://github.com/gojek/turing/blob/main/api/turing/cluster/controller.go#L751), when the ctx.Err() value is first checked, the call exits given a non-nil error and pods that satisfy the provided labels are never returned, causing termination logs to NOT be propagated to the user.

### Fix
The fix uses a new empty context via context.Background(), which will not be cancelled and that allows Kubernetes Go Client ListPods API to retrieve the correct pods based on provided labels. Then, the container error logs from Terminated state can be propagated as event logs to the user.

image

### Enhancements
In addition, this PR standardizes how context is being used for the following workflows:
- Creation/Updation: Use ctx with Deployment timeout as specified in config
- Deletion: Use context.Background() with no timeouts (Some functions in api/turing/cluster/controller.go have their timeout parameter removed)
  - GracePeriodSeconds exposed by Kubernetes Go client's `metav1.DeleteOptions{}` is optional

### DIFF
--- a/api/turing/cluster/controller.go
+++ b/api/turing/cluster/controller.go
@@ -719,7 +719,7 @@ func (c *controller) waitKnativeServiceReady(
 	for {
 		select {
 		case <-ctx.Done():
-			terminationMessage := c.getKnativePodTerminationMessage(ctx, svcName, namespace)
+			terminationMessage := c.getKnativePodTerminationMessage(context.Background(), svcName, namespace)
 			if terminationMessage == "" {
 				// Pod was not created (as with invalid image names), get status messages from the knative service.
 				svc, err := services.Get(ctx, svcName, metav1.GetOptions{})
--- a/api/turing/cluster/controller.go
+++ b/api/turing/cluster/controller.go
@@ -64,16 +64,13 @@ type clusterConfig struct {
 // Controller defines the operations supported by the cluster controller
 type Controller interface {
 	DeployKnativeService(ctx context.Context, svc *KnativeService) error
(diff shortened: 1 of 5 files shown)


krithika369:
1. Could we indicate in the description that this is a regression?
2. Have you also checked the calling code in `api/turing/service/router_deployment_service.go` as I see we are using `context.Background()` in some places and original context with deployment timeout in the others.

terryyylim:
> 1. Could we indicate in the description that this is a regression?

Done.

> 2. Have you also checked the calling code in api/turing/service/router_deployment_service.go as I see we are using context.Background() in some places and original context with deployment timeout in the others.

In functions where original context are being passed to functions, a defer statement has been specified for the context cancel call. In cases where ctx.Done() is called, the context is no longer used and errors are being returned. Unless you mean something else here like standardising how context is being initialised and passed? (i.e along the lines of if original context is not required, we use context.Background() instead of original context)

krithika369 (approved):
Code changes LGTM. Please feel free to merge once the test cases are addressed.

PR #217, opened 2022-07-20
Pass original request headers merged with enricher response headers to Pyfunc ensembler

krithika369:
This PR makes 2 user-facing changes:
* `engines/router/missionctl/handlers/http_handler.go` - Make the Enricher's response headers available to the Ensembler in the request headers. The merging of the headers was introduced in https://github.com/gojek/turing/pull/176, intended to be used for the routing step and we hadn't propagated it down to the Ensembler. However, allowing the Ensembler access to those values would help the Ensembler make decisions based on values computed by the Enricher.
* Pass request headers to the Pyfunc Ensembler. Previously, request headers were not available to Pyfunc ensemblers. The changes are described in detail below.

### Passing Ensembler Request Headers to the Pyfunc Ensembler
To support this, the optional `**kwargs` argument has been added to the `EnsemblerBase.ensemble` method, meant to take in contextual information. This will be used by the Pyfunc Ensembler Service to propagate request headers. The main changes are below.
* `engines/pyfunc-ensembler-service/pyfunc_ensembler_runner/handler.py` - Nest request headers and body under a single dict, when calling the MlFlow Pyfunc model's [predict](https://github.com/mlflow/mlflow/blob/v1.20.0/mlflow/pyfunc/__init__.py#L592).
* `sdk/turing/ensembler.py` - Pick up the headers from the model input and pass them to the `ensemble()` method. The signature of the `ensemble` method now takes in `**kwargs` which should be included in the custom implementation to take advantage of the headers. However, since some users may be on the older Pyfunc ensemblers, a `try-except` has been added to issue the call without headers if there is an associated exception. This special handling can be removed in a few releases, when users have started to include `**kwargs` in their implementation.

### Tests
The following scenarios have been tested to be working with the new Pyfunc service as of this PR:
* Deploying images already built using older SDK versions
* Building and deploying images from artifacts created by older SDK versions
* Building and deploying images from artifacts created by the new SDK version as of this PR, with or without the `**kwargs`

### DIFF
--- a/docs/how-to/create-a-router/configure-ensembler.md
+++ b/docs/how-to/create-a-router/configure-ensembler.md
@@ -26,7 +26,7 @@ It is not possible to select as the fallback response a route that has traffic r
 {% endhint %}
 
 ## Docker
-Turing will deploy specified image as a post-processor and will send the in the request payload the original request, responses from all routes, and the treatment configuration (if a Experiment Engine is selected, in Configure Experiment Engine), for ensembling. The ensembler's request headers will contain the original request headers sent to Turing, merged with the enricher's response headers (if there are duplicates, the value in the enricher's response headers will take precedence). To configure a Docker ensembler, there are 3 sections to be filled.
+Turing will deploy the specified image as a post-processor and will send the in the request payload the original request, responses from all routes, and the treatment configuration (if an Experiment Engine is selected, in Configure Experiment Engine), for ensembling. The ensembler's request headers will contain the original request headers sent to Turing, merged with the enricher's response headers (if there are duplicates, the value in the enricher's response headers will take precedence). To configure a Docker ensembler, there are 3 sections to be filled.
 
 Configure the Docker Container. There are 4 required inputs.
 
--- a/docs/how-to/create-a-router/configure-ensembler.md
+++ b/docs/how-to/create-a-router/configure-ensembler.md
@@ -2,7 +2,7 @@
 
 Turing currently supports ensemblers in the same fashion as the enrichers. The ensembling is controlled by the policy from the rule engine.
 
-Currently, there are 4 options available - no ensembler, a standard ensembler, Docker and an external ensembler.
+Currently, there are 4 options available - no ensembler, a standard ensembler, Docker and Pyfunc ensembler.
 
 ## No Ensembler
 The router will return a response from the route configured to act as the final response. This option is available only when **no experiment engine** is configured in Configure Experiment Engine.
@@ -26,7 +26,7 @@ It is not possible to select as the fallback response a route that has traffic r
 {% endhint %}
 
 ## Docker
-Turing will deploy specified image as a post-processor and will send the original request, responses from all routes, and the treatment configuration (if a Experiment Engine is selected, in Configure Experiment Engine), for ensembling. To configure a Docker ensembler, there are 3 sections to be filled.
+Turing will deploy specified image as a post-processor and will send the in the request payload the original request, responses from all routes, and the treatment configuration (if a Experiment Engine is selected, in Configure Experiment Engine), for ensembling. The ensembler's request headers will contain the original request headers sent to Turing, merged with the enricher's response headers (if there are duplicates, the value in the enricher's response headers will take precedence). To configure a Docker ensembler, there are 3 sections to be filled.
 
 Configure the Docker Container. There are 4 required inputs.
 
--- a/docs/how-to/create-a-router/configure-ensembler.md
+++ b/docs/how-to/create-a-router/configure-ensembler.md
@@ -26,7 +26,7 @@ It is not possible to select as the fallback response a route that has traffic r
 {% endhint %}
 
 ## Docker
-Turing will deploy the specified image as a post-processor and will send the in the request payload the original request, responses from all routes, and the treatment configuration (if an Experiment Engine is selected, in Configure Experiment Engine), for ensembling. The ensembler's request headers will contain the original request headers sent to Turing, merged with the enricher's response headers (if there are duplicates, the value in the enricher's response headers will take precedence). To configure a Docker ensembler, there are 3 sections to be filled.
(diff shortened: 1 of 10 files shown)


engines/pyfunc-ensembler-service/pyfunc_ensembler_runner/ensembler_runner.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, artifact_dir: str):
     def load(self):
         self._ensembler = pyfunc.load_model(self.artifact_dir)
 
-    def predict(self, inputs: Dict[str, Any]) -> List[Any]:
-        logging.info(f"Input request payload: {inputs}")
-        output = self._ensembler.predict(inputs)
+    def predict(self, body: Dict[str, Any], headers: Dict[str, str]) -> List[Any]:
```
krithika369, line 19: This call's MLFlow Pyfunc model's predict which is eventually received [here](https://github.com/gojek/turing/blob/main/sdk/turing/ensembler.py#L62) at runtime. The `self.ensemble` call from  `_ensemble_request` finally calls the implementation provided by the user.

engines/pyfunc-ensembler-service/tests/test_ensembler_runner.py
```diff
@@ -20,19 +20,42 @@
 
 
 @pytest.mark.parametrize(
-    "inputs,expected", [
+    "ensembler_uri,inputs,headers,expected", [
```
krithika369, line 23: Changes to test old and new ensemblers (created with and without `**kwargs`)

engines/pyfunc-ensembler-service/tests/conftest.py
```diff
@@ -6,6 +6,22 @@
 
 class TestEnsembler(PyFunc):
 
+    def initialize(self, artifacts: dict):
```
terryyylim, line 9: Didn't notice this previously, but perhaps we should standardize the typings? I see different options - `dict`, `Dict`, `Dict[with types]`.

sdk/turing/ensembler.py
```diff
@@ -87,18 +92,32 @@ def _ensemble_request(self, model_input: Dict[str, Any]) -> Any:
...
+            return self.ensemble(
+                input=request_body['request'],
+                predictions=routes_to_response,
+                treatment_config=request_body['response']['experiment'],
+                headers=model_input['headers']
+            )
+        except TypeError as e:
+            if "got an unexpected keyword argument 'headers'" in str(e):
+                # This handles the legacy ensemblers
+                # TODO: Deprecate support for legacy ensemblers
```
terryyylim, line 113: Should perhaps specify what future SDK version would we be deprecating the legacy ensemblers.

terryyylim (approved):
LGTM 🚀 , left some small comments!

engines/pyfunc-ensembler-service/tests/test_ensembler_runner.py
```diff
@@ -52,12 +75,12 @@ def get_app(self):
         return PyFuncEnsemblerServer(ensembler).create_application()
 
     def test_valid_request(self):
-        response = self.fetch('/ensemble', method="POST", body=dummy_long_request)
+        response = self.fetch("/ensemble", method="POST", body=dummy_long_request)
```
deadlycoconuts, line 78: Thanks for cleaning this up!

deadlycoconuts (approved):
LGTM! 🚀 Thanks for cleaning up some of the stuff in the SDK files; @terryyylim also raised some interesting points which are worth thinking about so I just "+1"-ed some of them :)

sdk/turing/ensembler.py
```diff
@@ -1,4 +1,6 @@
 import abc
+import deprecation
```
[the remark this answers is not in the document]
krithika369: Thanks for catching, and the suggestion - removed the `deprecation` import and added a warning log.

engines/pyfunc-ensembler-service/pyfunc_ensembler_runner/ensembler_runner.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, artifact_dir: str):
...
+        logging.info(f"Input request payload: {body}")
```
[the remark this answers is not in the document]
krithika369: Right now, the default logger uses INFO level and it's not possible to change that. So I left out the header log, to not bloat the logs further. Will capture an item in the backlog to make logging configurable.

docs/how-to/create-a-router/configure-ensembler.md
```diff
@@ -26,7 +26,7 @@ It is not possible to select as the fallback response a route that has traffic r
 {% endhint %}
 
 ## Docker
-Turing will deploy specified image as a post-processor and will send the original request, responses from all routes, and the treatment configuration (if a Experiment Engine is selected, in Configure Experiment Engine), for ensembling. To configure a Docker ensembler, there are 3 sections to be filled.
+Turing will deploy the specified image as a post-processor and will send the in the request payload the original request, responses from all routes, and the treatment configuration (if an Experiment Engine is selected, in Configure Experiment Engine), for ensembling. The ensembler's request headers will contain the original request headers sent to Turing, merged with the enricher's response headers (if there are duplicates, the value in the enricher's response headers will take precedence). To configure a Docker ensembler, there are 3 sections to be filled.
```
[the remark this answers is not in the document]
krithika369: Thanks! Rephrased for clarity.

engines/pyfunc-ensembler-service/tests/conftest.py
krithika369: Ideally yes. But there are also other inconsistencies like double vs single quote usage. We can put in a linter at some point to catch these everywhere. For now, I just changed this example of the test ensemblers to `Dict` to match the`ensemble` method.

sdk/turing/ensembler.py
krithika369: I think deprecating this depends on users moving away as well. So we need to do it at some point, but I'm not sure when. We will also need to remove some of the [default_route_id](https://github.com/gojek/turing/blob/main/sdk/turing/router/config/router_config.py#L135) related stuff. Will also track this in the backlog.

krithika369:
Thanks for the quick review, both! I'll be merging this shortly if no further comments / questions.

PR #218, opened 2022-07-20
Add SDK e2e tests

deadlycoconuts:
## Context
With the introduction of the Turing SDK, there has yet to be a set of e2e tests implemented to ensure that the SDK is always in sync with the development of the Turing API. This PR thus introduces a suite of e2e tests, based upon the API e2e tests, that utilise the SDK to interact with the API. Additional updates to the CI/CD workflows are also introduced to ensure that this suite completes successfully before allowing any merging/deployment from taking place.

## Modifications
- `.github/actions/setup-test-cluster/action.yaml` - Creation of a new reusable workflow from the chunk of steps found in `.github/workflows/turing.yaml` to create a test cluster; used when setting up a test cluster for the regular API e2e tests and the SDK e2e tests
- `.github/workflows/turing.yaml` - Introduction of a `test-e2e-sdk` job to set up the test cluster and to run the test suite
-  `sdk/Makefile` - Minor redefinition of the makefile to ensure that there are 2 different makefile targets, 1 for running the SDK unit tests and the other for running the SDK e2e tests
- `sdk/e2e/*` - Addition of the e2e test suite for the SDK; based upon the API e2e tests


.github/actions/setup-test-cluster/action.yaml
```diff
@@ -0,0 +1,215 @@
...
+
+    - name: Download Cluster Init Docker tar archive
+      uses: actions/download-artifact@v2
+      with:
+        name: ${{ inputs.cluster_init_tar_archive_name }}
+
+    - name: Run action cluster-init
+      uses: ./.github/actions/run-cluster-init
+      with:
+        cluster_name: ${{ env.CLUSTER_NAME}}
```
krithika369, line 85: Since these are the inputs required by the Set Up Test Cluster action, it would be better to pass these values down from the calling workflow:

```yaml
# .github/workflows/turing.yaml
      - name: Set Up Test Cluster
        uses: ./.github/actions/setup-test-cluster
        with:
          go-version: ${{ env.GO_VERSION }}
          ...
          matrix_values_file: ${{ matrix.valuesFile }}
          cluster_name: ${{ env.CLUSTER_NAME}}
          ...
```

.github/workflows/turing.yaml
```diff
@@ -354,175 +352,137 @@ jobs:
...
-        run: |
-          kubectl apply -f infra/e2e/turing.mockserver.yaml
+          echo "::group::logs deployment/turing"
+          kubectl logs deployment/turing
+          echo "::endgroup::"
+
+  test-e2e-sdk:
+    runs-on: ubuntu-latest
+    env:
+      CLUSTER_INIT_VERSION: ${{ needs.build-cluster-init.outputs.cluster-init-version }}
```
krithika369, line 413: Since these env vars are duplicated with the `test-e2e` job, shall we define them outside all jobs, at the top of the file?

.github/workflows/turing.yaml
```diff
@@ -354,175 +352,137 @@ jobs:
...
+      TURING_ROUTER_VERSION: ${{ needs.build-router.outputs.router-version }}
+      TEST_EXPERIMENT_ENGINE_PLUGIN_VERSION: ${{ needs.build-test-experiment-engine-plugin.outputs.test-experiment-engine-plugin-version }}
+    needs:
+      - build-api
+      - build-router
+      - build-test-experiment-engine-plugin
+      - build-cluster-init
+    strategy:
+      fail-fast: false
+      matrix:
```
krithika369, line 429: This matrix effectively produces 6 configurations. While it's helpful to be exhaustive, since the e2e tests are heavy (~10 mins for each job), it could be wasteful to run all 6 configs with every PR, unless they are expected to produce different results (eg: does `useInClusterConfig` matter for the SDK <> API tests? What about the Python version?). Besides, we also have 2 jobs for the API e2e already.

Given these, do you have any ideas on how we can reduce the test combinations?

deadlycoconuts: Hmm you're right, `useInClusterConfig` and the python versions shouldn't drastically affect the results of the e2e tests. I'll remove the redundant configurations then and leave the sdk e2e tests to solely run on remote cluster credentials and python 3.9. I'm not sure if we're aiming to implement multiple sets of e2e tests for each release of the SDK (but I doubt so) so I'm leaving that out too.

.github/workflows/turing.yaml
deadlycoconuts: Okay I've moved these env vars out to the top of the file so they are now scoped for the entire workflow, except for those that require intermediate jobs to be completed, i.e. env vars that are defined via `${{ needs.[job_name].[var_name] }}`

.github/actions/setup-test-cluster/action.yaml
```diff
@@ -0,0 +1,219 @@
+name: Set Up Test Cluster
```
deadlycoconuts, line 1: Created this reusable workflow since we're using it (setting up a test cluster) for both the regular e2e tests as well as the SDK e2e tests.

sdk/turing/ensembler.py
```diff
@@ -395,7 +395,7 @@ def match_dependency(spec, name):
         # conda dependencies whose spec differs slightly from Python's setuptools.
         # We could install the complete conda library but this is too bulky if the goal is
         # to just carry out this matching.
-        return spec == name or re.match(f'{name}[><=\s]+', spec) is not None
+        return spec == name or re.match(name + r"[><=\s]+", spec) is not None
```
deadlycoconuts, line 398: Edited this to fix the deprecation warning about the use of `\s` as an escape sequence in f strings: https://github.com/gojek/turing/runs/7608305528?check_suite_focus=true#step:6:63

.github/workflows/turing.yaml
```diff
@@ -354,175 +352,128 @@ jobs:
...
+          cluster_init_version: ${{ env.CLUSTER_INIT_VERSION }}
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
         with:
-          path: api/.go/pkg/mod/
-          key: |
-            gomod-${{ hashFiles('api/go.mod') }}
-          restore-keys: gomod-
+          python-version: 3.9
```
deadlycoconuts, line 454: Fixed the python version at 3.9 for the e2e tests

.github/workflows/turing.yaml
```diff
@@ -354,175 +352,128 @@ jobs:
...
-
-      - name: Cache Test Dependencies
-        uses: actions/cache@v2
+          go-version: ${{ env.GO_VERSION }}
+          turing_api_tar_archive_name: turing-api.${{ env.TURING_API_VERSION }}.tar
+          turing_router_tar_archive_name: turing-router.${{ env.TURING_ROUTER_VERSION }}.tar
+          experiment_engine_plugin_archive_name: test-experiment-engine-plugin.${{ env.TEST_EXPERIMENT_ENGINE_PLUGIN_VERSION }}.tar
+          cluster_init_tar_archive_name: cluster-init.${{ env.CLUSTER_INIT_VERSION }}.tar
+          use_in_cluster_config: false
+          values_file: "turing.values.remote.yaml"
```
deadlycoconuts, line 443: Fixed the remote credentials configuration when running the e2e tests

.github/actions/setup-test-cluster/action.yaml
deadlycoconuts: Hmm is there a general rule when deciding when we'll pass these values as inputs from the calling workflow (as shown in your example) or to access these values directly as environment variables? There are quite a number of other values that are accessed in the Set Up Test Cluster action (and other actions as well) as environment variables, e.g. `DOCKER_REPOSITORY`, `TURING_ROUTER_VERSION`, `CLUSTER_INIT_VERSION`, etc.

.github/workflows/turing.yaml
```diff
@@ -339,9 +339,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        name:
-          - "In-cluster credentials"
-          - "Remote cluster credentials"
+        name: ["In-cluster credentials", "Remote cluster credentials"]
```
krithika369, line 342: What do you think about only running the API e2e tests for `In-cluster credentials` and letting the SDK-based e2e for the other one (`Remote cluster credentials`) ? As with writing test case, this would be a way to reduce the space of variable combinations. In this case, I don't see much value in keeping separate API e2e tests for `Remote cluster credentials` when we have the same covered by the SDK tests.

krithika369: Along these lines, the regular e2e tests and the SDK based e2e tests could also be combined into a matrix under the same job using a smarter strategy: https://github.com/vlsi/github-actions-random-matrix

But, at this point, this may be over-optimising and we could consider this later instead, if we see a need for it.

sdk/e2e/02_update_router_invalid_test.py
```diff
@@ -0,0 +1,83 @@
...
+import turing
+
+from turing.router.config.experiment_config import ExperimentConfig
+from turing.router.config.resource_request import ResourceRequest
+from turing.router.config.enricher import Enricher
+from turing.router.config.common.env_var import EnvVar
+from turing.router.config.router_version import RouterStatus
+
+
+def test_update_router_invalid_config():
```
krithika369, line 14: My understanding was that pytest runs the tests in a random order. So I was expecting to see one of the solutions described here, to ensure order of execution: https://stackoverflow.com/questions/17571438/test-case-execution-order-in-pytest

How do we make sure that this test always runs after the first?

krithika369 (approved):
Left some suggestions / questions. Other than that, looks great. Thanks for this comprehensive PR and adding all the API tests through the SDK as well. 🎉

sdk/e2e/02_update_router_invalid_test.py
deadlycoconuts: I believe the tests within a module or a class run in a random order but module discovery seems to be something non-random; pytest appears to recursively look into each directory to collect individual test modules in an alphabetical order (doesn't say it explicitly[ here](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#conventions-for-python-test-discovery) but it kinda hints at it). 

In our case, I've named each file kinda alphabetically with `01_`, `02_`, ... `08_` to force an order during test collection and pytest doesn't seem to ever collect the tests in a different order other than the one specified 🤔

But yea arguably I think a cleaner and 'surer' way to ordering the tests would be to use the fixtures/hooks method suggested.

krithika369: I see.. Here's another SO answer that mentions that:
> The tests will most likely run in alphanumeric order based on the file name. 

https://stackoverflow.com/questions/44234414/which-file-is-executed-first-in-pytest/44234464#44234464

But it also highlights that parallelism could affect this behaviour. To be safe, since the tests are from different modules, shall we use fixtures to mark the dependency?

deadlycoconuts: Ok I've tried using fixtures to specify the test order as described in this response: https://stackoverflow.com/a/50575775. Unfortunately it doesn't work for chained fixtures, i.e. a test which itself is a fixture and uses another fixture (which is what we'd need since we need `08` to depend on `07`, which depends on `06`, ...), due to this bug specified in the warning of this section: https://stackoverflow.com/a/50575775:

> Warning
Note this mark has no effect in fixture functions. For example, this will not work as expected:
@pytest.mark.usefixtures("my_other_fixture")
@pytest.fixture
def my_fixture_that_sadly_wont_use_my_other_fixture():
    ...
Currently this will not generate any error or warning, but this is intended to be handled by [issue #3664](https://github.com/pytest-dev/pytest/issues/3664).

Instead, I've decided to use the `pytest_collection_modifyitems` hook recommended in `conftest.py`, which manually sorts out the collected test functions with respect to a fixed list of test names to ensure the correct test order during execution:

```python
TEST_ORDER = [
    "test_create_router",
    "test_update_router_invalid_config",
    "test_create_router_version",
    "test_undeploy_router",
    "test_deploy_router_invalid_config",
    "test_deploy_router_valid_config",
    "test_delete_router",
    "test_deploy_router_with_traffic_rules"
]

def pytest_collection_modifyitems(items):
    # modify test items in place to ensure that the e2e tests run in order
    item_mappings = {item.originalname: item for item in items}
    sorted_items = []
    for test_name in TEST_ORDER:
        sorted_items.append(item_mappings[test_name])
    items[:] = sorted_items
```

krithika369: I see. Cool, thanks for digging into this!

.github/workflows/turing.yaml
deadlycoconuts: I think that works, it does sound tedious and wasteful to be running too configurations for all the possible combinations. 

I've made some changes to the CICD workflows in order to adopt your suggestion (run API e2e tests for `In-cluster credentials` and SDK e2e tests for `Remote cluster credentials`): I've removed the `test-e2e-sdk` job to simplify the workflow in `turing.yaml` and instead added the SDK-specific steps ('Setup Python', 'Install Python dependencies', 'Run End-to-End Test Suite with Turing SDK') to the existing `test-e2e` job. These steps only get triggered when the value for `useSDK` is true; conversely, the API e2e tests only get triggered when this is false. I've also changed the matrix strategy to reflect this:

```yaml
      matrix:
        name: ["In-cluster credentials; API e2e", "Remote cluster credentials; SDK e2e"]
        include:
          - name: "In-cluster credentials; API e2e"
            useInClusterConfig: true
            valuesFile: "turing.values.in-cluster.yaml"
            useSDK: false
          - name: "Remote cluster credentials; SDK e2e"
            useInClusterConfig: false
            valuesFile: "turing.values.remote.yaml"
            useSDK: true
```

deadlycoconuts:
Thanks a lot for all the detailed comments! Just merged this! 🚀

PR #220, opened 2022-07-27
Add flags to deploy XP in Turing helm chart

terryyylim:

XP (Turing Experiments) has been released previously - https://github.com/caraml-dev/xp and this PR adds the ability to deploy XP via flags in Turing's helm chart.

`Update prior to closing`
E2E tests with PubSub is not doable without spinning up actual infrastructure on GCP. XP has a hard dependency on PubSub for message transmission between Management and Treatment Service. Making the message queue support Kafka would ideally unblock this task.


infra/charts/turing/requirements.yaml
```diff
@@ -14,3 +14,13 @@ dependencies:
     repository: file://subcharts/merlin
     tags:
       - mlp
+  - name: xp-management
+    version: 0.1.2
+    repository: https://turing-ml.github.io/charts
+    tags:
+      - mlp
```
krithika369, line 21: Although this is the default experiment engine for Turing, I think we should use a different tag for it, so that it can be turned off if an external user really doesn't want it (and so doesn't want to provision resources for it).

I think `xp` may be good enough as a tag and it can be set to `true` by default.

infra/charts/turing/templates/virtualservice-mlp.yaml
```diff
@@ -57,6 +57,18 @@ spec:
             host: {{ template "turing.merlin.name" . }}
             port:
               number: 8080
+    - match:
```
krithika369, line 60: How is this going to be used? I.e., how does Turing know when to call the `/api/xp/` endpoint?

infra/charts/turing/templates/virtualservice-mlp.yaml
```diff
@@ -85,6 +97,22 @@ spec:
...
+        - uri:
+            prefix: "/xp"
+      route:
+        - destination:
+            host: {{ template "turing.xpManagement.name" . }}
+            port:
+              number: 8080
+    - match:
+        - uri:
+            prefix: "/xp-treatment"
```
krithika369, line 110: Should we make it `/api/xp-treatment` for consistency with other APIs?

krithika369 (commented):
Thanks for this PR. On the whole it looks good. But this PR is missing the glue, i.e., the XP plugin configuration. With the current changes, I hope that XP is able to be successfully installed alongside Turing and its workflows can be exercised on the management and treatment services (if not, please help test that). But then, Turing is not 'aware' of XP. That should be configured though the default chart values for the experimentEngines, when XP is enabled. Then, it should be possible for Turing to call the XP management and treatment services during the respective workflows.

In the future, we may also look at configuring Turing and XP together (so that things like the MLP host don't need to be set twice) but I think those may be a part of the CaraML chart unification.

infra/charts/turing/templates/virtualservice-mlp.yaml
terryyylim: Updated this!

infra/charts/turing/templates/virtualservice-mlp.yaml
terryyylim: Good point, I think this will not be used, I've removed it.

infra/charts/turing/requirements.yaml
terryyylim: Updated the tag to be `xp` for deploying XP experiment engine with Turing.

infra/charts/turing/templates/_helpers.tpl
```diff
@@ -187,6 +195,7 @@ Sentry:
 {{ if .Values.turing.experimentEngines }}
 Experiment:
 {{ range $expEngine := .Values.turing.experimentEngines }}
+{{ if or (and (eq (toString $expEngine.name) "xp") (.Values).tags.xp ) (ne (toString $expEngine.name) "xp") }}
```
terryyylim, line 198: Added the conditional check here to "glue" XP to Turing if it is enabled by checking if:
1. experiment engine name is "xp" **AND** tags.xp is enabled **OR** 
2. experiment engine name is not "xp"

krithika369: The file you've modified makes sense to me. But let's think about the conditions.. What it does is exclude the `xp` props if the XP tag has not been set. This is good. But, if the user set a new experiment engine in their values file (say 'test-engine') and XP is enabled, what is the behavior we will see?

The requirements are:
1. XP should be enabled if the `xp` tag is enabled
2. The user should be able to add other experiment engines.
3. The user should be able to override the experiment engine configs for XP (I didn't mention this earlier but you can imagine that this is required to override the hosts, even for our own internal deployment).

The helm [merge](https://helm.sh/docs/chart_template_guide/function_list/#mergeoverwrite-mustmergeoverwrite) operation will not combine the list values but rather, it will replace the destination's list value (the default values file) with that of the source (user-supplied values file). We need to add some more handling around merging of the experiment engine configurations to achieve the requirements and I think it's easiest done on the merged values - after the configs are merged, if (XP tag is enabled and there is no existing config for xp in the experiment engines dict, we can append the default experiment engine config for XP).

I think we also need to modify the `openApiSpecOverrides` in a similar manner.

infra/charts/turing/templates/virtualservice-mlp.yaml
krithika369: I didn't actually mean that it should be removed - my question was really, how will Turing use this, hinting at the absence of the glue.

krithika369 (commented):
I think that the deployment will not be usable in the state that the PR is in today and I may have also missed some things in the review. Could you help test the changes after addressing the comments? It would help to add the acceptance criteria to the PR description as well.

krithika369:
@terryyylim, since work on this has been paused, let's mark this PR as draft.

krithika369:
Hey @terryyylim, given the recent conversations around the future ownership of https://github.com/caraml-dev/xp, we would need to de-prioritize this. To keep the context, could you help capture the status in the PR description (what the final blockers are to get this PR merged, that are yet to be addressed) and close it out? If and when we get back to this, it will be useful. Thanks!

PR #223, opened 2022-08-01
Satisfy Istio's requirement for named ports

romanwozniak:
### Context:

Istio [requires](https://istio.io/v1.0/docs/setup/kubernetes/spec-requirements/) service ports to be named in order to work correctly with the routing features. (It is also explained here in detail: [Protocol Selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/)). While the current chart works correctly with some basic usage of Istio (i.e. exposing a service via `VirtualService`), it fails to work as expected inside Istio's Service mesh.

This MR adds `http` name to Turing's `Service` and bumps up a patch version of the chart.

### DIFF
--- a/infra/charts/turing/Chart.yaml
+++ b/infra/charts/turing/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: "Turing: ML Experimentation System"
 name: turing
-version: 0.2.11
+version: 0.2.12
 appVersion: v1.0.0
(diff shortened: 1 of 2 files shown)


krithika369:
Thanks for this change. Do we also need to add it to the Merlin / MLP subcharts for the OSS users?

krithika369 (approved):

PR #225, opened 2022-08-04
Add UI handling with batch ensembling disabled

deadlycoconuts:
## Context
If batch ensembling is disabled on the Turing API, any attempts by the UI to display the ensembling jobs page will result in an error being shown. This is because the Turing API does not initialise a `EnsemblingJobController` to handle the ensembling job endpoints when batch ensembling is disabled, leading to a `404` response when the UI tries to send a request to the inexistent`/projects/{project_id}/jobs` endpoint when attempting to display the ensembling jobs page.

In order to gracefully avoid such a scenario, this PR introduces an additional landing view for the ensembling jobs page, displaying the usual `ListEnsemblingJobsView` if batch ensembling is enabled or a placeholder message when batch ensembling is disabled. This prevents the UI from sending requests to the inexistent endpoint and having to manually handle any resultant errors.

Screenshot 2022-08-04 at 7 39 38 PM

To track whether batch ensembling is enabled/disabled, the UI now consumes a new config variable `batchEnsemblingEnabled`. This is either passed through the `.env` file or via the helm chart values.

## Modifications
- `infra/charts/turing/values.yaml` - Addition of `batchEnsemblingEnabled` to `uiConfig.appConfig` in the helm chart values
- `ui/.env` - Addition of the env var `REACT_APP_BATCH_ENSEMBLING_ENABLED` that corresponds to `batchEnsemblingEnabled`
- `ui/src/config.js` - Addtion of config handling for the updated `appConfig`
- `ui/src/jobs/EnsemblingJobsRouter.js` - Replacement of `ListEnsemblingJobsView` with `ListEnsemblingJobsLandingView` to handle the ensembling jobs path
- `ui/src/jobs/list/ListEnsemblingJobsLandingView.js` - Addition of a new component to determine if the original `ListEnsemblingJobsView` or a placeholder message should be shown, depending on whether batch ensembling is enabled
- `ui/src/jobs/list/ListEnsemblingJobsView.js` - Minor refactoring to enable this class to become a child component of the new landing page

### DIFF
--- a/ui/src/jobs/list/ListEnsemblingJobsLandingView.js
+++ b/ui/src/jobs/list/ListEnsemblingJobsLandingView.js
@@ -1,0 +1,32 @@
+import { EuiPage, EuiPanel, EuiSpacer, EuiText } from "@elastic/eui";
+import React from "react";
+import { useConfig } from "../../config";
+import { ListEnsemblingJobsView } from "./ListEnsemblingJobsView";
+
+export const ListEnsemblingJobsLandingView = (props) => {
+  const {
+    appConfig: { batchEnsemblingEnabled },
+  } = useConfig();
+
+  return (
+    <>
+      {batchEnsemblingEnabled ? (
+        <ListEnsemblingJobsView {...props} />
+      ) : (
+        <EuiPage>
+          <EuiPanel>
+            <EuiText>
+              Batch ensembling has not been enabled for this deployment of
+              Turing.
+            </EuiText>
+            <EuiSpacer />
+            <EuiText>
+              To use batch ensembling, please enable it in the configuration
+              file used to deploy the Turing API.
+            </EuiText>
+          </EuiPanel>
+        </EuiPage>
+      )}
+    </>
+  );
+};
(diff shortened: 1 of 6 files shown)


ui/src/jobs/list/ListEnsemblingJobsLandingView.js
```diff
@@ -0,0 +1,32 @@
...
+        <EuiPage>
+          <EuiPanel>
+            <EuiText>
+              Batch ensembling has not been enabled for this deployment of
+              Turing.
+            </EuiText>
+            <EuiSpacer />
+            <EuiText>
+              To use batch ensembling, please enable it in the configuration
+              file used to deploy the Turing API.
```
krithika369, line 25: Nit: Can it be "Turing app" instead of "Turing API" ? It needs to be enabled both on the API and the UI, right? Besides, saying "API" in the UI view feels too technical.

krithika369 (approved):
Left one small comment. Other than that, LGTM.

Down the road, we could think about exposing an endpoint on the API for the UI to retrieve configs such as this, that are controlled by the API (rather than setting the config again on the UI).

ui/src/jobs/list/ListEnsemblingJobsLandingView.js
deadlycoconuts: That's true; I've edited the text to reflect "Turing app" instead as you suggested :)

deadlycoconuts:
Thanks for the comments, I'll be merging this now :)

PR #226, opened 2022-08-07
Fix helm chart template value and add pre-commit hooks

krithika369:
The motivation for this PR is to update the Helm template value in `infra/charts/turing/templates/_helpers.tpl` for the postgres port used, which was missed from the PR that updated it earlier: https://github.com/gojek/turing/pull/208

The incorrect template expression did not cause any problems in the e2e tests because the port used was `5432`, which is the same as the default port. Nonetheless, this PR corrects the mistake.

In addition, the [pre-commit](https://pre-commit.com/#intro) tool has been integrated which helps automate some additional checks / actions when some files are being modified (eg: regenerating the Helm docs with changes to the chart values). The  `.pre-commit-config.yaml` file has each of these hooks:
* Helm docs generation
* OpenAPI Python client generation (local hook, using the respective `make` target)
* Linting of Golang, Js, Python code (local hook, using the respective `make` target)

### Notes
* `README.md` - contains the installation details for the pre-commit hooks. We must run `make setup` on the repo once. Prior to this, it may be required to clear existing hooks (such as from husky, described further below) from the local environment, if the repo has already been cloned and used in the past.
```sh
$ git config --unset-all core.hooksPath
$ rm -rf .git/hooks
$ make setup
```
* `Makefile` - adding the dev `setup` target. Corresponding setup targets from other Makefiles are now redundant and have been removed.
* `ui/package.json` - Removing `husky` and `lint-staged` scripts and dev dependencies - these will now be covered by the common `pre-commit` hook. Having husky manage the hooks also conflicts with the pre-commit tool as they will both end up having to use the same git config value `core.hooksPath`.
* One notable difference between the husky hook for prettifying the UI vs the pre-commit hook is that any automatically modified files will not be automatically staged and added to the commit. Rather, the commit action will fail with a message and the developer is expected to `git add` the auto-modified files too. This behavior is the same for all pre-commit hooks and is apparently a conscious design decision (see https://github.com/pre-commit/pre-commit/issues/1099).

Screenshot 2022-08-07 at 3 27 02 PM

* Using the `files` filter in the pre-commit hook definitions can limit the running of hooks based on the modified files (instead of running all hooks always).

Screenshot 2022-08-07 at 11 27 42 PM

Most other modifications in this PR are auto-formatting fixes, particularly in the Python projects, from integrating [black](https://github.com/psf/black).


ui/package.json
```diff
@@ -43,18 +41,11 @@
     "react-scripts/**/postcss-normalize": "10.0.1",
     "**/react-beautiful-dnd": "git+https://github.com/gojekfarm/react-beautiful-dnd.git#master"
   },
-  "lint-staged": {
```
krithika369, line 46: This is now performed by the pre-commit prettier hook, minus the auto-staging of modified files. This is detailed in the PR description.

api/Makefile
```diff
@@ -11,22 +11,11 @@ clean:
...
-
 .PHONY: fmt
 fmt:
 	@echo "Formatting code..."
 	gofmt -s -w ${SRC_ROOT}
 
-.PHONY: lint
-lint: setup
-	@echo "Linting code..."
-	golangci-lint -v run --timeout 3m $(if $(filter true,$(fix)),--fix,)
```
romanwozniak, line 28: Is it required to remove it? What if I want to lint the code without committing the changes? Would it be better to just integrate `make lint` as a command in the pre-commit hook?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - repo: https://github.com/norwoodj/helm-docs
+    rev: v1.11.0
+    hooks:
+      - id: helm-docs-built
+        name: 'Helm Docs for Turing Chart'
+        files: '^infra/charts/turing/'
+        args:
+          - --chart-search-root=infra/charts/turing
+          - --template-files=./README.md.gotmpl
+      - id: helm-docs-built
```
romanwozniak, line 11: Is it expected that multiple hooks have the same `id`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - repo: https://github.com/golangci/golangci-lint
+    rev: v1.48.0
+    hooks:
+      # For the below hooks, we need to cd into the folder as we cannot execute golangci-lint
+      # outside the module. See: https://github.com/pre-commit/pre-commit/issues/1417
+      - id: golangci-lint
+        name: 'Linter for API'
+        files: '^api/'
+        types: [go]
+        entry: bash -c 'cd api && golangci-lint run --fix'
```
romanwozniak, line 26: Are we fine with the default timeout value now? It was previously set to 3 minutes

Makefile
```diff
@@ -24,3 +24,10 @@ build-image: version
 version:
 	$(eval VERSION=$(if $(OVERWRITE_VERSION),$(OVERWRITE_VERSION),v$(shell scripts/vertagen/vertagen.sh)))
 	@echo "turing version:" $$VERSION
+
+.PHONY: setup
+setup:
+	@echo "Setting up dev tools..."
+	@test -x pre-commit || pip install pre-commit
```
romanwozniak, line 31: Not sure if this works as intended:

```shell
➜  ~ test -x git || echo test
test
➜  ~ which git
/usr/local/bin/git
```

romanwozniak (approved):
LGTM

api/Makefile
krithika369: I was asking myself this. I rarely run `make lint` locally and usually the CI is the first to catch the problem (which is now expected to be sped up with the pre-commit hook). Anyway, I added it back.

But I went in the other direction, to let the pre-commit hook be run in the `--all-files` mode (as opposed running only if something changed, which is the behaviour of the automatic hook) in the make target. This is so that the set up is nicely contained in `.pre-commit-config.yaml` rather than installing `golangci-lint` externally.

```Make
.PHONY: lint
lint:
	@pre-commit run golangci-lint-experiment --all-files
```

.pre-commit-config.yaml
krithika369: Yes, the `id` is simply used to match the hook from the published repo, as there can be multiple. Eg: https://github.com/norwoodj/helm-docs/blob/master/.pre-commit-hooks.yaml

To 'distinguish' one hook from another where the same id is used, the `alias` property can be set (which I've now added to link to the respective Go modules' `make lint`:

```yaml
     - id: golangci-lint
        name: 'Linter for API'
        alias: golangci-lint-api
        files: '^api/'
        types: [go]
        entry: bash -c 'cd api && golangci-lint run --timeout 3m --fix'
      - id: golangci-lint
        name: 'Linter for the Router Engine'
        alias: golangci-lint-router
        files: '^engines/router/'
        types: [go]
        entry: bash -c 'cd engines/router && golangci-lint run --fix'
```

.pre-commit-config.yaml
krithika369: It hasn't taken as long for me recently. But fair point, I added it back for the `api` module.

Makefile
krithika369: Thanks for catching. I fixed it now using: `test -x "$(which pre-commit)" || pip install pre-commit`

api/Makefile
romanwozniak: > But I went in the other direction

I'm not a fan of this, would like to hear the opinions of other folks. I'm used to seeing `Makefile` as a mission control to  "rule them all" of the repo – i.e. to contain all the recipes of the repository. In my opinion, it's the make file that should contain the configuration of the recipe, where the pre-commit hook only helps to run this command and the prescribed time (pre-commit). In other words, make commands should work perfectly fine without the pre-commit, where the opposite is not guaranteed.

krithika369: I agree it's unconventional to have a "pre-commit" tool drive the Makefile scripts. But I think it's more than that (the name is less helpful - in this case we are talking about a `manual` hook): https://pre-commit.com/#advanced

Especially for something like linting, I doubt that we would invoke this manually as often. Secondly, for some of our components, the CI doesn't rely on the same mechanism either (we use the relevant GitHub action as opposed to `make lint`). So my preference is the following:
1. Remove the explicit lint target and simply rely on the hooks
2. Accept `pre-commit` as our standard dev tooling and be ok with it being invoked from the Makefile
3. Convert most/all pre-commit hooks to local and let the Makefile rule them all (what you're suggesting). I am less enthused by this because we have to handwrite the script for installing the linter, testing if its version matches, etc. than leveraging something that's declarative. But no strong opinion on this.

Other @gojek/turing-dev, any thoughts?

sdk/samples/batch_job/batch_job copy.py
```diff
@@ -0,0 +1,95 @@
+import time
```
deadlycoconuts, line 1: Is this file an unintended copy?

sdk/samples/router/create_router_with_pyfunc_ensembler_2.py
```diff
@@ -0,0 +1,150 @@
+import turing
```
deadlycoconuts, line 1: This file seems like it's unintentionally staged and committed :o

sdk/samples/batch_job/ensembler.py
```diff
@@ -0,0 +1,63 @@
+import time
```
deadlycoconuts, line 1: Is this file committed unintentionally?

api/Makefile
deadlycoconuts: I use `make lint` sometimes so it'll be less convenient to see it go and I see how it's weird to be seeing pre-commit hooks in a Makefile. However it's also true that this means we'll have to 'handwrite the script for installing the linter, testing if its version matches'. Not too sure what the usual convention to address this so I don't really have a hard preference.

sdk/samples/router/create_router_with_pyfunc_ensembler_2.py
krithika369: Damn, thanks for catching all of these, @deadlycoconuts ! I was trying to be careful about my local files at first but subsequently missed them. I'm going to revert the history and re-apply the linting changes just to be sure.

krithika369: TIL that force-pushing the branch all the way back to the first commit before that automatically closes the PR (which makes sense). I'll open a new one shortly.

api/Makefile
krithika369: Thanks for the input, @deadlycoconuts !

> Would there an issue keeping the make lint targets without pre-commit hooks and then accepting that we'll use the pre-commit hooks when committing as part our standard dev workflow? Only issue I can see with this is that we'll have 2 locations to maintain linters/versions, etc.

Yes, I considered this too but was not too keen on it because of the downside you've highlighted. Let's call it. # 4 :) 

@romanwozniak it seems neither Ziyi nor I have a strong opinion on this. And also this is something that may be applicable to all our repos. So, let's go with your call - are you still in favor of option # 3 ? And if so, should we be consistent about all the hooks being local, linking to the respective Make targets ?

sdk/samples/router/create_router_with_pyfunc_ensembler_2.py
deadlycoconuts: Oops, what happens if you now force-push new commits to it? Does the PR remain closed? :o

krithika369: I don't know but I guess I can try :)

krithika369: That did work. :)

Once again, thanks for catching this. There wasn't any secret info but those files should not have been there. Reminder to self to be extra careful about PRs with large changes and to review the file change summary:

Screenshot 2022-08-08 at 4 00 23 PM

api/Makefile
krithika369: Hello @romanwozniak, any thoughts after the recent comments on this thread?

romanwozniak: I still prefer option 3 the most, but it's on you to decide what works the best.

>  And if so, should we be consistent about all the hooks being local, linking to the respective Make targets ?

Generally, I like that, except that for the front-end module, npm/yarn effectively replaces Makefile with `package.json > scripts` (meaning, that I don't think to have a Makefile for the UI module, because it would just duplicate yarn script)

krithika369: Cool. I've changed all except the helm-docs generation (which is unlikely to be run locally using `make`) as local hooks.

PR #228, opened 2022-08-15
Add CaraML docs sync workflow

deadlycoconuts:
With the migration of `turing` to the CaraML organisation, we are syncing the docs from `turing` automatically to https://github.com/caraml-dev/docs. 

This PR introduces an additional workflow which performs this step, which can be found here: https://github.com/caraml-dev/docs/tree/main/.github/actions/trigger-remote-docs-sync

**Which issue(s) this PR fixes**:
None

### DIFF
--- a/.github/workflows/trigger-caraml-doc-sync.yml
+++ b/.github/workflows/trigger-caraml-doc-sync.yml
@@ -1,0 +1,22 @@
+# This workflow triggers sync of docs to caraml-dev/docs
+
+name: Trigger CaraML Docs Sync
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
+  trigger-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Remote Doc Sync Workflow
+        uses: caraml-dev/docs/.github/actions/trigger-remote-docs-sync@main
+        with:
+          module: 'router'
+          git_https: 'https://github.com/caraml-dev/turing.git'
+          doc_folder: 'docs'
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-caraml-doc-sync.yml
```diff
@@ -0,0 +1,22 @@
...
+      - 'v[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
+  trigger-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger Remote Doc Sync Workflow
+        uses: caraml-dev/docs/.github/actions/trigger-remote-docs-sync@main
+        with:
+          module: 'router'
```
terryyylim, line 17: Currently, `turing` docs are under `experiment` module (https://github.com/caraml-dev/docs/tree/main/module/experiment), is the plan here to keep Turing and XP separate or under the same module? And if we want it under the same module, then this workflow likely would need to change right?

deadlycoconuts: Hmm, I spoke to Eric about this yesterday and I think we were planning to put `turing` docs under 'Router` and `xp` docs under 'Experiment', both modules having the same level in the tree hierarchy. I'm open to using our existing structure of `xp` being under `turing` too; what do you all think? @krithika369

krithika369: Just noticed this - as we discussed during the stand up, for the public docs, I think the current structure makes more sense (as we will not be using the internal code names) than nesting Experiments under Routers.

terryyylim (approved):

PR #229, opened 2022-08-19
Update UI dependencies

deadlycoconuts:
## Context
With the update of `mlp-ui` dependencies in https://github.com/gojek/mlp/pull/54 and https://github.com/gojek/mlp/pull/55 (in particular the update of `Node.js` **14 → 16**, `React` **16 → 17**, `@elastic/eui` **32.3.0 → 66.0.0** and `react-scripts` **to 5.0.1**), there is a need to similarly update Turing UI, as it uses components from and is dependent on the `mlp-ui` package. 

## Main Modifications
- Replacement of the existing `Empty` component with the new `Page404` component
- Extensive adoption of the new `EuiPageTemplate` (see https://elastic.github.io/eui/#/templates/page-template) to manage page layouts; this replaces the existing usage of `EuiPage`, `EuiPageHeader`, `EuiPageSection`, etc. components
- Update of the existing colour palette manually specified to align with those of `v66` of `@elastic/eui` (see https://eui.elastic.co/#/theming/colors/values)

## Known Issues (Unable to be addressed in this PR)
- To follow up on `mlp-ui` to fix the `StepsWizardHorizontal` component as it does not seem to be highlighting as users progress through the timeline steps:
Screenshot 2022-08-31 at 5 41 00 PM
- The `EuiSearchBar` used by the `PodLogsViewer` has been updated in Elastic UI such that the `field_value_selection` filter (implemented as `EuiSelectable` under the hood) no longer passes a newly selected filter to the `onChange` handler when another filter has already been clicked. This causes a newly clicked `tail_lines` filter to not be reflected:
ezgif com-gif-maker


ui/package.json
```diff
@@ -3,21 +3,22 @@
...
     "query-string": "^7.0.1",
-    "react": "16.14.0",
+    "react": "17.0.2",
     "react-collapsed": "^3.0.1",
     "react-diff-viewer": "^3.1.1",
-    "react-dom": "16.14.0",
+    "react-dom": "17.0.2",
     "react-lazylog": "git+https://github.com/gojekfarm/react-lazylog#master",
     "react-scroll": "^1.7.16",
     "react-scroll-to-bottom": "^4.0.0",
```
romanwozniak, line 24: @romanwozniak,  please also update `react-scripts` to the stable version: ref https://github.com/gojek/mlp/pull/55

ui/src/PrivateLayout.scss
```diff
@@ -1,3 +1,7 @@
 .main-component-layout {
   padding-top: 49px;
 }
+
+.css-1asjec2-euiPageSection__content-l-restrictWidth {
```
romanwozniak, line 5: Is it something temporary? You are not intended to modify the `@emotion/react` CSS styles directly. Also, I'm not sure if this CSS class name is stable (you can guess it from this generated name `.css-1asjec2-...`)

ui/src/components/config_section/ConfigSectionTitle.js
```diff
@@ -1,15 +1,16 @@
 import React from "react";
-import { EuiIcon, EuiTextColor, EuiTitle } from "@elastic/eui";
+import { EuiIcon, EuiTextColor, EuiTitle, EuiSpacer } from "@elastic/eui";
 
 export const ConfigSectionTitle = ({ title, iconType }) => (
   <EuiTitle size="s">
-    <EuiTextColor color="secondary">
+    <EuiTextColor color="#007871">
```
romanwozniak, line 6: This doesn't look right. Maybe you can take some inspiration from this: https://elastic.github.io/eui/#/theming/customizing-themes ?

ui/src/jobs/details/EnsemblingJobDetailsView.js
```diff
@@ -31,47 +28,44 @@ export const EnsemblingJobDetailsView = ({ projectId, jobId, ...props }) => {
...
-                  prepend={
-                    <StatusBadge
-                      status={JobStatus.fromValue(jobDetails.status)}
-                    />
-                  }
-                />
-              </EuiPageHeaderSection>
-            </EuiPageHeader>
-
+    <EuiPageTemplate restrictWidth="95%">
```
romanwozniak, line 31: I can imagine that the width restriction would be the same on all the app pages. Maybe we can parameterize it? Something similar to this: https://github.com/elastic/eui/blob/aea4705f832b27e253ce433ea0ccfcae7a7414d4/wiki/consuming.md#a-recommended-method-to-consume-theming-variables-using-emotion

ui/src/PrivateLayout.scss
deadlycoconuts: Yeah it is something temporary, I was just publishing the image temporarily to test out something else; but I'm going to fix this before opening this PR.

ui/src/components/config_section/ConfigSectionTitle.js
krithika369: Based on the color code and the release notes, "secondary" is simply renamed as "success" where it was present: https://github.com/elastic/eui/releases/tag/v34.5.0

ui/src/router/details/RouterDetailsView.js
```diff
@@ -104,8 +101,8 @@ export const RouterDetailsView = ({ projectId, routerId, ...props }) => {
                 </RouterActions>
               </Fragment>
             )}
-            <EuiSpacer size="m" />
-
+          </EuiPageTemplate.Header>
+          <EuiPageTemplate.Section color={"transparent"}>
```
krithika369, line 105: Why do we need this `color={"transparent"}` ?

ui/src/jobs/details/EnsemblingJobDetailsView.js
krithika369: @romanwozniak are you suggesting consuming some Eui Global variables using Emotion as in the link? Or is it merely about the "parameterising" part which we currently do by defining configs? https://github.com/caraml-dev/turing/blob/main/ui/src/config.js#L51

ui/src/router/details/RouterDetailsView.js
deadlycoconuts: Otherwise this happens 😅
Screenshot 2022-08-30 at 7 12 00 PM

ui/src/jobs/details/config/prediction_config_section/PredictionsConfigSection.scss
```diff
@@ -1,9 +1,10 @@
-@import "~@elastic/eui/src/global_styling/variables/colors";
+@import "@elastic/eui/src/themes/amsterdam/colors_light";
+@import "@elastic/eui/src/themes/amsterdam/globals";
 
 .predictionSources {
-  .euiStepNumber--small {
-    background-color: $euiColorSecondary;
-  }
+  //.euiStepNumber--small {
```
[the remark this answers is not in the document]
deadlycoconuts: I've updated it to reference to correct colour but it seems like `v64` of Elastic UI has made the success colour really... cheerful and bright (see image below for a rough idea), which is pretty different from the darker tones of green/blue in our UI :/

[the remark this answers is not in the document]
krithika369: Haha IMO it's not too bad. Shall we get the changes deployed internally for everyone to play with and gather feedback? Also, if we are defining theme overrides, I think we should do it in the parent MLP library and consume in each of the components - Turing, Merlin, etc.

ui/package.json
romanwozniak: I meant you, @deadlycoconuts 😅

ui/src/jobs/details/EnsemblingJobDetailsView.js
deadlycoconuts: > Or is it merely about the "parameterising" part which we currently do by defining configs?
I can do this in the meantime and then see if we're unifying all these variables in `mlp-ui`

ui/src/components/config_section/ConfigSectionTitle.js
deadlycoconuts: Hmm something is overwriting the colour corresponding to `success` to black instead of `#00BFB3`. There're probably some styles in `mlp-ui` overwriting this.

ui/src/services/job/JobStatus.js
```diff
@@ -3,17 +3,17 @@ import { Enum, EnumValue } from "../enum/Enum";
 export const JobStatus = Enum({
   PENDING: EnumValue("pending", {
     label: "Pending",
-    color: "#F5A700",
+    color: "#FEC514",
```
deadlycoconuts, line 6: Updated the values of the status colours here with respect to those found here: https://eui.elastic.co/#/theming/colors/values

ui/src/jobs/details/EnsemblingJobDetailsView.js
```diff
@@ -31,47 +35,47 @@ export const EnsemblingJobDetailsView = ({ projectId, jobId, ...props }) => {
...
-                    <StatusBadge
-                      status={JobStatus.fromValue(jobDetails.status)}
-                    />
-                  }
-                />
-              </EuiPageHeaderSection>
-            </EuiPageHeader>
-
+    <EuiPageTemplate restrictWidth={restrictWidth} paddingSize={paddingSize}>
+      <EuiSpacer size="l" />
```
deadlycoconuts, line 39: Instead of trying to hack the default CSS styles to make the page sections having the desired spacing, I found a way to use `EuiSpacer`s to create the same effect:

Screenshot 2022-08-31 at 4 35 06 PM

This will probably remove the need to overwrite any existing styling configs.

ui/src/components/config_section/index.scss
```diff
@@ -9,7 +9,7 @@
 }
 
 .euiPanel--configSection {
-  .euiDescriptionList.euiDescriptionList--responsiveColumn {
+  .euiDescriptionList {
```
deadlycoconuts, line 12: This is done to keep the alternating row styling for description lists.

ui/src/components/config_section/ConfigSectionTitle.js
krithika369: You're right - `EuiTitle` on its own is using a black-ish color which overrides what we set:

Screenshot 2022-09-01 at 11 30 34 AM

It works if we wrap the Text entity in `<span>`:
```html
 <EuiTitle size="s">
    <span>
      <EuiTextColor color="success">
```

In the previous Eui version, you can see the Euititle style had the black color which was correctly being overridden by the color we set:
Screenshot 2022-09-01 at 11 37 38 AM


But even with the `span` workaround, there is an unexpected outcome - the text is still using the old "secondary" value which isn't cheerful: https://elastic.github.io/eui/#/display/text#coloring-text

I believe that's still fine? (Wont hurt eyes as much lol.)
Screenshot 2022-09-01 at 11 34 25 AM

krithika369: There is a similar discussion (although old) on the Eui repo, the suggestion seems to be to use a heading / span element inside the title: https://github.com/elastic/eui/issues/956

ui/src/jobs/list/ListEnsemblingJobsView.js
```diff
@@ -78,19 +76,21 @@ export const ListEnsemblingJobsView = (props) => {
...
-          </EuiPageHeaderSection>
-          <EuiPageHeaderSection>
-            <EuiButton fill disabled href={"jobs/create"}>
-              Submit Job
-            </EuiButton>
-          </EuiPageHeaderSection>
-        </EuiPageHeader>
-        <EuiPageContent>
+    <EuiPageTemplate restrictWidth={restrictWidth} paddingSize={paddingSize}>
+      <EuiSpacer size="l" />
```
terryyylim, line 80: Just curious, since I see it at quite a few instances where `EuiPageTemplate` is being used. What do these `EuiSpacer` at the start and end of a `EuiPageTemplate` achieve?

ui/src/router/activity_log/events_list/EventsList.js
```diff
@@ -19,33 +19,36 @@ export const EventsList = ({ events, status }) => {
...
-          username={<PulseLoader size={6} color={infoColor} />}
-          type="update"
-        />
-      )}
+      <EuiCommentList>
+        {events.map((event, idx) => (
+          <EuiComment
+            key={`event-${idx}`}
+            username={badge(event)}
+            timelineAvatarAriaLabel="System"
```
terryyylim, line 33: I don't think we need an `aria-label` here since its not an interactive icon anyway.

ui/src/router/activity_log/events_list/EventsList.js
```diff
@@ -19,33 +19,36 @@ export const EventsList = ({ events, status }) => {
...
+            timelineAvatar={timelineIcon(event)}
+            type="update"
```
terryyylim, line 35: What does `type="update"` do? I couldn't find this property in EUI documentation.

image

ui/src/config.js
```diff
@@ -60,6 +60,12 @@ const appConfig = {
     // Default number of tail log entries to be fetched
     defaultTailLines: 1000,
   },
+  pageTemplate: {
```
terryyylim, line 63: Are these page template configurations only for Turing, or all of our apps in general would likely face this issue? (If so, maybe a `TODO` to set these configurations on MLP parent app would be good) Would also be nice to write a simple note on why these values are needed.

terryyylim (commented):
New UI looking good! Thank you for this @deadlycoconuts ! Left some comments for addressing.

ui/src/components/config_section/ConfigSectionTitle.js
deadlycoconuts: Okay, so I guess this means we're going with the old `secondary` colour, and then letting `mlp-ui` do the upgrading of the styles to the new `success` colour :) I'll update all the instances where we're using `EuiTextColor` so that they'll use the old colour.

ui/src/jobs/list/ListEnsemblingJobsView.js
deadlycoconuts: It's to simulate the padding that we've removed from `EuiPageTemplate` (we're passing `"none"` as the `paddingSize` prop). The large `EuiSpacer`s at the top and bottom of a page adds some semblance of a padding at these places respectively, while the medium one just above an `EuiPageTemplate.Section` adds some space before the `EuiPageTemplate.Header` above it. See [here](https://github.com/caraml-dev/turing/pull/229#discussion_r959318467) for the effect I was trying to create

ui/src/router/activity_log/events_list/EventsList.js
deadlycoconuts: Just removed that :)

ui/src/router/activity_log/events_list/EventsList.js
deadlycoconuts: It's a legacy prop from an older implementation of the component. It has been removed from the latest changes that I've pushed.

ui/src/config.js
deadlycoconuts: For the moment they are only specified for Turing for consistency across the pages but yes the idea is to get them specified somewhere at the MLP level. This was pointed out in this comment here https://github.com/gojek/mlp/pull/56#discussion_r960366978.

For the moment I'll leave a comment in this config file mentioning that these values will soon be determined by the future `mlp-ui` version Turing shall consume.

terryyylim: Cool thanks for clarification! `nit`: Should we have `TO-DO` in the comment, so that its easier to locate?

deadlycoconuts: Okay, just added that in.

krithika369:
@deadlycoconuts can this PR be merged? Any unresolved issues still?

deadlycoconuts:
I think I've resolved most of the comments raised, if there are no outstanding comments about the look/feel of the UI, I guess I can merge this soon :)

romanwozniak:
> I think I've resolved most of the comments raised, if there are no outstanding comments about the look/feel of the UI, I guess I can merge this soon :)

One more thing, is there a way to re-introduce the indication of the current step in the multi-step wizard forms? 
Screenshot 2022-09-07 at 11 30 54

deadlycoconuts:
> One more thing, is there a way to re-introduce the indication of the current step in the multi-step wizard forms?
Hmm that component is imported from `mlp-ui` so I'll have to make the changes there.

krithika369:
> > I think I've resolved most of the comments raised, if there are no outstanding comments about the look/feel of the UI, I guess I can merge this soon :)
> 
> One more thing, is there a way to re-introduce the indication of the current step in the multi-step wizard forms? Screenshot 2022-09-07 at 11 30 54

@romanwozniak thanks for the reminder. As @deadlycoconuts mentioned, it is likely a fix in the MLP library so we decided to address it as a separate bugfix.

Shall we log this (and any other deferred regression) as a GitHub issue in the Turing repo and link it to the PR description?

krithika369 (approved):

deadlycoconuts:
Thanks for the numerous comments everyone! I'll be merging this now! 🚀

PR #231, opened 2022-08-29
Add name field to Traffic Rule

terryyylim:
In scope with grouping conditions/routes by Traffic rules in ViewRouter page, Traffic rules should have meaningful names that describe conditions for request routing logic. This PR adds `name` as a required field to the Traffic Rule entity. Implementation have been made for the API, UI and SDK.

**Which issue(s) this PR fixes**:

`NONE`

**UI Screenshots**:
**Rule name validation**
image
image

**Rule name form field**
image


api/turing/validation/validator.go
```diff
@@ -118,6 +120,23 @@ func newExperimentConfigValidator(expSvc service.ExperimentsService) func(valida
...
+	nameRegexDescription := strings.Join([]string{
+		"Name must be between 4-64 characters long, and begin with an alphanumeric character",
+		"and have no trailing spaces and can contain letters, numbers, blank spaces and the following symbols: -_()#$&:.",
+	}, " ")
+	if !trafficRuleNameRegex.MatchString(value) {
+		sl.ReportError(value, fieldName, "trafficRuleName", nameRegexDescription, fmt.Sprintf("%v", value))
+	}
+}
+
+func checkTrafficRulesUniqueName(sl validator.StructLevel, fieldName string, value []string) {
```
krithika369, line 133: Can we replace the custom uniqueness check with the tag: https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Unique

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -81,6 +82,21 @@ export const RuleCard = ({
         </EuiFlexItem>
       </EuiFlexGroup>
 
+      <EuiFormRow
```
krithika369, line 85: A screenshot of this, in the PR description would be nice, for illustration purposes.

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -81,6 +82,21 @@ export const RuleCard = ({
...
+        label="Name *"
+        isInvalid={!!get(errors, "name")}
+        error={get(errors, "name")}
+        fullWidth>
+        <EuiFieldText
```
krithika369, line 90: This got me thinking - when we get to the UI changes for the Traffic rules, we could add the rule name to the existing routes view as well.

krithika369 (commented):
On the whole it look good! I left some small comments. Besides this, as mentioned in the internal MR, let's move the data migration to the SQL migrations folder so it's easy to apply / revert for our OSS users. Thanks!

api/turing/validation/validator.go
terryyylim: Would you mean placing `"validate:unique=name"` tag for `TrafficRules` here - https://github.com/caraml-dev/turing/blob/main/api/turing/models/traffic_rules.go#L16?

krithika369: I think it needs to be added in the parent struct ([here](https://github.com/caraml-dev/turing/blob/main/api/turing/api/request/request.go#L27)). But I haven't used it in the past, so please help review the docs / test it out.

terryyylim: Indeed, we need to add to the parent struct, more specifically, `unique=Name` should come before `dive`.

api/db-migrations/000012_add_traffic_rule_name.up.sql
```diff
@@ -0,0 +1,14 @@
+UPDATE router_versions
```
krithika369, line 1: Missing semicolon at the end. But more importantly, are you sure this query is doing what it should? From what I can tell, this will update every row with the same result.

Could you help verify this and capture the results for a couple of rows, either here in the comments or in the PR description so it's easier to be on the same page?

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
krithika369: Missed this earlier - we also need to update the traffic rule schema used for yup validation (`ui/src/router/components/form/validation/schema.js`).

api/turing/validation/validator.go
krithika369: This looks good, thanks!

api/db-migrations/000012_add_traffic_rule_name.up.sql
terryyylim: Given current version = 11 and dirty = false `schema_migrations` table,

`Before`
image

`After`
image

By every row, do you mean for each router version?

krithika369: > By every row, do you mean for each router version?

Yes, 2 different router versions (each with multiple different rules, as you've captured above).

terryyylim: Ah, I think I got what you mean. I've updated the query now and the following screenshots depict changes to DB data after before and after app start-up with schema_migrations `version=11, dirty='f'`.

`Before`
image

`After`
image

krithika369: The output does look right to me, but strangely, when I tried to run it, I got the following error: `cannot extract elements from a scalar`. I dumped our internal dev DB to test the migration. Could you check this further?

krithika369: I think I figured out the problem - the `traffic_rules` column does not have a NOT NULL constraint and the null values are affecting the result. Could you update the query to handle such rows?

api/db-migrations/000012_add_traffic_rule_name.up.sql
```diff
@@ -0,0 +1,42 @@
...
+                (
+                    SELECT
+                        id,
+                        position,
+                        CONCAT('rule_', position::text) as name
+                    FROM
+                        router_versions,
+                        jsonb_array_elements(traffic_rules) with ordinality arr(elem, position)
+                ) AS t1
+                LEFT JOIN
```
krithika369, line 29: I don't think this LEFT JOIN is needed:

```sql
SELECT
    id,
    position,
    CONCAT('rule_', position::text) as name,
    elem
FROM
    router_versions2,
    jsonb_array_elements(traffic_rules) with ordinality arr(elem, position);
```

krithika369: In fact, I'd simplify the whole update query to this (still need to add in the null handling):
```sql
WITH tbl1 AS (
    SELECT id, position, CONCAT('rule_', position:: text) AS name, elem
    FROM router_versions2, jsonb_array_elements(traffic_rules) WITH ordinality arr(elem, position)
),
tbl2 AS (
    SELECT id, position, jsonb_build_object(
        'conditions', elem -> 'conditions',
        'routes', elem -> 'routes',
        'name', name
    ) AS traffic_rule FROM tbl1
),
tbl3 AS (
    SELECT id, json_agg(traffic_rule ORDER BY position) AS traffic_rules_updated
    FROM tbl2 GROUP BY id
)
UPDATE router_versions2 AS t SET traffic_rules=traffic_rules_updated
FROM tbl3
WHERE tbl3.id = t.id;
```

In general, I find the [CTE](https://www.geeksforgeeks.org/cte-in-sql/) query more readable.

api/db-migrations/000012_add_traffic_rule_name.up.sql
terryyylim: The issue is with "null" strings rather than null values, I've updated the query to handle these rows. Also tested them on internal dev and staging DB data.

api/db-migrations/000012_add_traffic_rule_name.up.sql
terryyylim: Updated the query

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
terryyylim: Added validation and screenshot in PR description.

api/db-migrations/000012_add_traffic_rule_name.down.sql
```diff
@@ -0,0 +1,30 @@
+UPDATE router_versions
```
krithika369, line 1: Can we update the down query to follow the same workflow as the up query? Currently, this is also missing the `traffic_rule ORDER BY position` as I added in the up query (but forgot to mention) which will ensure the rule ordering is preserved.

ui/src/router/components/form/validation/schema.js
```diff
@@ -84,6 +85,13 @@ const ruleConditionSchema = yup.object().shape({
 });
 
 const trafficRuleSchema = yup.object().shape({
+  name: yup
```
krithika369, line 88: Shall we also add the uniqueness check (which is easy to do on the frontend)? Similar to: https://github.com/caraml-dev/xp/blob/main/ui/src/experiments/components/form/validation/schema.js#L184

krithika369 (approved):
Added a couple more comments. Please also help test that the down query is working as expected after updating it. Once you've addressed the comments, feel free to merge. Thanks!

PR #232, opened 2022-09-01
Add Default Traffic Rule

terryyylim:
This is a follow-up PR, after `name` field has been introduced to Traffic Rule entity.

`API & UI Changes`
When the first traffic rule is added, a default traffic rule should be automatically added, which has no name nor conditions and only its routes are editable. A `"Select all"` option should be added to allow selection of all routes for the Default Traffic Rule.

The Default Traffic Rule should always be evaluated at last and acts as a fallback rule and on the UI, it should be separated from the Custom Traffic rules section. When the last Custom Traffic rule is removed, the Default Traffic rule should be removed as well.

Since ordering matters when defining Traffic rules, drag and drop functionality has also been introduced for Custom Traffic rules section. 

A minor enhancement has also been made to the current implementation, where the last selectable row for routes will be removed when all routes have been selected.

`DB Changes`
A new column `default_traffic_rule` is being added to `router_versions` table, which takes the structure of a jsonb object to provide more flexibility should we need to add and utilise other information related to a default traffic rule.

DB changes are handled accordingly:
- For DB rollup, Traffic rules = NULL/‘null’/[] -> default_traffic_rule = NULL (dont consider these rows when updating)
- For DB rollup, Traffic rules with no dangling routes -> default_traffic_rule = all routes
- For DB rollup, Traffic rules with dangling routes -> default_traffic_rule = only dangling routes
- For DB rolldown, if all routes are present in default_traffic_rule, it means there are no dangling routes

**UI workflow changes**

https://user-images.githubusercontent.com/25025366/189571965-8a5db686-b7b2-43b8-a3e4-bfb522984efe.mov

image
image

**Related PRs**
- https://github.com/caraml-dev/turing/pull/231

**TO-DO**
- Add Default Traffic Rule information to Router details view


api/turing/testdata/cluster/servicebuilder/router_version_success_traffic_splitting.json
```diff
@@ -53,6 +55,12 @@
       "routes": [
         "treatment-b"
       ]
+    },
+    {
+      "name": "default",
```
krithika369, line 60: Same comment as earlier - can we set control in the default rule instead?

api/turing/testdata/cluster/servicebuilder/router_configmap_traffic_splitting.yml
```diff
@@ -122,5 +158,7 @@ strategy:
         values:
         - region-b
       route_id: traffic-split-1
+    - conditions: null
```
krithika369, line 161: It may be better to use `[]` here instead of `null` to avoid any special handling.

api/turing/validation/validator.go
```diff
@@ -129,6 +129,43 @@ func checkTrafficRuleName(sl validator.StructLevel, fieldName string, value stri
...
+	if ruleLength == 1 {
+		sl.ReportError(rules, fieldName, "trafficRule", ruleCountDescription, "")
+	}
+	// Check for presence of Default Traffic Rule
+	hasDefaultRule := false
+	noDefaultRuleDescription := "Default rule should be provided."
+	if ruleLength > 1 {
+		for _, rule := range rules {
+			if rule.Name == "default" {
+				hasDefaultRule = true
```
krithika369, line 144: Once the value is set to true, we can `break` early. But there is another condition we must check - the `default` rule is the last in the rules list (if not, it will be matched early at runtime). So, can we replace this check with:
```golang
if ruleLength > 1 {
    if rules[ruleLength-1] != "default" {
        // Report error
    }
}
```

This makes me wonder, does the name `default` really matter? Should we check for the name at all or simply that the last rule has 0 conditions ?

api/turing/validation/validator.go
```diff
@@ -129,6 +129,43 @@ func checkTrafficRuleName(sl validator.StructLevel, fieldName string, value stri
 	}
 }
 
+func checkTrafficRuleCount(sl validator.StructLevel, fieldName string, rules models.TrafficRules) {
+	ruleLength := len(rules)
+	ruleCountDescription := "There should be either 0 or more than 1 traffic rules specified."
```
krithika369, line 134: I feel like this error can be more descriptive to help the users - `... more than 1 traffic rules (which must include a default rule)` specified.

api/turing/validation/validator.go
```diff
@@ -129,6 +129,43 @@ func checkTrafficRuleName(sl validator.StructLevel, fieldName string, value stri
...
+		// Default rule should be present if any custom rules exist
+		if !hasDefaultRule {
+			sl.ReportError(rules, fieldName, "trafficRule", noDefaultRuleDescription, "")
+		}
+	}
+}
+
+func checkTrafficRuleConditions(sl validator.StructLevel, fieldName string, rule *models.TrafficRule) {
+	noConditionsDefaultRuleDescription := "Default rule should not have any conditions specified."
+	noConditionsCustomRuleDescription := "Custom rule must have conditions specified."
```
krithika369, line 156: `Custom rule must have conditions specified.` -> `Custom rule must have at least 1 condition.` ?

api/turing/validation/validator.go
```diff
@@ -129,6 +129,43 @@ func checkTrafficRuleName(sl validator.StructLevel, fieldName string, value stri
...
+		if len(rule.Conditions) != 0 {
+			sl.ReportError(rule, fieldName, "trafficRule", noConditionsDefaultRuleDescription, "")
+		}
+	} else {
+		if len(rule.Conditions) == 0 {
+			sl.ReportError(rule, fieldName, "trafficRule", noConditionsCustomRuleDescription, "")
+		}
+	}
+}
+
```
krithika369, line 168: In general, all of the special handling around the default rule is making me think we may be better off splitting it into a new field or redoing the `traffic_rules` column to explicitly state in the contract. Do you have any ideas around this?

krithika369: Also, do we have an end-user migration strategy for this?

ui/src/components/card/CardHeader.js
```diff
@@ -6,7 +6,7 @@ import {
   EuiIcon,
 } from "@elastic/eui";
 
-export const ExperimentCardHeader = ({ onDelete, dragHandleProps }) => (
+export const CardHeader = ({ onDelete, dragHandleProps }) => (
```
terryyylim, line 9: Refactored `ExperimentCardHeader` to `CardHeader` for use across Experiment Engine and Traffic Rules Drag&Drop components.

api/turing/validation/validator.go
terryyylim: This PR is still pending 2 items as listed in the description's `TO-DO` section. I'm thinking to wrap this PR up with just the DB migration script, and leaving the default traffic rule information UI for the next PR since we're revamping the Router Details view and doing this now might be a waste of effort. wdyt?

krithika369:
Nit: In the PR description it says "The Default Traffic Rule should always be the last in the list". Would be clearer to say it should be evaluated at last and acts as a fallback rule, or something along those lines. Especially because in the UI, it's not the last in the list. :)

Another comment - shall we add a "Default Rule" title to that rule card?

api/e2e/test/08_deploy_router_with_traffic_rules_test.go
```diff
@@ -162,6 +162,13 @@ func TestDeployRouterWithTrafficRules(t *testing.T) {
                           },
                           "is_default": false,
                           "route": "control"
+                        },
```
krithika369, line 165: I'm a bit confused by this. Shouldn't the default rule be activated for this case which only calls the control route?

api/turing/cluster/servicebuilder/router.go
```diff
@@ -503,10 +504,22 @@ func buildFiberConfigMap(
...
 	// a traffic-splitting strategy based on these rules
 	if ver.TrafficRules != nil && len(ver.TrafficRules) > 0 {
+		// TrafficRule struct used requires the name and conditions field to be specified. But
+		// Default Traffic Rule has no name and a hardcoded name can be used instead since
+		// the name field is not used for traffic splitting strategy. Likewise, an empty slice
+		// of conditions can be used for the same reason.
+		rules := append(
+			ver.TrafficRules,
+			&models.TrafficRule{
+				Name:       "default-traffic-rule",
```
krithika369, line 514: It makes sense to have a name as there is no difference between the default and custom rules from the router's POV, for the sake of metrics and logs. This being the case, I think the API & UI should validate that the same name (although extremely unlikely) is not somehow set by the user.

api/turing/models/traffic_rules.go
```diff
@@ -10,7 +10,7 @@ import (
 
 type TrafficRule struct {
 	Name       string                         `json:"name" validate:"required,notBlank"`
-	Conditions []*router.TrafficRuleCondition `json:"conditions" validate:"required,notBlank,dive"`
+	Conditions []*router.TrafficRuleCondition `json:"conditions" validate:"required,dive"`
```
krithika369, line 13: We can now put back the `notBlank` here, no?

api/turing/validation/validator.go
```diff
@@ -129,6 +129,28 @@ func checkTrafficRuleName(sl validator.StructLevel, fieldName string, value stri
...
+	defaultTrafficRuleDescription := strings.Join([]string{
+		"Since 1 or more Custom Traffic rules have been specified,",
+		"a default Traffic rule is required.",
+	}, " ")
+	if defaultTrafficRule == nil {
+		sl.ReportError(defaultTrafficRule, fieldName, "defaultTrafficRule", defaultTrafficRuleDescription, "")
+	}
+}
+
+func checkTrafficRuleConditions(sl validator.StructLevel, fieldName string, rule *models.TrafficRule) {
```
krithika369, line 146: We don't need this anymore if putting back in `notBlank` in `api/turing/models/traffic_rules.go`, right?

api/turing/validation/validator.go
krithika369: I think the data migration is best coupled with the change that introduces the schema changes. If not, when other commits need to be deployed internally and this change is included, some workflows may be broken.

But I understand this PR has already gotten pretty big, so I agree with separating the migration script into another PR, if it will take some time to sort out. Something to keep in mind for the future.

Less concerned about the second bullet on the UI changes. On that note, we also need to update the user docs but that could go with the new UI card.

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -1,9 +1,15 @@
+// import React, { Fragment, useEffect, useState } from "react";
```
krithika369, line 1: This line should be removed.

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -13,17 +19,26 @@ import { get } from "../../../../../components/form/utils";
...
-import { newRule } from "../../../../../services/router/TuringRouter";
+import { newDefaultRule, newRule } from "../../../../../services/router/TuringRouter";
 
-export const RulesPanel = ({ rules, routes, onChangeHandler, errors = {} }) => {
+export const RulesPanel = ({ default_traffic_rule, rules, routes, onChangeHandler, errors = {}, default_traffic_rule_errors = {} }) => {
   const { onChange } = useOnChangeHandler(onChangeHandler);
 
   const onAddRule = () => {
+    // Default rule should only be added when there are custom rules
+    if (rules.length === 0) {
```
krithika369, line 29: Should this be `rules.length > 0 && ! default_traffic_rule` ?

ui/src/components/card/CardHeader.js
krithika369: Shall we name this `DraggableCardHeader` instead to more accurately reflect its use?

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -141,6 +154,16 @@ export const RuleCard = ({
           ))}
         </Fragment>
       </EuiFormRow>
+      : <EuiFormRow
+          label="Conditions *"
+          fullWidth>
+          <EuiCallOut title="Conditions are disabled for default rule." color="warning" iconType="help">
```
krithika369, line 160: Nit: for "the" default rule

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -150,56 +173,68 @@ export const RuleCard = ({
...
-                      options={
-                        idx < rule.routes.length
-                          ? routesOptions(route)
-                          : [noneRoute, ...routesOptions(route)]
-                      }
-                      valueOfSelected={route}
-                      onChange={onChange(`routes.${idx}`)}
+        {[...rule.routes, "_none_"].map((route, idx) => {
+          // Hide last row when all routes are selected
+          if (idx < routes.length) {
```
krithika369, line 178: Can we instead control this by only adding the `__none__` above if `rule.routes.length < routes.length` ? I feel that is more intentional.

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -150,56 +173,68 @@ export const RuleCard = ({
...
-                  ) : (
-                    <EuiIcon type="empty" size="l" />
-                  )}
-                </EuiFlexItem>
-              </EuiFlexGroup>
-              <EuiSpacer size="s" />
-            </Fragment>
-          ))}
+                      error={
+                        Array.isArray(get(errors, `routes.${idx}`))
```
krithika369, line 190: Is it possible it's not an array but an object or primitive value? Or is the `Array.isArray` simply meant to check that the field has an error? Can we instead just check this like `!!get(errors, `routes.${idx}`) ? ...` ?

ui/src/router/components/form/steps/RouterStep.js
```diff
@@ -45,6 +45,8 @@ export const RouterStep = ({ projectId }) => {
       </EuiFlexItem>
       <EuiFlexItem grow={false}>
         <RulesPanel
+          default_traffic_rule={get(data, "config.default_traffic_rule")}
+          default_traffic_rule_errors={get(errors, "config.default_traffic_rule")}
```
krithika369, line 49: For clarity, shall we rename `errors` below to `rules_errors` ?

krithika369 (commented):
Thanks for this PR. The changes largely look good! Left some comments.

api/turing/validation/validator.go
terryyylim: 👍🏻 Aligned with just including DB migration script for this PR and leaving UI changes for the next.

api/api/openapi.bundle.yaml
```diff
@@ -1851,12 +1851,17 @@ components:
       type: string
     RouterVersion:
       example:
+        default_traffic_rule:
+          routes:
+          - routes
```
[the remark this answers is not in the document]
terryyylim: This file is autogenerated using `make bundle-openapi` - https://github.com/caraml-dev/turing/blob/main/api/api/Makefile#L9, these values are similar to other parts of the file where array values are expected.

api/turing/models/traffic_rules.go
terryyylim: We can't because we're using this struct in `api/turing/cluster/servicebuilder/router.go` (https://github.com/caraml-dev/turing/pull/232/files#diff-b42915e302303b3cc6b675a5bbbb46862a411038f37f43d836b82037892bbf23R515). If we add `notBlank` here, we won't be able to set an empty conditions slice for Default Traffic Rule.

api/turing/validation/validator.go
terryyylim: Same reason for why we cannot do so - https://github.com/caraml-dev/turing/pull/232#discussion_r964668455.

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
terryyylim: Good catch, I'll be removing this, because essentially for each route selection option's error, it'll always be a string like `At least one route should be attached to the rule`, so the `isArray` check is always false, i.e https://github.com/caraml-dev/turing/blob/main/ui/src/router/components/form/components/router_config/rule_card/RuleCard.js#L164 has never been working correctly 😅 

It is sufficient to have the error propagated from its parent `EuiFormRow` here - https://github.com/caraml-dev/turing/blob/main/ui/src/router/components/form/components/router_config/rule_card/RuleCard.js#L148.

api/db-migrations/000013_add_default_traffic_rule.down.sql
```diff
@@ -0,0 +1,49 @@
...
+),
+tbl3 AS (
+    SELECT
+        id,
+        json_agg(traffic_rule ORDER BY position) FILTER (where traffic_rule IS NOT NULL) traffic_rules_updated
+    FROM tbl2 GROUP BY id
+)
+
+UPDATE router_versions AS t
+SET traffic_rules = (CASE WHEN traffic_rules_updated IS NOT NULL THEN traffic_rules_updated ELSE '[]'::json END)
```
terryyylim, line 45: `CASE WHEN` here is necessary for keeping the existing behaviour for `traffic_rules` column where it's and empty jsonb array `[]` by default.

api/turing/cluster/servicebuilder/router.go
terryyylim: Added validation to both API and UI, screenshot in description!

api/db-migrations/000013_add_default_traffic_rule.down.sql
terryyylim: Just tested the scripts on our internal staging DB, making a note here that currently the column contains `NULL` (no value) and `[]` for traffic_rules column. This script would convert all `NULL` to `[]` which is the expected behaviour since the DB model doesn't use a pointer currently - https://github.com/caraml-dev/turing/blob/main/api/turing/models/router_version.go#L41.

ui/src/router/components/form/components/router_config/RulesPanel.js
terryyylim: Updated the conditions!

api/turing/models/traffic_rules.go
krithika369: Technically, we can add the empty slice in this step right? There is no validation triggered afterwards? AFAIK, the only time this struct is used in validation is during handling of the API call to create / edit router (where we expect things to be `notBlank`). Am I missing something?

krithika369: One of the benefits of splitting the default traffic rule into a new field is also easier separation of the validation. If we are unable to set `notBlank` on that struct because of the step where we merge the rules for the Fiber config, I think we can resort to defining a separate struct for that phase.

api/e2e/test/08_deploy_router_with_traffic_rules_test.go
terryyylim: Good catch, this is caused by incorrect assertion of the test, I've fixed it.

api/turing/models/traffic_rules.go
terryyylim: >There is no validation triggered afterwards? AFAIK, the only time this struct is used in validation is during handling of the API call to create / edit router (where we expect things to be notBlank). 

This is correct and I got what you mean. I confused this field with another, and I've added `notBlank` tag back.

api/turing/validation/validator.go
terryyylim: This function has been removed.

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -13,17 +18,28 @@ import { get } from "../../../../../components/form/utils";
...
-export const RulesPanel = ({ rules, routes, onChangeHandler, errors = {} }) => {
+export const RulesPanel = ({ default_traffic_rule, rules, routes, onChangeHandler, rules_errors = {}, default_traffic_rule_errors = {} }) => {
   const { onChange } = useOnChangeHandler(onChangeHandler);
 
   const onAddRule = () => {
+    // Default rule should only be added when there are custom rules
     onChange("rules")([...rules, newRule()]);
+    // rule length check captures the case when all Traffic rules are first loaded
+    // null check captures the case when all Traffic rules are removed and added again
+    if ((rules.length > 0 && !default_traffic_rule) || default_traffic_rule === null) {
```
krithika369, line 31: > null check captures the case when all Traffic rules are removed and added again

If they are added again, then we still only need to worry about `rules.length > 0 && !default_traffic_rule` right? Would we ever have a scenario where `rules.length === 0 && default_traffic_rule === null` that we want to handle? And if so, why?

ui/src/router/components/form/components/router_config/RulesPanel.js
```diff
@@ -50,17 +77,44 @@ export const RulesPanel = ({ rules, routes, onChangeHandler, errors = {} }) => {
         </Fragment>
       }>
       <EuiFlexGroup direction="column" gutterSize="s">
+        {
+          rules.length > 0 ?
```
krithika369, line 81: Can re-write this as below more concisely, instead of the ternary operator:
```javascript
    {
          rules.length > 0 && (<EuiFlexItem key="default-route"> ... </EuiFlexItem>)
    }
```

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -62,42 +67,53 @@ export const RuleCard = ({
...
+  const routeSelectionOptions = (rule.routes.length < routes.length ? [...rule.routes,  "_none_"]: rule.routes);
+
   return (
     <EuiCard
       className="euiCard--routeCard"
-      title=""
+      title={isDefault ? "Default Traffic Rule" : ""}
       description=""
       textAlign="left">
+      {!isDefault ?
```
krithika369, line 79: Same comment as above for this and other parts of this file - can use `&&` in place of the ternary operator to write the conditional blocks more concisely, where `null` is the result of the else block.

krithika369 (commented):
I'm yet to review the SQL queries, left some small comments on the rest of the changes.

ui/src/router/components/form/components/router_config/RulesPanel.js
terryyylim: `rules.length === 0 && default_traffic_rule === null` - that happens on the first time a Traffic rule is being added. When we first run `onChange("rules")([...rules, newRule()]);` within `onAddRule`, rules that have been passed to the RulesPanel component stays as an empty array, hence `rules.length > 0` will never be true even after clicking the "Add Rule" button.

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,88 @@
...
+    FROM tbl1 LEFT JOIN tbl2 ON tbl1.id = tbl2.id
+),
+tbl4 AS (
+    SELECT id, jsonb_agg(DISTINCT routes) AS routes
+    FROM (
+        SELECT
+            rv.id
+            , r.id as routes
+        FROM router_versions rv
+        JOIN lateral jsonb_to_recordset(routes) as r("id" jsonb) ON true
```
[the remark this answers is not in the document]
krithika369: @terryyylim I had a mistake in the query I shared earlier, updated it ^.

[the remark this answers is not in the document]
terryyylim: Yeah I got the above already 🙏🏻 Trying to simplify the rest now!

ui/src/router/components/form/components/router_config/RulesPanel.js
krithika369: Haha ok.. I have another thought then - if this default rule is created from `onAddRule`, why do we even compare rules length? It's then enough to look at `!default_traffic_rule` only to add the default rule, right?

terryyylim: That's a good point, you're right `!default_traffic_rule` is all that's needed, updated to reflect this.

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,78 @@
...
+    GROUP BY t.id
+),
+exploded_traffic_rules AS (
+    SELECT
+        id
+        , position
+        , elem
+    FROM cleaned_router_versions, jsonb_array_elements(traffic_rules) WITH ordinality arr(elem, position)
+),
+-- Need to join from cleaned_router_versions since exploded_traffic_rules will exclude router versions which have no traffic rules
```
krithika369, line 38: Why do we care about router versions with no traffic rules?

api/api/openapi.bundle.yaml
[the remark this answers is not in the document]
krithika369: Right, but I think we can get around them if we added explicit examples. But ok with addressing this later as it's minor.

api/db-migrations/000013_add_default_traffic_rule.up.sql
krithika369: Overall the `up` query looks ok but perhaps we can further simplify this by focussing only on the records with traffic rules since that's our goal for updating the traffic_rules as well as default_traffic_rule. Could you try that?

I'll have one last look at the up & down sql scripts after this.

terryyylim: Yeah I think we can simplify if we focus only on the records with traffic rules - I'll test it on internal staging DB records as well.

api/db-migrations/000013_add_default_traffic_rule.down.sql
krithika369: > Just tested the scripts on our internal staging DB

I hope you mean a dump of the data from the staging DB and not running the query on the staging DB. 😅

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,75 @@
...
+    ) t
+    GROUP BY t.id
+),
+exploded_traffic_rules AS (
+    SELECT
+        id
+        , position
+        , elem
+    FROM cleaned_router_versions, jsonb_array_elements(traffic_rules) WITH ordinality arr(elem, position)
+    WHERE id IN (SELECT id FROM dangling_routes)
```
terryyylim, line 37: Added a filter to only alter `traffic_rules` if there are dangling routes.

api/db-migrations/000013_add_default_traffic_rule.down.sql
```diff
@@ -0,0 +1,50 @@
...
+    SELECT id, CASE WHEN traffic_rules = 'null' THEN '[]'::jsonb ELSE traffic_rules END AS traffic_rules
+    FROM router_versions
+),
+exploded_traffic_rules AS (
+    SELECT
+        id
+        , position
+        , elem
+    FROM cleaned_router_versions, jsonb_array_elements(traffic_rules) WITH ordinality arr(elem, position)
+    WHERE id IN (SELECT id FROM dangling_routes)
```
terryyylim, line 19: Added a filter to only alter traffic_rules if there are dangling routes.

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,75 @@
...
+        , CASE
+            WHEN elem -> 'routes' IS NOT NULL THEN jsonb_build_object(
+                'conditions', elem -> 'conditions',
+                'routes', (elem -> 'routes' || dangling_routes),
+                'name', elem -> 'name'
+            )
+            ELSE NULL
+            END AS traffic_rule
+    FROM
+        exploded_traffic_rules
```
terryyylim, line 52: Removed `LEFT JOIN` from cleaned_router_versions since we have `id` column and only need to alter affected rows, i.e those with dangling routes

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,75 @@
...
+        LEFT JOIN dangling_routes ON exploded_traffic_rules.id = dangling_routes.id
+),
+joined_updated_traffic_rules AS (
+    SELECT
+        id,
+        json_agg(traffic_rule ORDER BY position) traffic_rules_updated
```
terryyylim, line 58: Removed `FILTER` since there will be no `NULL` values since we only alter affected rows.

terryyylim (commented):
@krithika369 , summarised the latest DB script changes as described below.

api/db-migrations/000013_add_default_traffic_rule.down.sql
terryyylim: 😂 Yep it was ran locally using a DB dump!

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,75 @@
...
+    FROM exploded_traffic_rules_with_dangling_routes GROUP BY id
+)
+
+UPDATE router_versions AS t 
+SET default_traffic_rule = jsonb_build_object('routes', dangling_routes),
```
krithika369, line 63: Based on the above comments, this `UPDATE` can use a `CASE...WHEN` - you're already left joining with `dangling_routes`; if the right table cols are empty, set `default_traffic_rule` to a list of all route names (aggregate from `original_route_tbl`).

Similarly, may be you can set `traffic_rules = traffic_rules_updated` if there are dangling routes and just `traffic_rules = traffic_rules` otherwise.

api/db-migrations/000013_add_default_traffic_rule.down.sql
krithika369: This makes sense for the down query. 👍🏽

api/db-migrations/000013_add_default_traffic_rule.down.sql
```diff
@@ -0,0 +1,50 @@
+WITH dangling_routes AS (
+    SELECT
+        id
+        , default_traffic_rule -> 'routes' AS routes
+    FROM
+        router_versions
+    WHERE default_traffic_rule -> 'routes' IS NOT NULL
+),
+cleaned_router_versions AS (
+    SELECT id, CASE WHEN traffic_rules = 'null' THEN '[]'::jsonb ELSE traffic_rules END AS traffic_rules
```
krithika369, line 10: Also here, I think we can skip the cleaning step. No harm having it as the dangling routes filter below will ignore these kinds of too but could simplify the query a little by leaving out this step.

api/db-migrations/000013_add_default_traffic_rule.up.sql
terryyylim: >Similarly, may be you can set traffic_rules = traffic_rules_updated if there are dangling routes and just traffic_rules = traffic_rules otherwise.

I think this is not required anymore since `exploded_traffic_rules_with_dangling_routes` handles this with the `CASE WHEN...` for the routes field.

terryyylim:
I've updated the PR description with the conditions that the DB scripts check for/does.

api/db-migrations/000013_add_default_traffic_rule.up.sql
```diff
@@ -0,0 +1,90 @@
...
+        id,
+        json_agg(traffic_rule ORDER BY position) traffic_rules_updated
+    FROM exploded_traffic_rules_with_dangling_routes GROUP BY id
+)
+
+UPDATE router_versions AS t 
+SET default_traffic_rule = 
+    CASE
+        WHEN dangling_routes IS NOT NULL THEN jsonb_build_object('routes', dangling_routes)
+    ELSE jsonb_build_object('routes', all_routes) END,
```
krithika369, line 76: Is this correct? 🙃

terryyylim: For Traffic rules with no dangling routes, we'll set default_traffic_rule = all routes and for Traffic rules with dangling routes, we'll set default_traffic_rule = only dangling routes right? 😮

krithika369: Ok, it's actually correct. I was confused because `all_routes` is built from the `rule_route_tbl` (I imagined it should be taken from the `routes` column of `router_versions` table, or `route_tbl` in this case). But actually, for the condition under which they are used, they are equal. So it's fine. But may be good to clarify with a comment.

terryyylim: 👌🏻 Yeap no difference, but its clearer to use `route_tbl`, I've updated the query.

api/db-migrations/000013_add_default_traffic_rule.down.sql
```diff
@@ -0,0 +1,65 @@
...
+    SELECT id, route -> 'id' AS route_id FROM route_tbl
+),
+-- all routes for each router version
+all_routes AS (
+    SELECT
+        id, jsonb_agg(DISTINCT route_id) routes
+    FROM original_route_tbl
+    GROUP BY id
+),
+dangling_routes AS (
```
krithika369, line 17: For the down query the `all_routes` is `all_rule_routes` right? Meaning, we have to only take the routes that appear in the rules. Not all routes :O

krithika369 (approved):
Left a comment on `api/db-migrations/000013_add_default_traffic_rule.down.sql`. The rest LGTM.

Thanks for the extensive work on the data migration! Please go ahead and merge once the comment has been addressed. 🎉

api/db-migrations/000013_add_default_traffic_rule.down.sql
terryyylim: `all_routes` include all the routes for each router version. When we run the `EXCEPT` clause, it only selects from the `default_traffic_rule_routes` and exclude those where all routes match that particular row's default traffic rule's routes. This is just to capture the case where all routes have been used in `traffic_rules` column, and that there are no  dangling routes to be removed.

krithika369: > When we run the EXCEPT clause, it only selects from the default_traffic_rule_routes and exclude those where all routes match that particular row's default traffic rule's routes

I'm still confused. The `default_traffic_rule_routes` will always be a subset of `all_routes` which means `dangling_routes` will always be empty, no?

And this part below can be done earlier:
```
-- WHERE clause filters away rows with no traffic rules
    WHERE t.routes IS NOT NULL
```

The goal for the down query, wrt to updating the traffic_rules column in:
- If len(default_traffic_rule_routes) == len(all_routes) for a given id
- Else remove the default_traffic_rule_routes from each rule ? Now, on hindsight, is this even correct? What if the user created the `default_traffic_rule` with their own combination of routes?

This makes me think it's better if we don't touch the `traffic_rules` column at all during the down query - no harm keeping the routes in the rules instead of dangling, right?

terryyylim: >The default_traffic_rule_routes will always be a subset of all_routes which means dangling_routes will always be empty, no?

Yes, its always a subset, but what we're excluding here is a subset but **without equality**, which points to only router versions where all routes have been used in `traffic_rules` and we're setting all routes for `default_traffic_rule`.

>If len(default_traffic_rule_routes) == len(all_routes) for a given id

Yes, `If len(default_traffic_rule_routes) == len(all_routes) for a given id`, we won't do anything.

>What if the user created the default_traffic_rule with their own combination of routes?

Then this query is wrong. We're not capturing that case, cause there's no way to. I see this rollback used mainly in the event of rolling out only.

krithika369: > Yes, its always a subset, but what we're excluding here is a subset but without equality, which points to only router versions where all routes have been used in traffic_rules and we're setting all routes for default_traffic_rule.

Not sure I follow. If `default_traffic_rule_routes` had [{1, "route-1"}] and `all_routes` had [{1, "route-1"}, {1, "route-2"}], what is the result of `dangling_routes` ?

terryyylim: >Not sure I follow. If default_traffic_rule_routes had [{1, "route-1"}] and all_routes had [{1, "route-1"}, {1, "route-2"}], what is the result of dangling_routes

`dangling_routes` will be empty in this case.

krithika369: So what's the point of this query? If `default_traffic_rule_routes` is a subset of `all_routes`, this EXCEPT query will always result in an empty table, no? It's like set operation A - B.

krithika369: > Sorry, my typo above, it should be with equality - dangling_routes will be empty in this case.

Hmm.. Can you add an example of what you mean by "with equality" and "without" ?

terryyylim: >So what's the point of this query? If default_traffic_rule_routes is a subset of all_routes, this EXCEPT query will always result in an empty table, no? It's like set operation A - B.

It'll be able to track rules with dangling routes and revert in cases like this:
`default_traffic_rule_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2"]
 ```

`all_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2", "route-3"]
 ```

`dangling_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2"]
 ```

Then, `route-1` and `route-2` will be removed from all traffic_rules' routes.

Comment (terryyylim) [reply api/db-migrations/000013_add_default_traffic_rule.down.sql:17]: >Hmm.. Can you add an example of what you mean by "with equality" and "without" ?

**with equality**
`default_traffic_rule_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2", "route-3"]
 ```

`all_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2", "route-3"]
 ```

**without equality**
`default_traffic_rule_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2"]
 ```

`all_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2", "route-3"]
 ```

Comment (krithika369) [reply api/db-migrations/000013_add_default_traffic_rule.down.sql:17]: What about:

`default_traffic_rule_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-1", "route-2", "route-3"]
```

and 

`all_routes`
```
 id  |       routes
------+---------------------
 1 | ["route-2", "route-1", "route-3"]
```
?

Isn't it simpler just to compare the lengths (`len(default_traffic_rule_routes) == len(all_routes)` ) ?

terryyylim: I've updated the query, lmk if it makes sense 🙏🏻

PR #234, opened 2022-09-08
[API/SDK] Expose user-configurable autoscaling parameters

krithika369:
This PR adds the API / SDK changes for the configuration of autoscaling parameters by the users, on the Enricher / Router / Ensembler. The `metric` and `target` [configurations](https://knative.dev/docs/serving/autoscaling/autoscaling-metrics/) supported by Knative are now exposed and accordingly, the correct Knative Autoscaling class is applied by the API.

### Backward-compatibility
All along, the API has used `{"metric": "concurrency", "target": "1"}` as the autoscaling configuration for all resources.
* The existing data in the DB has been migrated using the same default.
* The SDK will apply the same defaults, if not provided by the user.
* The API treats the autoscaling configurations to be optional and applies the same defaults in `api/turing/api/request/request.go`. This is to support backward compatibility on the SDK, until all users have migrated to the latest version (a TODO note has been added in the API accordingly). This also makes the API changes compatible with the existing UI, until those changes are added.

### Changes
* `.pre-commit-config.yaml` - Add hook for automatically creating OpenAPI bundle when the API specs change

#### API
* `api/api/specs/routers.yaml` - Add autoscaling config to all relevant entities
* `api/api/specs/routers.yaml` - Add DB migrations
* `api/e2e/test/testdata/create_router_version.json.tmpl` - Update an e2e test case to explicitly set the autoscaling props. Similar changes are applied to `sdk/e2e/03_create_router_version_test.py` which is the corresponding SDK test.
* `api/turing/api/request/request.go` - Process autoscaling configs in the API or apply default if unset
* `api/turing/cluster/knative_service.go` - Propagate properties to the Knative service. Set the correct autoscaling class according to the metrics.
* `api/turing/config/config.go` - The previous deploy-time config for target concurrency is no longer required. Its handling is now removed elsewhere in the code too.
* `api/turing/testdata/cluster/servicebuilder/` - Updated testcases to try different autoscaling policies
* `api/turing/validation/validator_test.go` - Added validation tests for autoscaling properties

#### SDK
* `sdk/turing/router/config/autoscaling_policy.py` - Add autoscaling config
* Updated Enricher, Docker/Pyfunc ensembler configs and Router to handle it the autoscaling properties. Initialization of these configs all apply the autoscaling defaults if not supplied.
* Updated unit tests / e2e test case to cover the new configs


api/turing/models/enricher.go
```diff
@@ -6,17 +6,19 @@ type Enricher struct {
 	Model
 	// Fully qualified docker image string used by the enricher, in the
 	// format registry/repository:version.
-	Image string `json:"image" validate:"required"`
```
krithika369, line 9: Cleaning up unused validate tags and the validation is done on the `EnricherEnsemblerConfig` struct declared in the request package instead of this.

api/turing/api/request/request.go
```diff
@@ -22,12 +22,13 @@ type CreateOrUpdateRouterRequest struct {
...
-	TrafficRules     models.TrafficRules     `json:"rules" validate:"unique=Name,dive"`
-	ExperimentEngine *ExperimentEngineConfig `json:"experiment_engine" validate:"required,dive"`
-	ResourceRequest  *models.ResourceRequest `json:"resource_request"`
-	Timeout          string                  `json:"timeout" validate:"required"`
+	Routes            models.Routes             `json:"routes" validate:"required"`
+	DefaultRouteID    *string                   `json:"default_route_id"`
+	TrafficRules      models.TrafficRules       `json:"rules" validate:"unique=Name,dive"`
+	ExperimentEngine  *ExperimentEngineConfig   `json:"experiment_engine" validate:"required,dive"`
+	ResourceRequest   *models.ResourceRequest   `json:"resource_request"`
+	AutoscalingPolicy *models.AutoscalingPolicy `json:"autoscaling_policy" validate:"omitempty,dive"`
```
krithika369, line 30: `omitempty` is for backward compatibility. Can we removed when `getAutoscalingPolicyOrDefault` below is removed.

sdk/turing/router/config/autoscaling_policy.py
```diff
@@ -0,0 +1,41 @@
+from dataclasses import dataclass, field
+from heapq import merge
```
terryyylim, line 2: This import can be removed since it's not used anywhere.

sdk/turing/router/config/autoscaling_policy.py
```diff
@@ -0,0 +1,41 @@
...
+
+import turing.generated.models
+from turing.generated.model_utils import OpenApiModel
+
+
+@dataclass
+class AutoscalingPolicy:
```
terryyylim, line 9: Would be nice to add some docstrings to the `AutoscalingPolicy` class, e.g to describe supported values for metric parameter. Even better, we could introduce Enums here?

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -315,12 +351,14 @@ def __init__(
...
         env: List["EnvVar"],
         service_account: str = None,
+        autoscaling_policy: AutoscalingPolicy = DEFAULT_AUTOSCALING_POLICY,
     ):
         """
         Method to create a new Docker ensembler
 
         :param image: registry and name of the image
         :param resource_request: ResourceRequest instance containing configs related to the resources required
+        :param autoscaling_policy: AutoscalingPolicy instance containing configs for the deployment autoscaling
```
terryyylim, line 361: nit: `...for autoscaling the deployment`?

terryyylim (commented):
Thanks for the PR, @krithika369 ! Left some small comments to address 🙏🏻

api/turing/cluster/knative_service.go
```diff
@@ -91,14 +92,15 @@ func (cfg *KnativeService) buildSvcSpec(
...
+	timeout := DefaultRequestTimeoutSeconds
 
 	// Build annotations, set target concurrency of 1
 	annotations := map[string]string{
 		"autoscaling.knative.dev/minScale": strconv.Itoa(cfg.MinReplicas),
 		"autoscaling.knative.dev/maxScale": strconv.Itoa(cfg.MaxReplicas),
-		"autoscaling.knative.dev/target":   strconv.Itoa(cfg.TargetConcurrency),
-		"autoscaling.knative.dev/class":    knativeSvcDefaults.AutoscalingClass,
+		"autoscaling.knative.dev/metric":   cfg.AutoscalingMetric,
+		"autoscaling.knative.dev/target":   cfg.AutoscalingTarget,
```
leonlnj, line 102: not related to the MR, CMMIW, the [default in knative ](https://knative.dev/docs/serving/autoscaling/concurrency/#soft-limit)for this is 100 but we are setting to 1 only, any reason or am I understanding something wrong?

api/turing/api/request/request.go
leonlnj: any reason to remove the default? just thinking what if user don't know the difference, is this deliberate to force them to understand how they want to scale?

leonlnj:
did we install [HPA extension](serving-hpa) previously somewhere?

leonlnj (approved):
thanks! LGTM. I learned lots reading through the knative docs.

api/turing/api/request/request.go
krithika369: The default is set by the clients - Turing UI and SDK, if the user doesn't explicitly supply or change these values. So, the API will expect them to be there.

api/turing/cluster/knative_service.go
krithika369: The default value of 1 that we use is inspired by kserve's default to create the underlying knative resources - it will allow each replica to only handle 1 request at a given time (which may be an overkill but it's a safe value).

sdk/turing/router/config/autoscaling_policy.py
krithika369: Done!

krithika369: Done! Thanks for the suggestion. :)

krithika369:
> did we install [HPA extension](serving-hpa) previously somewhere?

@leonlnj This is a great point - thanks for raising. We have installed it internally but the public `cluster-init` setup doesn't have it. I will be creating another PR for the helm chart update. I will include this change there.

PR #235, opened 2022-09-08
Configure Turing UI to use Standard Ensembler for Custom Experiment Engines

deadlycoconuts:
## Context
With the changes introduced in #237 and #233 to support the usage of Standard Ensemblers with Custom Experiment Engines, the Turing UI has been updated in this PR to consume Standard Ensembler components exposed by the UI of these custom engines. 

There are two new components that Turing UI will now consume - `EditStandardEnsemblerConfig` and `StandardEnsemblerConfigDetails`. As their names suggest, they are used respectively for the create/edit router form and the router details view respectively. 

Screenshot 2022-09-21 at 5 57 49 PM

Changes are made to the existing `StandardEnsemblerFormGroup` and `EnsemblerConfigSection` components to either display the regular treatment mappings panel/table if the configured experiment engine is of the `standard` type, and an exposed UI component otherwise. In either case, the fallback route panel/config section will always be displayed in both scenarios.

## Main Modifications
- `ui/src/components/ensembler/StandardEnsemblerLoaderComponent.js` - Addition of a new loader for the Standard Ensembler components
- `ui/src/router/components/configuration/components/EnsemblerConfigSection.js` - Refactoring of the child components to be displayed when a Standard Ensembler is selected in a single `StandardConfigViewGroup` component
- `ui/src/router/components/configuration/components/standard_config_section/StandardConfigViewGroup.js` - Addition of a new component that displays either the treatment mappings view or the exposed component, depending on the experiment engine type
- `ui/src/router/components/form/components/ensembler_config/standard_ensembler/StandardEnsemblerFormGroup.js` - Addition of extra logic to display the exposed UI component if a custom experiment engine is configured
- `ui/src/router/components/form/components/ensembler_config/typeOptions.js` - Addition of a new description for the standard ensembler if a custom experiment engine is configured
- `ui/src/services/ensembler/StandardEnsembler.js` - Addition of a new `route_name_path` field to the standard ensembler object


api/api/openapi.bundle.yaml
```diff
@@ -2299,8 +2304,12 @@ components:
           items:
             $ref: '#/components/schemas/EnsemblerStandardConfig_experiment_mappings'
           type: array
+        route_name_path:
+          pattern: ^\w+(?:\.\w+)*$
```
deadlycoconuts, line 2308: This regex pattern here allows only period-delimited words of letters (or underscores) to be accepted as a valid `route_name_path`. Specifying this here also dispenses the need for manually crafting regex checks on the API for this field.

ui/src/components/page_navigation/PageNavigation.js
```diff
@@ -85,7 +85,6 @@ const MoreActionsButton = ({ actions }) => {
       panelPaddingSize="none"
       anchorPosition="downRight">
       <EuiContextMenuPanel
-        hasFocus={false}
```
deadlycoconuts, line 88: Removed this because `hasFocus` is no longer a prop for `EuiContextMenuPanel` (see https://elastic.github.io/eui/#/navigation/context-menu).

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -14,13 +14,26 @@ const typeOptions = [
   {
     value: "standard",
     inputDisplay: "Standard",
+    expenginetype: "standard",
```
deadlycoconuts, line 17: Added this field so the function `ensemblerTypeOptions` is able to select one of the two `typeOptions` that both have the `value` "standard".

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -62,11 +75,10 @@ export const ensemblerTypeOptions = (engineProps) => {
     return typeOptions.filter((o) => o.value !== "standard");
   }
   // Ensembler must be selected when there is an experiment engine
-  const ensemblerOptions = typeOptions.filter((o) => o.value !== "nop");
-  if (
-    !engineProps?.standard_experiment_manager_config
-      ?.experiment_selection_enabled
-  ) {
+  const ensemblerOptions = typeOptions.filter(
+    (o) => o.value !== "nop" && (o.expenginetype === engineProps.type || o.expenginetype === undefined)
```
deadlycoconuts, line 79: This additional condition selects the standard ensembler `typeOption` with the correct `expenginetype`. If a `typeOption` does not have the `expenginetype` field set (which implicitly means it's not a standard ensembler), then it will not be removed.

Note: Just noticed the 2 subconditions here `(o.expenginetype === engineProps.type || o.expenginetype === undefined)` can be inverted for slightly better computation 😅

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -62,11 +75,10 @@ export const ensemblerTypeOptions = (engineProps) => {
...
+  );
+  if (engineProps?.standard_experiment_manager_config?.experiment_selection_enabled === false) {
```
deadlycoconuts, line 81: This new condition (compared with `(!engineProps?.standard_experiment_manager_config?.experiment_selection_enabled)` previously) ensures that the experiment manager config has to explicitly disallow experiment selection before we remove the standard ensembler option (i.e. standard ensemblers will be selectable with custom experiment engines).

This contrasts with the previous condition which can return true whenever the `engineProps` does not have have the `standard_experiment_manager_config.experiment_selection_enabled` path defined (i.e. standard ensemblers are not selectable with custom experiment engines).

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -14,13 +14,26 @@ const typeOptions = [
...
+  {
+    value: "standard",
+    inputDisplay: "Standard",
+    expenginetype: "custom",
+    description: (
+      <Fragment>
+        Turing will select the route response corresponding to the route name specified
+        in the treatment configuration. The route name path will be used to locate the
+        name of the route within the treatment configuration.
+      </Fragment>
```
deadlycoconuts, line 34: Added a separate standard ensembler description for custom experiment engines.

ui/src/components/ensembler/StandardEnsemblerLoaderComponent.js
```diff
@@ -0,0 +1,57 @@
...
+  useEffect(() => {
+    setReady(ready);
+    setFailed(failed);
+  }, [setReady, setFailed, ready, failed]);
+
+  return null;
+};
+
+// Dynamic Script Loading component wrapper
+export const StandardEnsemblerLoaderComponent = ({
```
terryyylim, line 20: I think there's room to combine this component with the existing `ExperimentEngineLoaderComponent`. The only real difference is the props naming for `remoteUi` vs. `experimentEngine`, even the `LoadDynamicScript` is the same. Perhaps, we can shift this file to `src/components/remote_component` now that it is used in both ensembler and experiments' related components?

ui/src/router/components/configuration/components/standard_config_section/StandardConfigViewGroup.js
```diff
@@ -0,0 +1,95 @@
...
+  type
+}) => {
+  const { getEngineProperties, isLoaded } = useContext(ExperimentEngineContext);
+
+  const engineProps = getEngineProperties(type);
+
+  return (
+    !!standardConfig && (
+      <EuiFlexGroup direction="column">
+        <>
```
terryyylim, line 60: I think this is not required.

ui/src/router/components/form/components/ensembler_config/typeOptions.js
terryyylim, line 17: nit: camelCase here so `expEngineType`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
terryyylim, line 81: Would `!(engineProps?.standard_experiment_manager_config?.experiment_selection_enabled)` suffice?

terryyylim (changes requested):
LGTM on the whole, left some comments to address, thanks @deadlycoconuts !

ui/src/router/components/configuration/components/standard_config_section/StandardConfigViewGroup.js
deadlycoconuts: Thanks for catching that 😅

ui/src/router/components/form/components/ensembler_config/typeOptions.js
deadlycoconuts: Hmm the reason why I left this as entirely lowercase was because the items in `typeOptions` do not get the `expEngineType` field stripped off when `typeOptions` gets passed into `EuiSuperSelect` as a prop, since `expEngineType` is a completely custom field not recognised by EUI. What then happens is that these items get rendered as DOM elements, with the `expEngineType` field being treated as a prop and causing this warning: 

Screenshot 2022-09-23 at 11 17 18 AM

An alternative to avoiding this is to consciously strip the `typeOptions` of the `expEngineType` field every single time they are being used but I find this harder to maintain and to enforce.

ui/src/router/components/form/components/ensembler_config/typeOptions.js
deadlycoconuts: Nope, a custom experiment engine would not have the `standard_experiment_manager_config` field set, which will cause the condition `!(engineProps?.standard_experiment_manager_config?.experiment_selection_enabled)` to evaluate as `!(undefined)`  which gives `true`, causing the standard ensembler to be removed? 🤔 It isn't much different from the original `(!engineProps?.standard_experiment_manager_config?.experiment_selection_enabled)` right?

ui/src/services/job/JobStatus.js
```diff
@@ -3,22 +3,22 @@ import { Enum, EnumValue } from "../enum/Enum";
 export const JobStatus = Enum({
   PENDING: EnumValue("pending", {
     label: "Pending",
-    color: "#FEC514",
+    color: "warning",
```
deadlycoconuts, line 6: Changing these colours to use the default EUI colour template wherever possible.

ui/src/services/status/Status.js
```diff
@@ -3,17 +3,17 @@ import { EnumValue, Enum } from "../enum/Enum";
...
   FAILED: EnumValue("failed", {
     label: "Failed",
-    color: "#BD271E",
+    color: "danger",
     iconType: "cross",
   }),
   PENDING: EnumValue("pending", {
     label: "Updating",
-    color: "#FEC514",
+    color: "warning",
```
krithika369, line 16: Thanks for changing these. What about the "Not Deployed" status? Shall we change it to `default` ? And any previous occurrence using `subdued` to `default` as well? https://elastic.github.io/eui/#/display/badge

ui/src/components/ensembler/StandardEnsemblerLoaderComponent.js
deadlycoconuts: Okay done, now both of those components got reduce to a single `RemoteLoaderComponent` :)

ui/src/services/status/Status.js
deadlycoconuts: Hmm `default` isn't as dark as `#6A717D` but I guess that works too :) I'll replace all the old subdued values with `default` then :)

ui/src/components/remote_component/RemoteLoaderComponent.js
```diff
@@ -17,9 +17,10 @@ const LoadDynamicScript = ({ url, setReady, setFailed }) => {
 };
 
 // Dynamic Script Loading component wrapper
-export const ExperimentEngineLoaderComponent = ({
+export const RemoteLoaderComponent = ({
```
krithika369, line 20: Nit: I think it's ok to keep this component as it was, except may be make the `componentName` configurable. After all, the std ensembler UI is tied to the experiment engine and it's the engine whose configs we are loading.

If it helps, we can rename it from `ExperimentEngineLoaderComponent` -> `ExperimentEngineComponentLoader` :D

ui/src/router/components/form/components/ensembler_config/typeOptions.js
krithika369: Do we _have to_ introduce this? The existing description seems generic enough to conceptually cover both cases, no?

> Turing will select the response from one of the routes, based on the configured mapping between routes and experiment treatments

krithika369: It may also help simplify some of the implementation (like https://github.com/caraml-dev/turing/pull/235/files#r976051034)

ui/src/components/remote_component/RemoteLoaderComponent.js
deadlycoconuts: Wait so do you mean we should keep both components separate, i.e. `StandardEnsemblerLoaderComponent` and `ExperimentEngineLoaderComponent`, or combining them as @terryyylim had suggested? 😅

ui/src/router/components/form/components/ensembler_config/typeOptions.js
deadlycoconuts: Well, I was originally thinking that "configured mapping between routes and experiment treatments" doesn't accurately describe whatever we're going to display in the standard ensembler configuration panel, i.e. the "assigned routes" and "assigned experiments" (there's no mention of "treatment" anywhere). 

But technically, since we are expecting the experiment engines to implement the standard ensembler component to be consumed, I guess it makes sense that we should probably keep this description as generic as possible, and then expect more tooltips/information within the consumed component to explain what's going on.

ui/src/router/components/form/components/ensembler_config/typeOptions.js
terryyylim: I see, thanks for clarification! The current approach is cleaner than consciously stripping the `typeOptions` value, so let's keep it this way.

terryyylim (commented):
Thanks @deadlycoconuts , I'll leave @krithika369 to approve this PR since there's the rename comment yet to be resolved.

ui/src/components/remote_component/RemoteLoaderComponent.js
krithika369: I just noticed the older comment. I do think it makes sense to combine the way you have it. But I don't think this should be called a generic `RemoteLoaderComponent` because the logic inside is still looking at getting some url, configs (which are coupled to the experiment engine). So, the implementation can remain as is, but I think it still makes sense to call it something related to experiment engine - hence my suggestion is to use `ExperimentEngineComponentLoader` as the name.

ui/src/router/components/form/components/ensembler_config/typeOptions.js
krithika369: Yes, I think we can keep this description generic. Any additional details specific to the std / custom engines can be embedded within those components (like the Flyout we have for XP's remote component for the std ensembler). That way we can do away with the special handling based on `expenginetype` introduced in this file.

deadlycoconuts: Okay done, I've removed the standard ensembler option for custom experiment engines so it's now the same as before :)

ui/src/components/remote_component/RemoteLoaderComponent.js
deadlycoconuts: Ok! I've just renamed the component as you have suggested!

deadlycoconuts:
Thanks a lot for all the comments everyone! I've added some short descriptions to the docs about standard ensemblers with custom experiment engines. Besides that there shouldn't be other changes besides those that are meant to address the comments above.

docs/how-to/create-a-router/configure-ensembler.md
```diff
@@ -25,6 +29,19 @@ In addition, a fallback route may be configured whose results will be used at ru
...
+For routers configured with Custom Experiment Engines, the router will return a response from one of the routes 
+corresponding to the route name that is found within the treatment configuration returned by the experiment 
+engine. At run time, the router will attempt to access the route name within the treatment configuration received via a 
+user-configured path. 
+
+A fallback route also has to be configured in order to capture cases whereby the route name found in the treatment 
+configuration does not correspond to any of the routes configured, or when the user-configured path is invalid with 
+respect to the treatment configuration received.
+
+The UI for Standard Ensemblers with Custom Experiment Engines is consumed directly as a remote component that is
```
krithika369, line 42: This line is better removed from the user docs - is it more of a dev doc detail.

krithika369 (approved):
The docs LGTM. Thanks for this!

terryyylim (approved):

deadlycoconuts:
Thanks a lot for the comments everyone! I'll be merging this right away! 🚀🚀

PR #237, opened 2022-09-09
Apply new Standard Ensembler routing policy with Custom Experiment Engines 

deadlycoconuts:
## Context
With the introduction of changes to the routing strategy of the Turing Router to support Standard Ensemblers for Custom Experiment Engines in #233, this PR introduces follow-up modifications to allow the API to set up Turing Routers that use a Standard Ensembler and a Custom Experiment Engine. The API also performs additional validation checks to ensure that any router create/update request payload does not contain both `experiment_mappings` and `route_name_path` specified in the `EnsemblerStandardConfig`.

## Minor Fixes
This PR also includes the replacement of [deprecated](https://pkg.go.dev/io/ioutil#section-documentation) `io/util` functions from `go 1.16/1.17` with their corresponding `io` or `os` counterparts:
- `ioutil.NopCloser` -> `io.NopCloser`
- `ioutil.ReadAll` -> `io.ReadAll`
- `ioutil.ReadFile` -> `os.ReadFile`
- `ioutil.TempFile` -> `os.CreateTemp`
- `ioutil.WriteFile` -> `os.WriteFile`

## Main Modifications
- `api/api/specs/routers.yaml` - Addition of the `route_name_path` field to `EnsemblerStandardConfig`
- `api/turing/api/request/request.go` - Addition of validation step to ensure `ExperimentMappings` and `RouteNamePath` are not both set at the same time
- `api/turing/cluster/servicebuilder/router.go` - Addition of step to add `route_name_path` to the config map that will be used by the Turing Router upon deployment
- `api/turing/models/ensembler.go` - Addition of `RouteNamePath` to the db model `EnsemblerStandardConfig` struct

## Future Changes (Not included in this PR)
- Addition of the `route_name_path` field to Standard Ensemblers within the Turing SDK
- Addition of the new e2e test (api/e2e/test/09_deploy_router_with_route_name_path_std_ensembler_test.go) as a Turing SDK e2e test


api/api/openapi.bundle.yaml
```diff
@@ -2299,8 +2304,12 @@ components:
           items:
             $ref: '#/components/schemas/EnsemblerStandardConfig_experiment_mappings'
           type: array
+        route_name_path:
+          pattern: ^\w*(?:\.\w+)*$
+          type: string
```
deadlycoconuts, line 2309: This regex pattern here allows only period-delimited words of letters (or underscores) to be accepted as a valid `route_name_path`. Specifying this here also dispenses the need for manually crafting regex checks on the API for this field.

api/api/specs/routers.yaml
```diff
@@ -797,6 +798,9 @@ components:
                 description: "route id of the routes configured in the router"
                 type: "string"
                 example: "route-1"
+        route_name_path:
+          type: "string"
+          pattern: '^\w*(?:\.\w+)*$'
```
deadlycoconuts, line 803: This regex pattern here allows only period-delimited words of letters (or underscores) to be accepted as a valid `route_name_path`. Specifying this here also dispenses the need for manually crafting regex checks on the API for this field.

api/turing/testdata/cluster/servicebuilder/router_version_missing_bigquery.json
```diff
@@ -54,21 +54,24 @@
...
-        },
-        "endpoint": "echo?delay=10ms",
-        "timeout": "2s",
-        "port": 8080,
-        "env": [
-            {
-                "name": "TEST_ENV",
-                "value": "ensembler"
-            }
-        ]
```
deadlycoconuts, line 72: Updated the schema to reflect the current schema of an ensembler. The previous schema was incorrect but it didn't break anything because another (earlier) section of the router config was causing the test to fail as expected.

api/turing/api/request/request_test.go
```diff
@@ -422,6 +422,75 @@ func TestRequestBuildRouterVersionWithDefaults(t *testing.T) {
...
+
+	// Set up mock Ensembler service
+	ensemblerSvc := &mocks.EnsemblersService{}
+
+	// Define tests
+	tests := map[string]struct {
+		testEnsembler models.Ensembler
+		err           string
+	}{
+		"failure | both experiment mappings and route name path are set": {
```
leonlnj, line 457: maybe a positive to test case to ensure it is passing when expected to be?

api/api/specs/routers.yaml
krithika369: Oooh what if the nested path actually has a key with "-" in it? Shall we remove this check and keep it flexible like the `TrafficRuleCondition` ?

api/api/specs/routers.yaml
```diff
@@ -796,6 +796,7 @@ components:
       nullable: true
       required:
```
krithika369, line 797: If only one of the 2 props is supposed to be set, should we remove both from `required`?

api/turing/models/ensembler.go
```diff
@@ -33,6 +33,7 @@ const (
 
 type EnsemblerStandardConfig struct {
 	ExperimentMappings []ExperimentMapping `json:"experiment_mappings" validate:"required,dive"`
+	RouteNamePath      string              `json:"route_name_path"`
```
krithika369, line 36: The `ExperimentMappings` should no longer have `validate:"required"`, right? Just `dive` ?

krithika369 (commented):
On the whole, the PR looks good! Thanks for meticulously updating all the test scenarios too. I left some small comments. Will have another quick look once they are reviewed / addressed.

api/api/specs/routers.yaml
deadlycoconuts: Hmm if it's just dashes we can make it `^[\w\-]*(?:\.[\w\-]+)*$` instead? Are you referring to `field` within `TrafficRuleCondition`? That one doesn't have any checks right?

api/api/specs/routers.yaml
deadlycoconuts: Hmm that's true, I was thinking of keeping both `required` and letting requests to the API use `null` values for an array/string (`[]` and `""`) to denote the empty field but I guess the reverse works too :)

api/turing/validation/validator.go
```diff
@@ -46,6 +49,24 @@ func NewValidator(expSvc service.ExperimentsService) (*validator.Validate, error
 	return instance, nil
 }
 
+func validateEnsemblerStandardConfig(sl validator.StructLevel) {
```
deadlycoconuts, line 52: Had to go for a `StructLevel` validation in the end, which churns out error logs with the `CreateOrUpdateRouterRequest.Config.Ensembler.StandardConfig.[...]` path instead of the `FieldLevel` validation option as discussed earlier. Apparently the go-playground validator is [unable](https://github.com/go-playground/validator/issues/470) to perform `FieldLevel` validations on structs, even with a custom validation function defined.

This isn't too much of an issue, except for the enforced error logging path, which indirectly forces us to duplicate the error messages returned e.g. instead of returning "EnsemblerStandardConfig cannot have both ExperimentMappings and RouteNamePath set" with the `CreateOrUpdateRouterRequest.Config.Ensembler.StandardConfig` path, we can only return "ExperimentMappings cannot be set when RouteNamePath is set" with the `CreateOrUpdateRouterRequest.Config.Ensembler.StandardConfig.ExperimentMappings` path AND "RouteNamePath cannot be set when ExperimentMappings is set" with the `CreateOrUpdateRouterRequest.Config.Ensembler.StandardConfig.RouteNamePath` path.

api/turing/cluster/servicebuilder/router.go
```diff
@@ -491,7 +491,13 @@ func buildFiberConfigMap(
...
 
 	if ver.Ensembler != nil && ver.Ensembler.Type == models.EnsemblerStandardType {
-		propsMap["experiment_mappings"] = ver.Ensembler.StandardConfig.ExperimentMappings
+		if ver.Ensembler.StandardConfig.ExperimentMappings != nil &&
+			len(ver.Ensembler.StandardConfig.ExperimentMappings) != 0 {
+			propsMap["experiment_mappings"] = ver.Ensembler.StandardConfig.ExperimentMappings
+		}
+		if ver.Ensembler.StandardConfig.RouteNamePath != "" {
+			propsMap["route_name_path"] = ver.Ensembler.StandardConfig.RouteNamePath
+		}
```
deadlycoconuts, line 500: These conditional values ensure that only non-nil/non-empty values are passed to the fiber config.

api/turing/validation/validator.go
krithika369: Makes sense. As for the duplicate error messages, I think we can just log the message once instead of calling `sl.ReportError` twice under each check. I guess it's ok to associate the error to either field `ExperimentMappings` or `RouteNamePath` ? I'm also ok with leaving it as it is - no strong opinion.

krithika369 (approved):
LGTM! Thanks for the changes.

Since we haven't yet added the `route_name_path` to the SDK, can we do that in a separate PR? https://github.com/caraml-dev/turing/blob/main/sdk/turing/router/config/router_ensembler_config.py#L35

api/turing/validation/validator.go
deadlycoconuts: That makes sense; I've removed the duplicated messages and decided to keep the message concerning `ExperimentMappings` in both instances (where both fields are set or neither fields are set), and drop the one concerning `RouteNamePath`.

api/api/specs/routers.yaml
deadlycoconuts: Okay as discussed earlier, I'll be removing the regex changes to allow users more flexibility (at their own discretion) at specifying the route path names. Any resultant errors from any incorrect parsing of the path would be returned by the router itself.

deadlycoconuts:
Thanks a lot for reviewing the PR and for all the comments! I'll be merging this right now and will be following up with a separate PR for the SDK-related changes and e2e tests! 🚀🚀

PR #238, opened 2022-09-12
Grpc fiber integration with UPI in router

leonlnj:
### Description
This MR integrates the fiber grpc implementation within the router, to support the simplest configuration of a Turing router with grpc [UPI](https://github.com/caraml-dev/universal-prediction-interface) routes.

Current router will not work with experiment engine or traffic rules yet. HTTP functionality are working as per normal.

*Note: Merging to a feature branch for this. CI aren't triggering since the target isn't a main branch. CI can be viewed under fork [repo's action](https://github.com/leonlnj/turing/actions).

### Changes
- New UPI mission control that supports `Route`
`engines/router/missionctl/mission_control_upi.go` 
Mission control contains the Route/Enrich/Ensemble logic. For now only route is added which is simply invoking the grpc fiber component.

- New UPI server
`engines/router/missionctl/server/grpc/server.go`
Similar to the http handler. `PredictValues` of the UPI contract is actually similar to `/predict` endpoint. The UPI server will call the UPI mission control's route.

- New env var in `RouterConfig` to indicate http or grpc router
`engines/router/missionctl/config/config.go`
This env var should eventually be created by API Server when deploying router.

- Introduce ErrorProtocol for mapping Turing existing error into respective status code. Rename `HttpError` into `TuringError` 
`engines/router/missionctl/errors/errors.go`
Referred to this [doc](https://github.com/grpc/grpc/blob/master/doc/http-grpc-status-mapping.md) as guideline. 

- Entrypoint to read envar and to create upi or grpc. Cmux library usage to serve http and grpc on same port.
`engines/router/missionctl/server/application.go`

Screenshot 2022-09-16 at 8 44 27 PM
Screenshot 2022-09-16 at 8 44 50 PM

- Benchmarking for grpc/http UPI 
This benchmark will use a common grpc/http client to call the Route directly versus calling through Turing
`engines/router/missionctl/internal/benchmark/benchmark_e2e_test.go`

#### Broken implementation (Traffic rules and experiment engine)
- `engines/router/missionctl/fiberapi/routing_strategy.go`
Skipping grpc request for now. Experiment engine will be skipped. Eventually there should be a mapping into http payload to call the experiment engine, after extracting the experiment engine mapping and values from grpc request with protopath. 
This will integrate with upcoming standard ensembler feature as well.

- `engines/router/missionctl/fiberapi/traffic_splitting_strategy.go` and `engines/router/missionctl/fiberapi/fan_in.go`
not working with grpc request.
To be revisited with protopath and new trafic rules.

#### Minor changes
- `engines/router/missionctl/fiberapi/utils.go` - minor refactor to separate fiber component and handler creation.
- moved `http` directory to `server/http`
Added generics for proto metadata and http header since both are `map[string][]string`

### Performance
Direct - Calling Route server directly
Turing - Calling Route server with Turing as proxy

For grpc, the generated service client is used for all.
For http, the route server uses `protojson` to convert request and response.
```
Benchmark_Upi_Grpc_Direct_Small-2   	    9782	    114621 ns/op	   17702 B/op	     312 allocs/op
Benchmark_Upi_Grpc_Direct_Large-2   	     247	   4813512 ns/op	 2428954 B/op	   27230 allocs/op
Benchmark_Upi_Grpc_Turing_Small-2   	    2696	    378148 ns/op	   60782 B/op	    1055 allocs/op
Benchmark_Upi_Grpc_Turing_Large-2   	     100	  12425686 ns/op	 88843 B/op	  108667 allocs/op
Benchmark_Upi_Http_Direct_Small-2   	    2917	    395544 ns/op	   71293 B/op	     656 allocs/op
Benchmark_Upi_Http_Direct_Large-2   	      54	  2204 ns/op	 5844320 B/op	  120654 allocs/op
Benchmark_Upi_Http_Turing_Small-2   	    2156	    554439 ns/op	   96649 B/op	     953 allocs/op
Benchmark_Upi_Http_Turing_Large-2   	      46	  24040678 ns/op	 9929781 B/op	  121016 allocs/op
```
 
### Headers transmission
Passing of headers can be rather confusing as there is a need to state explicitly whether its an incoming or outgoing context. And Turing is a server to the user client, while acting as the client (through fiber) to the Routes (server). 

The [guidelines](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md) were followed and verified that headers sent by a Routes is indeed accessible to the user client.

Screenshot 2022-09-12 at 1 07 51 PM


api/turing/testdata/cluster/servicebuilder/router_configmap_default.yml
```diff
@@ -2,6 +2,7 @@ id: test-svc
 routes:
 - endpoint: http://www.mocky.io/v2/5e4caccc310000e2cad8c071
   id: control
+  protocol: ""
```
leonlnj, line 5: current FiberConfig has the newly added field `protocol` which is parse to be empty from the api server for now. Subsequently api server will enforce `http` or `grpc`

deadlycoconuts:
Omg that's a lot of line changes 😶

engines/router/go.mod
```diff
@@ -5,32 +5,34 @@ go 1.18
 require (
 	bou.ke/monkey v1.0.2
 	cloud.google.com/go/bigquery v1.14.0
+	github.com/buger/jsonparser v1.1.1
 	github.com/caraml-dev/turing/engines/experiment v0.0.0
+	github.com/caraml-dev/universal-prediction-interface v0.0.0-20220912035120-8fbb12ddbc0e
```
pradithya, line 10: Can you bump this to 0.2.0

engines/router/missionctl/instrumentation/metrics/metrics.go
```diff
@@ -80,3 +80,18 @@ func GetStatusString(status bool) string {
 	}
 	return statusLabels.Failure
 }
+
+// GetMeasureDurationFunc return the func that measures the duration of the request
+func GetMeasureDurationFunc(err error, componentID string) func() {
```
krithika369, line 85: This function is hardcoding the metric name (TuringComponentRequestDurationMs) and labels. So I feel it doesn't belong in this package where the functions are generic. Also based on the comment on `engines/router/missionctl/fiberapi/fan_in.go`, I'm not convinced we need it.

engines/router/missionctl/server/grpc/server.go
```diff
@@ -0,0 +1,157 @@
...
+	upiv1 "github.com/caraml-dev/universal-prediction-interface/gen/go/grpc/caraml/upi/v1"
+	"github.com/gojek/fiber"
+	fibergrpc "github.com/gojek/fiber/grpc"
+	"github.com/opentracing/opentracing-go"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/reflection"
+)
+
+const turingReqIDHeaderKey = "Turing-Req-ID"
```
krithika369, line 28: Minor thing - but should we move this to a `constants.go` or something so it can be reused for all handlers?

engines/router/missionctl/fiberapi/fan_in.go
```diff
@@ -145,17 +158,7 @@ func (fanIn *EnsemblingFanIn) collectResponses(
...
-		map[string]func() string{
-			"status": func() string {
-				return metrics.GetStatusString(err == nil)
-			},
-			"component": func() string {
-				return "fanin_marshalResponse"
-			},
-		},
-	)
+	timer := metrics.GetMeasureDurationFunc(err, "fanin_marshalResponse")
```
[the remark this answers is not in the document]
leonlnj: oh right i missed this out, thanks for catching

engines/router/missionctl/config/config.go
```diff
@@ -178,6 +187,7 @@ type EnrichmentConfig struct {
 type RouterConfig struct {
 	ConfigFile string        `split_words:"true" required:"true"`
 	Timeout    time.Duration `default:"20ms"`
+	Protocol   Protocol      `default:"http"`
```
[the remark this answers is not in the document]
leonlnj: right, will uppercase it

krithika369 (commented):
Thanks for refactoring the server / handlers! Left some comments.

engines/router/missionctl/config/config.go
```diff
@@ -39,6 +39,15 @@ const (
 	NopLogger ResultLogger = "NOP"
 )
 
+// Protocol is used for router config to indicate spinning up of respective router mission control
+type Protocol string
+
+const (
+	UPI  Protocol = "UPI"
```
pradithya, line 46: Do you have objection to make it consistent with Merlin https://github.com/gojek/merlin/blob/969f9d0d7fe7631884498659978e19e75b92bfa1/swagger.yaml#L1669 ?

engines/router/missionctl/errors/errors.go
```diff
@@ -28,6 +29,13 @@ const (
 	TimeOut
 )
 
+type ErrorProtocol string
+
+const (
+	HTTP ErrorProtocol = "HTTP"
```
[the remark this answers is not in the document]
pradithya: Shall we reuse `Protocol` type instead of creating `ErrorProtocol` ?

engines/router/README.md
```diff
@@ -74,3 +74,4 @@ __Note:__ The app uses the `in_forward` plugin to write data to a TCP socket.
     -H "X-B3-Traceid: 950f2de0b8430e9fa30ec88c39471716" \
     -d '{}'
 ```
+4. Grpc routes needs to be serving with reflection on, in order for UPI router to initialise successfully.
```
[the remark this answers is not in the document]
leonlnj: For UPI no, since the contract will be known and they are not called by Fiber. This behavior is as such because Fiber has no knowledge of UPI and uses reflection to get the response proto.

It should be part of Turing service definition later on with the following signature.

```
service RouterService {
  rpc Enrich(PredictValuesRequest) returns (PredictValuesRequest);
  rpc Ensemble(EnsemblerRequest) returns (PredictValuesResponse);
}
```

engines/router/missionctl/errors/errors.go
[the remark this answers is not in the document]
leonlnj: UPI and GRPC both uses grpc status code, so I didnt introduce UPI here as I think the error code and status when it pipe down to Turing like `errors.Newf(errors.BadResponse, ....` for both UPI and GRPC will be the same.

api/turing/testdata/cluster/servicebuilder/router_configmap_default.yml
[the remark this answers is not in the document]
leonlnj: Eventually no, I will be adding the grpc implementation to API server after and it should always indicate the protocol but building the cluster resources.

engines/router/missionctl/mission_control_upi.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	return &missionControlUpi{
+		fiberRouter: fiberRouter,
+	}, nil
+}
+
+func (us *missionControlUpi) Route(
+	ctx context.Context,
+	fiberRequest fiber.Request) (
```
pradithya, line 42: nit: the parenthesis formatting seems out of place

engines/router/README.md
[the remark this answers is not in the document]
krithika369: I see.. Then, is there a way to init the Fiber component with the UPI contract for the routes too? Or is that the problem you had summarised [here](https://github.com/gojek/fiber/pull/3#issuecomment-1230235098) and we will continue to use reflection for Fiber in all cases?

engines/router/missionctl/mission_control_upi.go
```diff
@@ -0,0 +1,97 @@
...
+
+	grpcResponse, ok := resp.(*fibergrpc.Response)
+	if !ok {
+		turingError = errors.NewTuringError(
+			errors.Newf(errors.BadResponse, "unable to parse fiber response into grpc response"), errors.GRPC,
+		)
+		return nil, turingError
+	}
+
+	var responseProto upiv1.PredictValuesResponse
```
pradithya, line 69: Can you create a card in our backlog to address the `Marshal` and `Unmarshall` proto in this code. I think this is the largest contributor to the current performance.

engines/router/missionctl/mission_control_upi.go
```diff
@@ -0,0 +1,97 @@
...
+
+	return &missionControlUpi{
+		fiberRouter: fiberRouter,
+	}, nil
+}
+
+func (us *missionControlUpi) Route(
+	ctx context.Context,
+	fiberRequest fiber.Request) (
+	*upiv1.PredictValuesResponse, *errors.TuringError) {
```
pradithya, line 43: What happen if we return dynamic message here?

engines/router/missionctl/server/grpc/server.go
```diff
@@ -0,0 +1,157 @@
+package grpc
```
pradithya, line 1: I think it should be under upi package

engines/router/missionctl/server/grpc/server.go
```diff
@@ -0,0 +1,157 @@
...
+	*upiv1.PredictValuesResponse, error) {
+	var predictionErr *errors.TuringError
+	defer metrics.GetMeasureDurationFunc(predictionErr, tracingComponentID)()
+
+	// Create context from the request context
+	ctx = turingctx.NewTuringContext(ctx)
+	// Create context logger
+	ctxLogger := log.WithContext(ctx)
+	defer func() {
+		_ = ctxLogger.Sync()
```
pradithya, line 91: Why do we need to `Sync` when exiting `PredictValues` ? Wouldn't that slow it down?

engines/router/missionctl/server/grpc/server.go
```diff
@@ -0,0 +1,157 @@
...
+	if tracing.Glob().IsEnabled() {
+		var sp opentracing.Span
+		sp, _ = tracing.Glob().StartSpanFromContext(ctx, tracingComponentID)
+		if sp != nil {
+			defer sp.Finish()
+		}
+	}
+
+	fiberRequest := &fibergrpc.Request{
+		Message:  req,
```
pradithya, line 118: For the UPI we'll also want to populate the [`prediction_id`](https://github.com/caraml-dev/universal-prediction-interface/blob/8fbb12ddbc0e7f88d822d36ef685ef6c56475ca3/proto/caraml/upi/v1/upi.proto#L100) with a unique identifier (similar to turing id). Shall we reuse turing id in this case? cc: @krithika369

pradithya: Another doubt I have is to whether store these in the grpc Metadata or in the payload. The drawback of putting it in the payload is we'll need to copy the proto message.

engines/router/missionctl/mission_control_upi_test.go
```diff
@@ -0,0 +1,255 @@
...
+	err = protojson.Unmarshal(fileByte, upiRequest)
+	require.NoError(b, err)
+
+	conn, err := grpc.Dial(fmt.Sprintf(":%d", grpcport1), grpc.WithTransportCredentials(insecure.NewCredentials()))
+	require.NoError(b, err)
+	client := upiv1.NewUniversalPredictionServiceClient(conn)
+
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		benchMarkUpiResp, _ = client.PredictValues(context.Background(), upiRequest)
```
pradithya, line 237: Can you make the response from server adjustable? Currently it only returns a mocked response which is not representative of the actual use case. One alternative that you could take is to populate the `prediction_result_table` with the `prediction_table` of the request.

engines/router/missionctl/testdata/upi_small_payload.json
```diff
@@ -0,0 +1,277 @@
+{
```
pradithya, line 1: nit: Rather than hardcoding the payload shall we make a utility function to programmatically create one?

engines/router/missionctl/mission_control_upi_test.go
pradithya: Additionally, can you try to profile the code (using pprof) when executing the benchmark to see whether there is some low hanging fruit for optimization. I would love to reduce the performance to be as close as the plain grpc one.

engines/router/missionctl/mission_control_upi.go
pradithya: Asking to understand whether it can be used to address https://github.com/caraml-dev/turing/pull/238#discussion_r969130672

engines/router/README.md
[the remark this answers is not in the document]
leonlnj: For the [current implementation in fiber](https://github.com/gojek/fiber/blob/main/grpc/dispatcher.go#L108) no,

I believe it is do-able if we change it to do a search in the protoregistry (since its already registered in Turing) by providing the proto response type name.

engines/router/missionctl/server/application.go
```diff
@@ -47,37 +48,48 @@ func Run() {
...
-	// Register metrics handler
-	if cfg.AppConfig.CustomMetrics {
-		http.Handle("/metrics", promhttp.Handler())
-	}
-
-	// Serve
-	log.Glob().Infof("listening at port %d", cfg.Port)
-	if err := http.ListenAndServe(cfg.ListenAddress(), http.DefaultServeMux); err != nil {
-		log.Glob().Errorf("Failed to start Turing Mission Control API: %s", err)
+	if cfg.RouterConfig.Protocol == config.UPI {
```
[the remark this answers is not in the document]
leonlnj: hmmm

I will remove the `default http` and change it into `required` in `config.go` - this will enforce the protocol is always passed in. 

Will also do what u suggestion, to test again no matches.

engines/router/missionctl/server/grpc/server.go
leonlnj: i thought of similar but it will be a standalone file+package with this one variable. Will put it in `server/constant` for now

engines/router/missionctl/config/config.go
leonlnj: not from me, will do and remove grpc for now (till we have custom grpc)

engines/router/missionctl/mission_control_upi.go
leonlnj: we can but we would still need to do the marshaling in the UPIServer to satisfy the interface regardless? 

```
PredictValues(ctx context.Context, req *upiv1.PredictValuesRequest) (*upiv1.PredictValuesResponse, error) 
```

engines/router/missionctl/server/grpc/server.go
leonlnj: hmm im not 100% sure on this, the http handler are behaving the same too.

It is [written in zap comment](https://github.com/uber-go/zap/blob/master/logger.go#L274) that application should handle it before exiting and also [sync in their quickstart](https://github.com/uber-go/zap#quick-start). 

But im not sure when should it be synced exactly.

pradithya: > It is [written in zap comment](https://github.com/uber-go/zap/blob/master/logger.go#L274) that application should handle it before exiting and also [sync in their quickstart](https://github.com/uber-go/zap#quick-start).

It's correct, you should call `Sync` explicitly when exiting the program (i.e. shutting down), other than that calling `Sync` should be avoided unless necessary. Depending on the logger sink, `Sync` might cause performance degradation. E.g. if the sink is the file, that means every call to `Sync` will write to the file.

engines/router/missionctl/mission_control_upi.go
pradithya: Ah yes, you are right

engines/router/missionctl/server/upi/server.go
```diff
@@ -0,0 +1,167 @@
...
+	*upiv1.PredictValuesResponse, *errors.TuringError) {
+
+	// Create response channel to store the response from each step. 1 for route now,
+	// should be 4 when experiment engine, enricher and ensembler are added
+	respCh := make(chan grpcRouterResponse, 1)
+	defer close(respCh)
+
+	// Defer logging request summary
+	defer func() {
+		go logTuringRouterRequestSummary(
```
tiopramayudi, line 151: i think there is a chance that respCh already closed and not all the value in the channel already logged

engines/router/missionctl/mission_control_upi.go
```diff
@@ -0,0 +1,97 @@
...
+	grpcResponse, ok := resp.(*fibergrpc.Response)
+	if !ok {
+		turingError = errors.NewTuringError(
+			errors.Newf(errors.BadResponse, "unable to parse fiber response into grpc response"), errors.GRPC,
+		)
+		return nil, turingError
+	}
+
+	var responseProto upiv1.PredictValuesResponse
+	payloadByte, err := proto.Marshal(grpcResponse.Payload().(proto.Message))
```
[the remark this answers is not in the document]
leonlnj: nope, tried it while implementing fiber previously

image

engines/router/missionctl/server/grpc/logutils.go
```diff
@@ -0,0 +1,87 @@
...
+) {
+
+	// Create a new TuringResultLogEntry record with the context and request info
+	logEntry := resultlog.NewTuringResultLogEntry(ctx, timestamp, header, body.String())
+
+	// Read incoming responses and prepare for logging
+	for resp := range mcRespCh {
+		// If error exists, add an error record
+		if resp.err != "" {
+			logEntry.AddResponse(resp.key, "", nil, resp.err)
```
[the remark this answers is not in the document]
leonlnj: did u meant to change to this?
```
	if resp.err != "" {
		logEntry.AddResponse(resp.key, "", nil, resp.err)
		continue
	}
	logEntry.AddResponse(resp.key, resp.body.String(), resultlog.FormatHeader(resp.header), "")
```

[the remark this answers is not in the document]
tiopramayudi: yes

engines/router/missionctl/server/application.go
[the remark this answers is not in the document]
leonlnj: just realized i cant put it as required, until i patch the API Server. will do so in an upcoming MR. setting it back as default for now.

engines/router/missionctl/server/application.go
```diff
@@ -47,37 +48,51 @@ func Run() {
...
+	case config.UPI:
+		// Init mission control
+		missionCtl, err := missionctl.NewMissionControlUPI(
+			cfg.RouterConfig.ConfigFile,
+			cfg.AppConfig.FiberDebugLog,
+		)
+		if err != nil {
+			log.Glob().Panicf("Failed initializing Mission Control: %v", err)
+		}
+		s := upi.NewUPIServer(missionCtl, cfg.Port)
```
pradithya, line 61: We'll need to expose http metrics server in addition to the UPI gRPC server

engines/router/missionctl/mission_control_upi.go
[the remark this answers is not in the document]
leonlnj: no longer using dynamicpb

engines/router/missionctl/server/application.go
leonlnj: added [cmux](https://github.com/soheilhy/cmux) to host both grpc and http on the same port. Seems to work (screenshots in description)

engines/router/missionctl/errors/errors.go
```diff
@@ -71,49 +73,62 @@ func GetType(err error) ErrorType {
...
+		case BadInput:
+			code = http.StatusBadRequest
+		case BadResponse:
+			code = http.StatusBadGateway
+		case NotFound:
+			code = http.StatusNotFound
+		default:
+			code = http.StatusInternalServerError
+		}
+	} else if p == protocol.GRPC {
```
pradithya, line 93: There are 2 definition of protocol currently, one in "github.com/gojek/fiber/protocol" another one in "engines/router/missionctl/config/". Can we merge it?

leonlnj: my bit of dilemma here is that `fiber` should only have `http` and `grpc` since it is message agnostic. 

In the event of upcoming custom grpc or other message types, Turing would need new `protocol` to identify them, whereas fiber doesn't need to (it will still be a grpc dispatcher regardless). Hence it was intentional to have `UPI` and `HTTP` here.

pradithya: Ah ok, didn't realized the first one is protocol in fiber.

pradithya: @leonlnj Maybe add alias in the import to differentiate between those 2. E.g. 
```
import (

`fiberProtocol "github.com/gojek/fiber/protocol"`

)
```

engines/router/missionctl/fiberapi/routing_strategy.go
```diff
@@ -58,55 +59,59 @@ func (r *DefaultTuringRoutingStrategy) SelectRoute(
...
-		expResultCh <- experimentResponse
-		close(expResultCh)
-	}
-
-	// If error, log it and return the fallback(s)
-	if expErr != nil {
-		log.WithContext(ctx).Errorf(expErr.Error())
-		return nil, fallbacks, nil
-	}
+	if req.Protocol() != protocol.GRPC {
```
pradithya, line 63: nit: Little advice, nominal case should be indented as minimal as possible. i.e. instead of wrapping the nominal case (protocol.HTTP), we can exit early and leave the rest of the code untouched.

```go
if req.Protocol() == protocol.GRPC {
    return return nil, fallbacks, nil
}

```

pradithya (commented):
Thanks, few minor comments above, the rest LGTM

leonlnj:
merging this to `grpc_upi` feature branch. will open a WIP draft PR from `grpc_upi` into `main` containing all the upi features. While creating smaller PR to the `grpc_upi` feature by feature.

PR #240, opened 2022-09-13
Disallow Dangling routes with Traffic rules

terryyylim:
With the introduction of `Default Traffic Rule` in https://github.com/caraml-dev/turing/pull/232, this PR adds validation to disallow routes that are not associated to any Traffic Rules (a.k.a dangling routes) _**when Traffic rules are specified**_. The motivation is because routes that are not used in any Traffic Rules (if specified), will never be called.

**Validation changes**
`UI`
image

`API`
image

`SDK`
image

### DIFF
--- a/api/turing/validation/validator.go
+++ b/api/turing/validation/validator.go
@@ -139,7 +139,6 @@ func checkDefaultTrafficRule(
 	sl validator.StructLevel,
 	fieldName string,
 	defaultTrafficRule *models.DefaultTrafficRule,
-	defaultRouteID *string,
 ) {
 	defaultTrafficRuleDescription := strings.Join([]string{
 		"Since 1 or more Custom Traffic rules have been specified,",
@@ -148,11 +147,18 @@ func checkDefaultTrafficRule(
 	if defaultTrafficRule == nil {
 		sl.ReportError(defaultTrafficRule, fieldName, "defaultTrafficRule", defaultTrafficRuleDescription, "")
 	}
+}
 
+func validateDefaultTrafficRuleDefaultRoute(
+	sl validator.StructLevel,
+	fieldName string,
+	defaultTrafficRule *models.DefaultTrafficRule,
+	defaultRouteID *string,
+) {
 	// DefaultRouteId should be present in Default Traffic Rule
 	if defaultTrafficRule != nil && defaultRouteID != nil {
 		missingDefaultRouteDescription := fmt.Sprintf(
-			"Fallback Route (DefaultRouteId): '%s' should be associated for the Default Traffic Rule", *defaultRouteID)
+			"Fallback Route (DefaultRouteId): '%s' should be associated to the Default Traffic Rule", *defaultRouteID)
 		containsDefaultRouteID := false
 		for _, route := range defaultTrafficRule.Routes {
 			if route == *defaultRouteID {
@@ -217,7 +223,9 @@ func validateRouterConfig(sl validator.StructLevel) {
 	allRuleRoutesSet := set.New()
 	if routerConfig.TrafficRules != nil {
 		if len(routerConfig.TrafficRules) > 0 {
-			checkDefaultTrafficRule(sl, "DefaultTrafficRule", routerConfig.DefaultTrafficRule, routerConfig.DefaultRouteID)
+			checkDefaultTrafficRule(sl, "DefaultTrafficRule", routerConfig.DefaultTrafficRule)
+			validateDefaultTrafficRuleDefaultRoute(
+				sl, "DefaultTrafficRule", routerConfig.DefaultTrafficRule, routerConfig.DefaultRouteID)
 			if routerConfig.DefaultTrafficRule != nil {
 				for _, route := range routerConfig.DefaultTrafficRule.Routes {
 					allRuleRoutesSet.Insert(route)
--- a/api/turing/validation/validator.go
+++ b/api/turing/validation/validator.go
@@ -6,11 +6,12 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/caraml-dev/turing/engines/router"
+	"github.com/golang-collections/collections/set"
 
 	"github.com/caraml-dev/turing/api/turing/api/request"
 	"github.com/caraml-dev/turing/api/turing/models"
 	"github.com/caraml-dev/turing/api/turing/service"
+	"github.com/caraml-dev/turing/engines/router"
 	"github.com/go-playground/validator/v10"
 	"github.com/go-playground/validator/v10/non-standard/validators"
 )
@@ -138,6 +139,7 @@ func checkDefaultTrafficRule(
 	sl validator.StructLevel,
 	fieldName string,
 	defaultTrafficRule *models.DefaultTrafficRule,
+	defaultRouteID *string,
 ) {
 	defaultTrafficRuleDescription := strings.Join([]string{
 		"Since 1 or more Custom Traffic rules have been specified,",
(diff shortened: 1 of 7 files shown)


api/turing/validation/validator.go
```diff
@@ -146,6 +148,43 @@ func checkDefaultTrafficRule(
 	if defaultTrafficRule == nil {
 		sl.ReportError(defaultTrafficRule, fieldName, "defaultTrafficRule", defaultTrafficRuleDescription, "")
 	}
+
+	// DefaultRouteId should be present in Default Traffic Rule
+	if defaultTrafficRule != nil && defaultRouteID != nil {
+		missingDefaultRouteDescription := fmt.Sprintf(
+			"Fallback Route (DefaultRouteId): '%s' should be associated for the Default Traffic Rule", *defaultRouteID)
```
krithika369, line 155: Nit: `should be associated for the` -> `should be associated to the`

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
```diff
@@ -57,20 +57,16 @@ export const RouteSelectionPanel = ({
   const routeOptions = routes
     .filter((e) => !!e.id)
     .map((e) => {
-      const isDisabled =
```
krithika369, line 60: For the UI flow, I think it makes sense to keep the validation on the fallback route options and disable those options that don't meet the criteria (if traffic rules are present, the route should be added to all traffic rules and the default rule). The reasons are:
1. It helps to give users get early feedback rather than wait for the form validation to trigger failures.
2. The final route (No ensembler) or fallback route (standard ensembler) config is usually done after the routes and traffic rules. In a scenario where the user tries to submit the form with an error, any subsequent changes will keep re-triggering the validation. And so, if they keep selecting bad route options in the ensembler section, the form will keep jumping back to the routes panel with the validation error.

ui/src/router/components/form/validation/schema.js
```diff
@@ -58,6 +58,60 @@ const validateRuleNames = function(items) {
...
+        this.createError({
+          path: `${this.path}[${idx}].id`,
+          message: "This route should be removed since they have no Traffic Rule(s) associated and will never be called.",
+        })
+      );
+    }
+  });
+
+  // Validate Nop/Standard Ensembler
+  const ensembler_type = this.options.parent.ensembler.type
```
krithika369, line 82: I think it's better for the yup UI validation to be warning on the ensembler config rather than the routes for the same reason as above. Besides, we already have the `validRouteSchema` on `fallback_response_route_id` and `final_response_route_id` - can we look at manipulating that definition rather than adding all the ensembler based validation here? We can rename this function from `validateRoutes` to `validateDanglingRoutes`.

api/turing/validation/validator.go
```diff
@@ -175,24 +214,55 @@ func validateRouterConfig(sl validator.StructLevel) {
 	}
 
 	// Validate traffic rules
+	allRuleRoutesSet := set.New()
 	if routerConfig.TrafficRules != nil {
 		if len(routerConfig.TrafficRules) > 0 {
-			checkDefaultTrafficRule(sl, "DefaultTrafficRule", routerConfig.DefaultTrafficRule)
```
krithika369, line 180: While this check is logically correct, do you think it'll be better to move it into a separate check: `validateDefaultRouteTrafficRules(routeID, rules, defaultRoute)`

Within that function, you can append the default rule to the other rules and run the check to make sure that routeID appears in each rule. That will be clearer and better isolated, right?

```golang
if routerConfig.TrafficRules != nil {
...
    if routerConfig.DefaultTrafficRule != nil {
        validateDefaultRouteTrafficRules(routeID, rules, defaultRoute)
```

wdyt?

terryyylim: Would you mean to shift the newly added logic to `checkDefaultTrafficRule` into a new function called `validateDefaultRouteTrafficRules `?

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
terryyylim: >that don't meet the criteria (if traffic rules are present, the route should be added to all traffic rules and the default rule)

I might be misunderstanding, but wouldn't this mean any route can be selected for the fallback route, so why would there be a need to disable the options?

ui/src/router/components/form/validation/schema.js
terryyylim: `validRouteSchema` is currently being used for Custom and Default Traffic Rules schema, and they don't require such sophisticated validation. Perhaps another separate function?

api/turing/validation/validator.go
krithika369: Yes!

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
krithika369: A route can be selected as fallback if:
* There are no traffic rules on the router
* There are traffic rules on the router and this route appears in all the rules including the default rule

If there are traffic rules on the router and a route doesn't appear in all rules (including default rule), we can disable it from this dropdown and prevent it from being selected in the first place instead of picking up this error on validation. Does it make sense?

ui/src/router/components/form/validation/schema.js
krithika369: Ahh got it. It's related to this: https://github.com/caraml-dev/turing/pull/240/files#r970360776

With the previous workflow, no way you can select an invalid route (previously, invalid route was any route with traffic rules) as final / fallback route. Now the condition for what is valid and invalid has changed. But if we keep that workflow, the UI will not allow such a value. So there is nothing additional to validate here in terms of the traffic rule and default route - such checks are done in the API and this invalid scenario can't be hit from the UI.

No harm having another layer of yup validations but I guess it's not strictly necessary.

terryyylim: Guess this PR could have been much simpler 😅 I'll remove this validation then!

ui/src/router/components/form/components/ensembler_config/RouteSelectionPanel.js
terryyylim: Clear to me, I've updated the condition for this!

api/turing/validation/validator.go
```diff
@@ -148,6 +149,50 @@ func checkDefaultTrafficRule(
 	}
 }
 
+func validateDefaultTrafficRuleDefaultRoute(
```
krithika369, line 152: Hey I meant move the newly added logic and merge the validation of default route against both the traffic rules and default traffic rules (my first comment): https://github.com/caraml-dev/turing/pull/240/files#r970310168

You can append everything into one array of rules and check that each rule has the default route.

krithika369: So the logic under `for ruleIdx, rule := range routerConfig.TrafficRules { ...` in the main  `validateRouterConfig` function can be simpler and we don't have to write the same kind of checks twice.

krithika369 (approved):
Left one comment on the API validation. The rest LGTM! Thanks, Terence!

krithika369:
One small request - could you update the screenshot in PR (especially the UI) based on the latest changes? Could serve as a quick reference later. Thanks.

terryyylim:
> One small request - could you update the screenshot in PR (especially the UI) based on the latest changes? Could serve as a quick reference later. Thanks.

I updated it already actually! Is there anything else you wanted in the screenshot that has not been captured?

api/turing/validation/validator.go
terryyylim: Ah sorry! I misread your comment. I've updated the validation now, should be much cleaner, thanks for suggestion 🙏🏻

terryyylim: I'll merge by EOD if there's nothing else!

krithika369:
> > One small request - could you update the screenshot in PR (especially the UI) based on the latest changes? Could serve as a quick reference later. Thanks.
> 
> I updated it already actually! Is there anything else you wanted in the screenshot that has not been captured?

Ah my bad, I was reading the description from an older tab. Thanks for this!

api/turing/validation/validator.go
```diff
@@ -148,6 +149,53 @@ func checkDefaultTrafficRule(
 	}
 }
 
+func validateDefaultRouteTrafficRules(
+	sl validator.StructLevel,
+	fieldName string,
+	defaultTrafficRules models.TrafficRules,
```
krithika369, line 155: Super nit: Why `defaultTrafficRules` here? Shall we just call it `trafficRules`  ? 😅

api/turing/validation/validator.go
krithika369: Np. I left a super nit comment. Please feel free to merge after reviewing. Nothing else from me. :)

api/turing/validation/validator.go
terryyylim: Oh yes thats a typo, updated! 🙏🏻

PR #242, opened 2022-09-15
Add validation for overlapping rules

terryyylim:
With the recent improvements to Traffic Rules, this PR adds an additional layer of validation for users, to validate that Traffic Rule conditions are not overlapping. Any violations will be returned in an error response that can be rectified by the user easily.

### DIFF
--- a/api/turing/validation/validator.go
+++ b/api/turing/validation/validator.go
@@ -196,6 +196,81 @@ func checkDanglingRoutes(
 	}
 }
 
+func validateConditionOrthogonality(
+	sl validator.StructLevel,
+	fieldName string,
+	allRules models.TrafficRules,
+) {
+	trafficConditionsFieldMap := map[string][]string{}
+	trafficConditionsValueMap := map[string]map[string]*set.Set{}
+	for _, rule := range allRules {
+		trafficConditionsValueMap[rule.Name] = map[string]*set.Set{}
+		trafficConditionsFieldMap[rule.Name] = []string{}
+		for _, condition := range rule.Conditions {
+			conditionValues := []interface{}{}
+			for _, val := range condition.Values {
+				conditionValues = append(conditionValues, val)
+			}
+			trafficConditionsValueMap[rule.Name][condition.Field] = set.New(conditionValues...)
+			trafficConditionsFieldMap[rule.Name] = append(trafficConditionsFieldMap[rule.Name], condition.Field)
+		}
+	}
+
+	allErrors := []string{}
+	for idx, rule1 := range allRules {
+		rule1Fields, ok := trafficConditionsFieldMap[rule1.Name]
+		// Check that there are rule conditions
+		if ok {
+			// Check that the current rule and the other are orthogonal
+			rulesOverlap := true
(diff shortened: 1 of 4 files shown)


api/turing/validation/validator.go
```diff
@@ -196,6 +196,81 @@ func checkDanglingRoutes(
...
+	allErrors := []string{}
+	for idx, rule1 := range allRules {
+		rule1Fields, ok := trafficConditionsFieldMap[rule1.Name]
+		// Check that there are rule conditions
+		if ok {
+			// Check that the current rule and the other are orthogonal
+			rulesOverlap := true
+
+			for _, rule2 := range allRules[idx+1:] {
+				// Rules with same name should fail unique name validation
```
krithika369, line 228: Is this comment correct? I think you mean to say "should fail the overlapping rules validation" ?

api/turing/validation/validator.go
```diff
@@ -196,6 +196,81 @@ func checkDanglingRoutes(
...
+func validateConditionOrthogonality(
+	sl validator.StructLevel,
+	fieldName string,
+	allRules models.TrafficRules,
+) {
+	trafficConditionsFieldMap := map[string][]string{}
+	trafficConditionsValueMap := map[string]map[string]*set.Set{}
+	for _, rule := range allRules {
+		trafficConditionsValueMap[rule.Name] = map[string]*set.Set{}
+		trafficConditionsFieldMap[rule.Name] = []string{}
```
krithika369, line 208: Do we need this data structure? Can we just rely on `trafficConditionsValueMap` instead?

```golang
for idx, rule1 := range allRules {
	rule1Fields, ok := trafficConditionsValueMap[rule1.Name] {
		for field, currSet := range rule1Fields {
...
```

This will also make the code more concise.

api/turing/validation/validator_test.go
```diff
@@ -645,6 +645,122 @@ func TestValidateTrafficRules(t *testing.T) {
 				"'Fallback Route (DefaultRouteId): 'route-a' should be associated to all Traffic Rules' tag",
 			}, " "),
 		},
+		"failure | Overlapping Traffic Rules": {
```
krithika369, line 648: Can we also add a test case with multiple rules and different combinations of the conditions that result in a success?

api/turing/validation/validator.go
```diff
@@ -255,5 +330,6 @@ func validateRouterConfig(sl validator.StructLevel) {
 	// Validate dangling routes
```
krithika369, line 330: This comment needs to be updated to include the orthogonality check.

krithika369 (approved):
Left some comments. Please feel free to merge this PR when you've reviewed / addressed them. The rest LGTM.

api/turing/validation/validator.go
terryyylim: Oh, yes it's correct. What I meant to say here was that this case can be skipped for orthogonality checks, because it would have failed the unique name check anyway, and it doesn't make sense to check the same rule against itself.

PR #243, opened 2022-09-20
Bugfix: Pyfunc ensemblers - copy autoscaling config to docker config

krithika369:
Autoscaling policy configs were added in https://github.com/caraml-dev/turing/pull/234. When a Pyfunc ensembler is deployed, the autoscaling policy config should be copied to the Docker config of the ensembler, like the other deployment properties. This PR adds this change and a unit test to cover this scenario.


leonlnj (approved):
LGTM!

PR #244, opened 2022-09-20
UI bug fix and mlp-ui dependency upgrade

terryyylim:
This PR consists of UI bug fixes:
- Bug fixes from mlp-ui dependency: https://github.com/gojek/mlp/pull/59
- Some bugs was introduced in a previous [PR](https://github.com/caraml-dev/turing/pull/239)
  - where the RulesConfigTable considered the presence of Default Traffic Rule for Routers with no Traffic Rules.
  - where the RoutesConfigTable showed an expansion button even with no Traffic Rules

### DIFF
--- a/ui/src/router/components/configuration/components/router_config_section/RulesConfigTable.js
+++ b/ui/src/router/components/configuration/components/router_config_section/RulesConfigTable.js
@@ -23,16 +23,18 @@ export const RulesConfigTable = ({ routes, rules = [], defaultTrafficRule }) => 
     });
 
     // Append Default Traffic Rule
-    const defaultRuleRoutes = [];
-    defaultTrafficRule.routes.forEach((ruleRoute) => {
-        const filteredRoute = routes.find(route => route.id === ruleRoute);
-        defaultRuleRoutes.push(filteredRoute);
-    });
-    updatedRules.push({
-      name: "default-traffic-rule",
-      conditions: [],
-      routes: defaultRuleRoutes
-    });
+    if (defaultTrafficRule) {
+      const defaultRuleRoutes = [];
+      defaultTrafficRule.routes.forEach((ruleRoute) => {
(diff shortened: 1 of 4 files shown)


ui/src/router/components/configuration/components/router_config_section/RulesConfigTable.js
```diff
@@ -23,16 +23,18 @@ export const RulesConfigTable = ({ routes, rules = [], defaultTrafficRule }) =>
...
-    defaultTrafficRule.routes.forEach((ruleRoute) => {
-        const filteredRoute = routes.find(route => route.id === ruleRoute);
-        defaultRuleRoutes.push(filteredRoute);
-    });
-    updatedRules.push({
-      name: "default-traffic-rule",
-      conditions: [],
-      routes: defaultRuleRoutes
-    });
+    if (defaultTrafficRule) {
```
deadlycoconuts, line 26: So just to confirm what was happening previously was that we've been appending a 'ghost' rule (see below) to the updated rules list, when there should be 0 rules? 

```js
{
      name: "default-traffic-rule",
      conditions: [],
      routes: [],
}
```

terryyylim: Ah nope, that's actually this part here - https://github.com/caraml-dev/turing/pull/244/files#diff-4356a77431e53504f977550d197cb1c75c1754691de008de5d288d578469e5e3R115.

deadlycoconuts: Ahhh okay, in which case I've got a different comment, why don't we do `if (routes)` or `if (routes.length !== 0)` instead? 😅

deadlycoconuts (approved):
LGTM! Thanks for this! 🚀

PR #245, opened 2022-09-21
Bugfix: Handle non-string ensembler dependencies in conda_env

krithika369:
When logging an ensembler, we create / replace the python version conda dependency such that it matches the minor version. Eg: `python=3.7.*`

The comparison function that matched the python dependency from the conda_env dependencies treated each item as a string whereas they could also be objects ([example](https://www.mlflow.org/docs/latest/python_api/mlflow.pyfunc.html#mlflow.pyfunc.log_model)).

This resulted in an error when attempting to create ensemblers with explicit `pip` dependencies. This PR addresses this bug and updates the unit test cases to cover complex dependencies.


deadlycoconuts (approved):
LGTM! 🚀 Thanks for looking into this!

PR #246, opened 2022-09-21
Bugfix: Migration for autoscaling configs when using pyfunc ensemblers

krithika369:
Follow up from the previous related bugfix PR: https://github.com/caraml-dev/turing/pull/243

As the docker configs are also present in Pyfunc ensembler configs, the migration scripts need to account for them as well and update the autoscaling policy if the `docker_config` is not NULL (which can happen if the image was never successfully built).


deadlycoconuts (approved):
Thanks for digging this out and for fixing this! LGTM! 🚀

PR #247, opened 2022-09-22
Accept HPA memory target as %

krithika369:
When the autoscaling metric is set to "memory", Knative expects the target to be supplied in the unit of `Mi` ([ref](https://pkg.go.dev/knative.dev/serving/pkg/apis/autoscaling)). It is common for users to configure the HPA using resource utilization % instead and this is the unit used for the "cpu" metric by Knative as well. Moreover, if `Mi` is used as the unit, the users may need to update the autoscaling target value every time the memory requested is altered.

Thus, this PR changes the handling of the target value set for the memory metric so that the saved config will be treated as `%` and subsequently applied as `Mi`. This will make the configuration consistent for the end users.

### Main changes
* `api/turing/cluster/knative_service.go` - Add method `getAutoscalingTarget` for manipulating the target, if "memory" is used.
* `api/turing/cluster/knative_service_test.go` - Add unit tests for the `getAutoscalingTarget` method, to test the conversions
* Update the documentation in the UI, for the memory unit

CC: @ariefrahmansyah

### DIFF
--- a/api/turing/cluster/knative_service_test.go
+++ b/api/turing/cluster/knative_service_test.go
@@ -7,6 +7,7 @@ import (
 
 	tu "github.com/caraml-dev/turing/api/turing/internal/testutils"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
 	resource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -247,7 +248,7 @@ func TestBuildKnativeServiceConfig(t *testing.T) {
 									"autoscaling.knative.dev/minScale": "5",
 									"autoscaling.knative.dev/maxScale": "6",
 									"autoscaling.knative.dev/metric":   "memory",
-									"autoscaling.knative.dev/target":   "70",
+									"autoscaling.knative.dev/target":   "358",
 									"autoscaling.knative.dev/class":    "hpa.autoscaling.knative.dev",
 								},
 							},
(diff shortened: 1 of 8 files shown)


api/turing/cluster/models.go
```diff
@@ -125,7 +125,7 @@ type ConfigMap struct {
 
 // Ref:
 // https://github.com/knative/serving/blob/release-0.14/pkg/reconciler/revision/resources/queue.go#L115
-func computeResource(resourceQuantity resource.Quantity, fraction float64) resource.Quantity {
+func ComputeResource(resourceQuantity resource.Quantity, fraction float64) resource.Quantity {
 	scaledValue := resourceQuantity.Value()
```
terryyylim, line 129: TIL this helps with verifying the number, incase of overflowing values.

api/turing/cluster/knative_service_test.go
```diff
@@ -267,9 +268,92 @@ func TestBuildKnativeServiceConfig(t *testing.T) {
...
+			},
+			expectedTarget: "1434", // (70/100) * (2Gi * 1024)
+		},
+		"memory | G": {
+			cfg: &KnativeService{
+				BaseService: &BaseService{
+					MemoryRequests: resource.MustParse("2G"),
+				},
+				AutoscalingMetric: "memory",
+				AutoscalingTarget: "70",
```
terryyylim, line 332: Would a test case with `AutoscalingTarget` > 100 be useful?

terryyylim (approved):
LGTM, thanks @krithika369 !

api/turing/cluster/knative_service_test.go
krithika369: Good question. I considered it, but decided not to restrict the values. For example, we have the resource limit set to be a factor of the request. What if the user really only wants the autoscaling to kick in at 105% ? This is probably unlikely but I couldn't think of a strong reason to enforce this check.

krithika369:
Thanks for the review! Will merge it shortly.

PR #248, opened 2022-09-22
Add Standard Ensembler with Route Name Path to SDK

deadlycoconuts:
## Context
With the changes introduced in #237 and #233 to support the usage of Standard Ensemblers with Custom Experiment Engines, the Turing API has been updated to handle different request/response payloads, most notably the addition of the `route_name_path` field in the standard ensembler configuration, thus necessitating a change to the Turing SDK to also handle these changes.

This PR introduces minor changes to the `EnsemblerStandardConfig` and `StandardRouterEnsemblerConfig` to support these changes as well as a new e2e test to mirror that of the newest API test introduced in #237.

## Main Modifications
- `.github/workflows/turing.yaml` - Removal of the `paths-ignore` value representing all sdk files in the repo
- `sdk/e2e/09_deploy_router_with_route_name_path_std_ensembler_test.py` - Addition of an e2e test that corresponds to the latest API e2e test added in #237
- `sdk/turing/router/config/router_ensembler_config.py` - Addition of the `route_name_path` field to `EnsemblerStandardConfig` and `StandardRouterEnsemblerConfig`


.github/workflows/turing.yaml
```diff
@@ -27,7 +27,6 @@ on:
       - "docs/**"
       - "engines/pyfunc-ensembler-job/**"
       - "engines/pyfunc-ensembler-service/**"
-      - "sdk/**"
```
deadlycoconuts, line 30: With the addition of SDK e2e tests, I believe it's probably useful to make the main CI pipeline run (e2e tests) whenever changes to the SDK are made.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -32,20 +32,41 @@ def to_open_api(self) -> OpenApiModel:
 
 @dataclass
 class EnsemblerStandardConfig:
-    experiment_mappings: List[Dict[str, str]]
+    experiment_mappings: List[
+        turing.generated.models.EnsemblerStandardConfigExperimentMappings
+    ]
```
deadlycoconuts, line 37: Replaced the type hint because it was always of the wrong type 👀

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -487,14 +509,20 @@ def to_open_api(self) -> OpenApiModel:
...
         Method to create a new standard ensembler
 
-        :param experiment_mappings: configured mappings between routes and treatments
+        :param route_name_path: configured routh name path that points to the route name within a given treatment config
+                :param experiment_mappings: configured mappings between routes and treatments
+        :param fallback_response_route_id: configured final response route to be used as a fallback
         """
-        self.experiment_mappings = experiment_mappings
+        self.route_name_path = route_name_path
+        self.experiment_mappings = experiment_mappings if experiment_mappings else []
```
deadlycoconuts, line 525: Since `route_name_path` and `experiment_mappings` are optional fields, we'll need to handle such edge cases. when both are empty, we'll be letting the API return an error.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -487,14 +509,20 @@ def to_open_api(self) -> OpenApiModel:
...
+        self,
+        fallback_response_route_id: str,
+        experiment_mappings: List[Dict[str, str]] = None,
+        route_name_path: str = "",
     ):
         """
         Method to create a new standard ensembler
 
-        :param experiment_mappings: configured mappings between routes and treatments
+        :param route_name_path: configured routh name path that points to the route name within a given treatment config
```
terryyylim, line 520: nit: typo `routh` -> `route`

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -487,14 +509,20 @@ def to_open_api(self) -> OpenApiModel:
...
+                :param experiment_mappings: configured mappings between routes and treatments
```
terryyylim, line 521: nit: formatting is misaligned

sdk/turing/router/config/router_ensembler_config.py
terryyylim: Why not just configure the default `[]` value in the parameters?
```
experiment_mappings: List[Dict[str, str]] = [],
```

If `None` is configured, typically the typing used should be `Optional`, rather than a `List` in this case.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -545,6 +582,9 @@ def to_open_api(self) -> OpenApiModel:
                 )
                 for experiment_mapping in self.experiment_mappings
             ],
+            route_name_path=self.route_name_path
+            if self.route_name_path is not None
+            else "",
```
terryyylim, line 587: I believe an empty string "" is required because the API validator requires this field to be set right? Also, in what cases will this value be `None`, asking this because the default value should already be "" based on how you initialise this class, so I think this is not required?

sdk/e2e/09_deploy_router_with_route_name_path_std_ensembler_test.py
```diff
@@ -0,0 +1,123 @@
+import os
```
terryyylim, line 1: Is this test necessary? I'm just thinking we don't include any E2E tests for all other ensembler types too. And, by this logic, shouldn't we also add an experiments_mapping ensembler test?

terryyylim (approved):
LGTM, thanks @deadlycoconuts ! 🚀  Left some comments to be addressed.

sdk/e2e/09_deploy_router_with_route_name_path_std_ensembler_test.py
krithika369: It's a good question. Probably doesn't make sense to have a test for the "route name path" specifically. But we don't have any other e2e test for the standard ensembler at the moment. So I think it's useful to add this. Likewise, we have a test case internally using experiment_mappings, with the proprietary experiment engine.

.github/workflows/turing.yaml
krithika369, line 30: If only the SDK files are modified, why should we run the whole turing workflow? Is it for the e2e tests?

sdk/turing/router/config/router_ensembler_config.py
krithika369: > If None is configured, typically the typing used should be Optional, rather than a List in this case

Agree with this - Optional is more appropriate.

sdk/turing/router/config/router_ensembler_config.py
krithika369: I have the same question about why we need to manipulate the value here - can we just use `None` for experiment_mappings and route_name_path if they are not set? The API doesn't require them to be sent in, right?

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: > Why not just configure the default [] value in the parameters?

I believe there's an issue with using mutable default arguments in Python, which can potentially lead to unexpected behaviour - this causes warnings to pop on certain IDEs:
https://medium.com/geekculture/do-not-use-objects-as-default-arguments-in-python-1c940212db2e

Thanks for the reminder on `Optional`, I'll add that in to the type hint.

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: Wow thanks for catching that!

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: Thanks for spotting that! Strange black didn't fix that 🤔

sdk/e2e/09_deploy_router_with_route_name_path_std_ensembler_test.py
deadlycoconuts: > And, by this logic, shouldn't we also add an experiments_mapping ensembler test?
> But we don't have any other e2e test for the standard ensembler at the moment.
We do!! There's an e2e test that uses the standard ensembler with the proprietary experiment engine (it's found in the blandly named `01_create_router_test` file). We can perhaps rename that `01` test with a more explicit name to make it appear more explicitly as a test using the standard ensembler with experiment mappings if that helps 🤔

> Is this test necessary?
Well, it is an SDK mirror of the `09_deploy_router_with_route_name_path_std_ensembler_test` that we introduced for the API and so far we've decided to maintain both sets of e2e tests.

.github/workflows/turing.yaml
deadlycoconuts: Yep, it's to trigger the e2e tests (we want to check if the existing workflows still work after changes to the SDK are made) but come to think of it, we can perhaps create a new `turing/e2e/**` directory and dump both the API e2e and SDK tests there, if that helps managing the CI/CD workflows more easily?

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: > I believe an empty string "" is required because the API validator requires this field to be set right?

> can we just use None for experiment_mappings and route_name_path if they are not set? The API doesn't require them to be sent in, right?

Yeah the OpenAPI classes needs some value to be set and we can't just pass a `None` value directly. One way to go around this is to selectively pass non-`None` values into the `kwargs` dict:

```python
        kwargs = {}
        if self.experiment_mappings is not None:
            kwargs["experiment_mappings"] = self.experiment_mappings
        if self.route_name_path is not None:
            kwargs["route_name_path"] = self.route_name_path

        return turing.generated.models.EnsemblerStandardConfig(
            **kwargs,
        )
```

I've refactored the `StandardRouterEnsemblerConfig` class a little so we aren't manipulating the values of `route_name_path` in the `to_open_api` method but I've also changed the default values of `experiment_mappings` and `route_name_path` members to be `None`. 

> Also, in what cases will this (`route_name_path`) value be None[?]

In general, one of `route_name_path` or `experiment_mappings` will be `None` but not the other. This corresponds to the scenario when the user only wants the standard ensembler to be explicitly by the experiment mappings OR the route name path. 

Note also that the SDK doesn't perform any validations to ensure one and only of these two fields are set, but the API will perform this and return an error if this property is violated.

sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -487,14 +514,22 @@ def to_open_api(self) -> OpenApiModel:
...
         Method to create a new standard ensembler
 
+        :param route_name_path: configured route name path that points to the route name within a given treatment config
         :param experiment_mappings: configured mappings between routes and treatments
+        :param fallback_response_route_id: configured final response route to be used as a fallback
         """
-        self.experiment_mappings = experiment_mappings
+        self.route_name_path = route_name_path
+        self.experiment_mappings = (
+            experiment_mappings  # sif experiment_mappings else None
```
terryyylim, line 531: Why not just `self.experiment_mappings = experiment_mappings`? And, I think the comment is self-explanatory.

api/api/openapi.bundle.yaml
```diff
@@ -2350,8 +2350,10 @@ components:
         experiment_mappings:
           items:
             $ref: '#/components/schemas/EnsemblerStandardConfig_experiment_mappings'
+          nullable: true
           type: array
         route_name_path:
+          nullable: true
```
deadlycoconuts, line 2356: These two `nullable` conditions have to be added to force the autogenerated openapi classes to accept `None` values for both `experiment_mappings` and `route_name_path`.

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: Ahah thanks for catching the comment that's not meant to be a comment 😅

sdk/turing/router/config/router_ensembler_config.py
deadlycoconuts: > Yeah the OpenAPI classes needs some value to be set and we can't just pass a None value directly.

I've managed to change this by setting the `nullable` field as `True` within the OpenAPI specs for both `experiment_mappings` and `route_name_path`.

sdk/e2e/09_deploy_router_with_route_name_path_std_ensembler_test.py
krithika369: >  We do!! There's an e2e test that uses the standard ensembler with the proprietary experiment engine

It's using the docker ensembler, no?

https://github.com/caraml-dev/turing/blob/main/api/e2e/test/testdata/create_router_nop_logger_proprietary_exp.json.tmpl#L77

deadlycoconuts: Oh wait, oops I got ensemblers confused with the proprietary experiment engine 😅

deadlycoconuts: But it wouldn't take too much effort to set up an e2e test with the proprietary experiment engine and a standard ensembler 😅

krithika369: I think the e2e tests are not meant to test every combination of values (that's expected to be done by the unit/integration tests). So, IMO renaming this test to `09_deploy_router_with_std_ensembler_test`, to have it represent std ensembler in general, may suffice.

deadlycoconuts: Okay I've just renamed the test :)

deadlycoconuts:
Thanks everyone for the comments! I'll be merging this now! :)

PR #252, opened 2022-09-27
Bugfix: Fix tailLines filter for logs viewer

krithika369:
With the recent upgrade of Eui, the search bar's `field_value_selection` with single-select no longer works as expected. Logged it here: https://github.com/elastic/eui/issues/6271

However, `multiSelect` set to and/or works and since we want it to behave as a single-select, we can simply take the latest value from the filters. The downsides:
* The [autoClose](https://elastic.github.io/eui/#/forms/search-bar#search-bar-filters) property will be ignored and the popover will remain open after a selection, until it is dismissed.
* The selected option is not highlighted in the list

These are deviations from the old behavior but IMO, it's workable until the original issue is fixed.

https://user-images.githubusercontent.com/23465343/192650834-a37eacae-dea5-4fe1-9a14-b1f958d784bb.mov

### DIFF
--- a/ui/src/components/pod_logs_viewer/PodLogsViewer.js
+++ b/ui/src/components/pod_logs_viewer/PodLogsViewer.js
@@ -7,6 +7,7 @@ import {
 } from "@elastic/eui";
 import { LazyLog, ScrollFollow } from "react-lazylog";
 import { slugify } from "@gojek/mlp-ui";
+import isArray from "lodash/isArray";
 
 import "./PodLogsViewer.scss";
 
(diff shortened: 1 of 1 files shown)


ui/src/components/pod_logs_viewer/PodLogsViewer.js
```diff
@@ -30,7 +31,9 @@ export const PodLogsViewer = ({
         name: !!query.tail_lines
           ? `Last ${query.tail_lines} records`
           : "From the container start",
-        multiSelect: false,
+        // single select behavior is buggy in recent versions of Eui
+        // See: https://github.com/elastic/eui/issues/6271
```
deadlycoconuts, line 35: Whoa, thanks for creating an issue on the elastic eui repo for this!

deadlycoconuts (approved):
LGTM! Thanks for looking into this!  🚀

PR #253, opened 2022-09-28
Bugfix: Fix diff page to display router rules

deadlycoconuts:
## Context
Previously the router diff page does not display the router rules even when changes to them are made. This PR thus introduces a tiny fix that allows the rules configured to be displayed on the diff page.

An additional fix to address https://github.com/caraml-dev/turing/pull/251#discussion_r981861582 has also been made to ensure that only the `route_name_path` or `experiment_mappings` field must be set within the `standard_ensembler` map.

### DIFF
--- a/ui/src/router/components/form/validation/schema.js
+++ b/ui/src/router/components/form/validation/schema.js
@@ -247,11 +247,22 @@ const mappingSchema = yup.object().shape({
   route: yup.string().required("Treatment needs to be mapped back to a route"),
 });
 
-const standardEnsemblerConfigSchema = yup.object().shape({
-  route_name_path: yup.string().nullable(),
-  experiment_mappings: yup.array(mappingSchema).nullable(),
-  fallback_response_route_id: validRouteSchema,
-});
+const standardEnsemblerConfigSchema = yup
+  .object()
+  .shape({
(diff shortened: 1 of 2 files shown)


ui/src/router/components/form/validation/schema.js
```diff
@@ -247,11 +247,22 @@ const mappingSchema = yup.object().shape({
   route: yup.string().required("Treatment needs to be mapped back to a route"),
 });
 
-const standardEnsemblerConfigSchema = yup.object().shape({
-  route_name_path: yup.string().nullable(),
-  experiment_mappings: yup.array(mappingSchema).nullable(),
-  fallback_response_route_id: validRouteSchema,
-});
+const standardEnsemblerConfigSchema = yup
```
krithika369, line 250: When this test fails, where would it be captured in the form? Could you add a screenshot?

krithika369 (approved):
LGTM! Thanks for these fixes!

ui/src/router/components/form/validation/schema.js
deadlycoconuts: Oops it wouldn't be captured anywhere as of now. In fact, the test wouldn't ever fail because of the automatic appending of the path prefix.

PR #254, opened 2022-09-29
Fix v1.7.0-rc1 bugs

deadlycoconuts:
## Context
With `v1.7.0-rv1` released, several bugs where observed and this PR serves to introduces bug fixes to address them:

- Router version diff page crashing whenever traffic rules are absent 
   - Traffic rules (including the default rule) are only added to the pretty print only when they are present

### DIFF
--- a/ui/src/services/version/RouterVersion.js
+++ b/ui/src/services/version/RouterVersion.js
@@ -163,14 +163,14 @@ export class RouterVersion {
     };
 
     if (this.default_traffic_rule) {
-      pretty.router["default_traffic_rule"] = {routes: this.default_traffic_rule.routes.join()};
+      pretty.router["default_traffic_rule"] = {routes: this.default_traffic_rule.routes.join(", ")};
     }
     if (this.rules?.length > 0) {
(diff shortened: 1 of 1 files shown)


krithika369 (approved):

ui/src/services/version/RouterVersion.js
```diff
@@ -168,6 +161,19 @@ export class RouterVersion {
...
+
+    if (this.default_traffic_rule) {
+      pretty.router["default_traffic_rule"] = {routes: this.default_traffic_rule.routes.join(", ")};
+    }
+    if (this.rules?.length > 0) {
+      pretty.router["rules"] = this.rules.map((rule) => ({
+        conditions: rule.conditions.map((condition) => ({
+          [condition.field_source]: `${condition.field} ${condition.operator.toUpperCase()} [${condition.values.join(", ")}]`
+        })),
+        routes: rule.routes.join(", "),
```
deadlycoconuts, line 173: Added a comma and a space here to split the routes/rule/condition names apart!

PR #255, opened 2022-09-29
[UPI] Api Server

leonlnj:
### Description
Addition of capability to deploy [UPI router](https://github.com/caraml-dev/turing/pull/238) via API Server. 

Overall the concept is very simple, to add `Protocol` to `RouterVersion`. Based on the flag, the endpoint/knative/services/env vars will be configured accordingly to run the UPI-enabled routers. Fiber yaml for individual routes will be inferred base on the same flag.

No new mandatory inputs are required, HTTP will serve as default.

CI can be viewed [here](https://github.com/leonlnj/turing/actions/runs/3225173564/jobs/5277199321).

### Changes
- API Specs
Addition of `Protocol` to `RouterVersion` to reflect the type of Router. 

- API Server
  - Using `Protocol` fields supplied for Router to handle UPI/HTTP. 
  - Adding knative `h2c` for http2 connection in `api/turing/cluster/knative_service.go`. Liveliness probe is not added for now due to a [bug in knative <1.2.3](https://github.com/knative/serving/pull/12479/files). Will be defer to a later date.
  - HTTP prefixed not configured for UPI `api/turing/cluster/servicebuilder/router.go`
  - Router endpoint configured  based on UPI/HTTP in `api/turing/service/router_deployment_service.go`

- New column for DB
Addition of `Protocol` to `router_versions` in `router_versions` Table

- E2E
Additional simple UPI test that created a mock UPI route, and making a call via Turing UPI Router

### Tasks
- [X] Combine service and method in Fiber
- [x] Merge Fiber PR to main and update `go.mod`
- [X] Add protocol to `Router`
- [X] Add protocol to `Router.Routes`
- [X] Add db migration script. Default existing routes and router to HTTP
- [x] Deploy grpc Router with Knative
- [x] Modify hardcoded endpoint to grpc host
- [x] Update SDK for compatibility
- [x] Add test image in upi repo
- [x] E2e Test for sdk and api server with UPI echo server as route

### Deferred
- Prom metrics integration with `h2c` knative
- liveness probe implementation and knative upgrade with bugfix for liveness probe for `h2c` knative


engines/router/missionctl/server/application.go
```diff
@@ -67,26 +67,31 @@ func Run() {
 		}
 
 		s := upi.NewUPIServer(missionCtl)
-		if cfg.AppConfig.CustomMetrics {
-			m := cmux.New(l)
-			grpcL := m.Match(cmux.HTTP2HeaderField("content-type", "application/grpc"))
-			httpL := m.Match(cmux.HTTP1Fast())
+		m := cmux.New(l)
+		grpcL := m.Match(cmux.HTTP2HeaderField("content-type", "application/grpc"))
```
pradithya, line 71: @leonlnj Is the `content-type` case-sensitive? I was wondering whether the request to metrics server routed to incorrect handler due to this.

api/turing/models/router.go
```diff
@@ -60,17 +58,3 @@ func (r *Router) ClearCurrRouterVersion() {
 	r.ClearCurrRouterVersionID()
 	r.CurrRouterVersion = nil
 }
-
-// RouterResponse is an alias for the Router, to enable custom marshaling
-type RouterResponse Router
-
-// MarshalJSON is a custom marshaling function for the Router, to manipulate the
-// endpoint.
-func (r *Router) MarshalJSON() ([]byte, error) {
```
leonlnj, line 69: removing it as it append `/v1/predict` for all cases and appending it where `http` is constructed instead in `api/turing/service/router_deployment_service.go`

api/api/specs/routers.yaml
```diff
@@ -580,6 +582,18 @@ components:
         - "pending"
       default: "pending"
 
+    RouteProtocol:
```
pradithya, line 585: I think `RouteProtocol` vs `RouterProtocol` will be confusing to differentiate, can we find better name? And also, we should differentiate `protocol` field in `RouterConfig` and the `protocol` field in `Route` since both are called `protocol` but accept different type.

leonlnj: how about changing `RouterProtocol` to `RouterProtocolType` and `router_protocol_type` in json, since its more router specific as well? 

While keeping the one for `Route` same since `endpoint`, `protocol`, `service_method` seems to more generic.

pradithya: Maybe `RouterType` ? So we have `HTTP` router and `GRPC` router.

api/turing/cluster/servicebuilder/router_test.go
```diff
@@ -64,7 +64,7 @@ func TestNewRouterService(t *testing.T) {
 	// Define tests
 	tests := map[string]testSuiteNewService{
 		"success | basic": {
-			filePath:     filepath.Join(testDataBasePath, "router_version_basic.json"),
```
pradithya, line 67: Is there any reason to modify existing test instead of adding new set of test for upi?

api/turing/cluster/knative_service.go
```diff
@@ -125,18 +125,25 @@ func (cfg *KnativeService) buildSvcSpec(
 	resourceReqs := cfg.buildResourceReqs(cfg.UserContainerLimitRequestFactor)
 
 	// Build container spec
+	var containerName string
```
pradithya, line 128: should be `portName`

sdk/turing/generated/model/route.py
```diff
@@ -26,6 +26,10 @@
     validate_get_composed_info,
 )
 
+def lazy_import():
```
pradithya, line 29: Curious, why do we need this?

pradithya: Oh, this is generated code

api/api/specs/routers.yaml
leonlnj: I thought of that but it could refer to many different things without context. No strong opinion either way

api/turing/cluster/servicebuilder/router_test.go
leonlnj: most of other test cases already covered the http service building logic, I try to refrain from adding more json config tests to keep it leaner as its pretty hard to work with the current size of files and there are a lot of object construct in place here, hence the decision to turn use it as the grpc basic. 

No strong opinion, could always add another one 😅

api/turing/service/router_deployment_service_test.go
```diff
@@ -248,7 +248,7 @@ func TestDeployEndpoint(t *testing.T) {
 	)
 
 	assert.NoError(t, err)
-	assert.Equal(t, fmt.Sprintf("http://%s-router.models.example.com", routerVersion.Router.Name), endpoint)
+	assert.Equal(t, fmt.Sprintf("http://%s-router.models.example.com/v1/predict", routerVersion.Router.Name), endpoint)
```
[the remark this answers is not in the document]
leonlnj: ah this is due to `https://github.com/caraml-dev/turing/pull/255/files#r987485618`, previously it was done at the json unmarshalling which cause all endpoint to be appended with such, so that logic is moved to the service and constructed here instead.

Although this tests expects `http://%s-router.models.example.com` previously, you will never get it from api calls as during unmarshalling it will become `http://%s-router.models.example.com/v1/predict`

api/turing/cluster/servicebuilder/router_test.go
pradithya: I think it's better to add new one.

api/api/specs/routers.yaml
leonlnj: on second thought, while keeping Route Protocol intact to be in sync with Fiber struct but changing `RouterProtocol` to `RouterType` will be confusing since Merlin is using `Protocol` for http/upi/grpc, which is identical to `RouterType` but not `Protocol` of Fiber (GRPC/HTTP only, no UPI).

Changing `Protocol` of Fiber to `RouteType` is confusing as well, as there is already a `type` field to indicate the type of Fiber component. 

Fiber.Protocol = HTTP/GRPC
Merlin.Protocol = Router.Protocol = HTTP/GRPC/UPI

In short, if we sync with Merlin, it will clash with Fiber since both `Protocol` means different thing in context.

What I can propose is to change Fiber and Routes (although `protocol` make total sense within that context) to AppProtocol.

For router, to sync with merlin
Protocol RouterProtocol `json:"protocol"` -> Protocol Protocol `json:"protocol"` 
For Fiber and Routes (turing model)
Protocol RouteProtocol `json:"protocol"` -> AppProtocol AppProtocol `json:"app_protocol"` 
Fiber to rename `Protocol` type to `AppProtocol`, likewise throughout the config.

wdyt?

pradithya: Ah I see.

I agree with the first part of the proposal. Let's make it consistent with Merlin fo the `RouterProtocol`.

On the other hand,  for fiber protocol

> Protocol RouteProtocol json:"protocol" -> AppProtocol AppProtocol json:"app_protocol"
Fiber to rename Protocol type to AppProtocol, likewise throughout the config.

Does it make sense to call it `ServiceType` ? so we have `HTTP` service and `GRPC` service at fiber level? Other alternatives: `AppType`, `RouteProtocol`.

engines/router/missionctl/server/application.go
leonlnj: yes it is but from what i searched and learnt, http2 headers are all lower-case before encoded by design, so it should be safe.

https://www.rfc-editor.org/rfc/rfc7540#section-8.1.2
```
 Just as in HTTP/1.x, header field names are strings of ASCII
 characters that are compared in a case-insensitive fashion.  However,
 header field names MUST be converted to lowercase prior to their
 encoding in HTTP/2.
```
https://evertpot.com/http-2-finalized/

api/turing/models/router_version.go
```diff
@@ -15,6 +16,13 @@ const (
 	RouterVersionStatusUndeployed RouterVersionStatus = "undeployed"
 )
 
+type RouterProtocol config.Protocol
```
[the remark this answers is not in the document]
leonlnj: ah good question. I was thinking to do that or not because `config.Protocol` comes from the router's code. 

I was thinking to wrap it here so the api server code uses the type in models package instead of depending directly on the router. 

Probably overthinking, i will use `config.Protocol` directly

api/api/specs/routers.yaml
leonlnj: asked for a couple of opinions, `RouteProtocol` still made the most sense. Although having `route` appended looks weird it itself, cause usually for naming, we do `animal.name` instead of `animal.animal_name`. Even if two object has the same field, there is never a guarantee for them to be the same type/struct. Either way to me its just different kind of confusion, regardless user should always follow the swagger doc and input accordingly 😅 .

If no more opinion will change to this 😅 
For router, to sync with merlin, `config.Protocol` is define in router code [upi or http].
`Protocol RouterProtocol json:"protocol"` -> `Protocol config.Protocol json:"protocol"`
For Routes (turing model) 
`Protocol RouteProtocol json:"protocol"` -> `RouteProtocol RouteProtocol json:"route_protocol"`
Fiber will continue to use `protocol` in config and code (looks really weird if its `request.route_protocol()` instead of `request.protocol()` )

.github/workflows/turing.yaml
```diff
@@ -389,6 +389,7 @@ jobs:
           GOPATH: ${{ github.workspace }}/api/.go
           TEST_ID: ${{ github.run_id }}
           MOCKSERVER_ENDPOINT: http://mockserver
```
krithika369, line 391: Nit: Shall we call this `MOCKSERVER_HTTP_ENDPOINT` for consistency? Similar comment about the change in `api/e2e/test/00_all_e2e_test.go`.

api/api/specs/routers.yaml
krithika369: Do we need to store the route protocol? If we assume that protocols will not be mixed, can we infer the route protocol from that of the router at deploy time (HTTP_JSON -> http, UPI_V1 / CUSTOM_GRPC -> grpc)?

engines/router/missionctl/server/application.go
```diff
@@ -67,26 +67,31 @@ func Run() {
...
+		httpL := m.Match(cmux.Any())
 
-			mux := http.NewServeMux()
+		mux := http.NewServeMux()
+		mux.Handle("/v1/internal/", http.StripPrefix(
+			"/v1/internal",
+			handlers.NewInternalAPIHandler([]string{
```
krithika369, line 77: Why has this been added?

engines/router/missionctl/server/application.go
```diff
@@ -67,26 +67,31 @@ func Run() {
...
-			go s.Run(grpcL)
-			go func() {
-				if err := httpS.Serve(httpL); err != nil {
-					log.Glob().Errorf("Failed to serve http server: %s", err)
-				}
-			}()
-			if err := m.Serve(); err != nil {
-				log.Glob().Errorf("Failed to serve cmux: %s", err)
+		}
+		httpS := &http.Server{Handler: mux}
```
krithika369, line 85: Super nit: I was wondering why this is `https`. 😅  May be `httpServer` would be clearer?

leonlnj: oops, copied it from the mux readme, will do so 👍

leonlnj:
> Do we need to store the route protocol? If we assume that protocols will not be mixed, can we infer the route protocol from that of the router at deploy time (HTTP_JSON -> http, UPI_V1 / CUSTOM_GRPC -> grpc)?
regarding this, 
> https://github.com/caraml-dev/turing/pull/255#discussion_r991768061

I thought of that but decided to add it to the model for it to be 100% reflective of the fiber.yaml's route section. We could drop it (removes confusion with `protocol`) as well actually, now that i think of it.

api/api/specs/routers.yaml
leonlnj: I thought of that but decided to add it to the model for it to be 100% reflective of the fiber.yaml's route section. We could drop it (removes confusion with protocol) as well actually, now that i think of it.

engines/router/missionctl/server/application.go
leonlnj: oops my bad, will remove the enricher/ensembler uri

"/v1/internal/" health handler was added for queue-proxy probe

sdk/turing/router/config/route.py
```diff
@@ -37,7 +40,6 @@ def endpoint(self) -> str:
 
     @endpoint.setter
     def endpoint(self, endpoint: str):
-        Route._verify_endpoint(endpoint)
```
leonlnj, line 40: validation is now done at `RouteConfig` init instead, since there is no more `Protocol` field for Routes.

sdk/turing/router/config/router_config.py
```diff
@@ -122,15 +136,14 @@ def routes(self) -> List[Route]:
...
-                self._routes = routes
-            elif all(isinstance(route, dict) for route in routes):
-                self._routes = [Route(**route) for route in routes]
-            else:
-                self._routes = routes
-        else:
-            self._routes = routes
+        if isinstance(routes, list) and all(isinstance(route, dict) for route in routes):
+            routes = [Route(**route) for route in routes]
+        for route in routes:
```
leonlnj, line 141: validation moved here

pradithya:
@leonlnj How do you run the CI for this PR?

leonlnj:
> @leonlnj How do you run the CI for this PR?

manual `workflow_dispatch` in [fork](https://github.com/leonlnj/turing/actions/runs/3225173564/jobs/5277199321)

its bit awkward as the CI doesn't run automatically if its not merging into `main`

image

pradithya:
@leonlnj You can add this to run the CI in this PR https://github.com/caraml-dev/turing/blob/25d91bfeb302b9aede958eb486b6569afa08a68d/.github/workflows/turing.yaml#L26

pradithya (approved):
LGTM! Thanks @leonlnj . Let's merge it once the CI passed. 🥳

PR #256, opened 2022-10-03
Fix Autoscaling config info in the Pyfunc ensembler view

terryyylim:
Incorrect typo of `dockerConfig` which should have been `pyfuncConfig`.


deadlycoconuts (approved):
LGTM! Thanks for the quick fix! 🚀

krithika369:
@terryyylim thanks for fixing this. One nit: let's make the PR title a little more descriptive so its scope is clearer in the release notes. Something like: `Fix Autoscaling config info in the Pyfunc ensembler view` ?

PR #257, opened 2022-10-03
Bugfix: Fix RFC-1123 hostname string validation for ensembler names

deadlycoconuts:
## TL;DR
Problem: Users create ensembler names with underscores -> Invalid Kaniko job names -> Invalid K8s jobs -> Router deployments fail

Solution: Add additional API checks to block usage of underscores -> Discovery of bug in existing `go-validator` validation tag to validate ensembler names -> Update `go-validator` -> Done

## Context
Pyfunc ensemblers are built as a web service and containerised into a single image upon being deployed as part of a router deployment. This image building process takes place in Kaniko builders run as _jobs_ on a Kubernetes cluster. The Turing API names these jobs automatically, in a way which [incorporates](https://github.com/caraml-dev/turing/blob/main/api/turing/imagebuilder/imagebuilder.go#L133) the user-registered name of the ensemblers to be built:

```go
kanikoJobName := ib.nameGenerator.generateBuilderName(
	request.ProjectName,
	request.ResourceName,  // this refers to the name of the ensembler
	request.ResourceID,
	request.VersionID,
)
```
where the method `generateBuilderName` is given by:

```go
func (n *ensemblerJobNameGenerator) generateBuilderName(
	projectName string,
	modelName string,
	modelID models.ID,
	versionID string,
) string {
	partialVersionID := getPartialVersionID(versionID, 5)
	return fmt.Sprintf("batch-%s-%s-%d-%s", projectName, modelName, modelID, partialVersionID)
}
```
Errors thus occur when users define their ensembler names with underscores (`_`) , because this causes the job names generated to similarly contain them, rendering the job names of the Kaniko builders [**invalid**](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) when run in a Kubernetes cluster, since Kubernetes requires the job names to fulfil standards defined in RFC 1123, i.e.

- contain only lowercase alphanumeric characters or '-'
- start with an alphanumeric character
- end with an alphanumeric character

Router deployments using ensemblers that have underscores in their names thus fail immediately since the image building jobs get rejected instantly.

## Fix
The original intention of the fix was to introduce additional API checks to validate the ensembler names submitted by users and then return appropriate errors to prevent the above mentioned errors from occurring. 

However, it was found that the existing `GenericEnsembler` struct that gets created when handling ensembler creation/update API calls already contains the necessary `go-playground` validator `hostname_rfc1123` [validation](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Hostname_RFC_1123) for those RFC 1123 standards on the `Name` field:

```go
type GenericEnsembler struct {
	Model

	ProjectID ID `json:"project_id" gorm:"column:project_id"`

	Type EnsemblerType `json:"type" gorm:"column:type" validate:"required,oneof=pyfunc"`

	Name string `json:"name" gorm:"column:name" validate:"required,hostname_rfc1123,lte=20,gte=3"`
}
```

Clearly, there was an apparent bug with the `hostname_rfc1123` validation tag that led to the observed bug that this PR is trying to address. Upon further investigation, it was found that the existing version (`v10.9.0`) of the `go-playground` validator used by the Turing API does not contain the necessary bug fix for the `hostname_rfc1123` validation tag. 

Fortunately, all that's needed is to update its version to `v10.11.0`, which [includes](https://github.com/go-playground/validator/releases/tag/v10.11.0) the required [bug fix](https://github.com/go-playground/validator/pull/912) (PR #912). And since `v10.11.1` is the newest release with an additional bug fix, why not just update the dependency to the latest version? 🙃

## Miscellaneous
This PR also makes a tiny modification to an incorrect docstring mentioning Kaniko jobs as 'pods'.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/caraml-dev/turing/engines/router v0.0.0
 	github.com/getkin/kin-openapi v0.76.0
 	github.com/ghodss/yaml v1.0.0
-	github.com/go-playground/validator/v10 v10.9.0
+	github.com/go-playground/validator/v10 v10.11.1
 	github.com/gojek/fiber v0.0.0-20201008181849-4f0f8284dc84
 	github.com/gojek/merlin v0.0.0
 	github.com/gojek/mlp v1.4.7
@@ -151,7 +151,7 @@ require (
 	go.mongodb.org/mongo-driver v1.1.2 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
 	go.uber.org/multierr v1.6.0 // indirect
-	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
+	golang.org/x/crypto v0.0.0-20211215153901-e495a2d5b3d3 // indirect
 	golang.org/x/net v0.0.0-20220722155237-a158d28d115b // indirect
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4 // indirect
 	golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f // indirect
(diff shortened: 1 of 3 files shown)


krithika369 (approved):
LGTM. Thanks for tracking this down! 🎉

PR #258, opened 2022-10-07
Use initContainer for downloading experiment engine plugin binaries

deadlycoconuts:


.github/workflows/turing.yaml
```diff
@@ -41,8 +41,8 @@ env:
   GO_LINT_VERSION: v1.48.0
   CLUSTER_NAME: turing-e2e
   ISTIO_VERSION: 1.9.9
-  KNATIVE_VERSION: v0.18.3
-  KNATIVE_ISTIO_VERSION: v0.18.1
+  KNATIVE_VERSION: 1.0.1
+  KNATIVE_ISTIO_VERSION: 1.0.0
```
deadlycoconuts, line 45: Just some bumping to the version numbers.

api/turing/api/request/request_test.go
```diff
@@ -65,7 +65,7 @@ func makeTuringExperimentConfig(clientPasskey string) json.RawMessage {
 	return expEngineConfig
 }
 
-var defaultRouteID string = "default"
+var defaultRouteID = "default"
```
deadlycoconuts, line 68: Removed the redundant type declared after the variable name. I've repeated this a couple of times across the codebase to address prompts from our linters.

infra/cluster-init/config-features.yaml
```diff
@@ -0,0 +1,14 @@
...
+  namespace: knative-serving
+  labels:
+    app.kubernetes.io/name: knative-serving
+    app.kubernetes.io/component: controller
+    app.kubernetes.io/version: devel
+  annotations:
+    knative.dev/example-checksum: "4d5feafc"
+data:
+  kubernetes.podspec-volumes-emptydir: "enabled"
+  kubernetes.podspec-init-containers: "enabled"
```
deadlycoconuts, line 14: These are the two flags that need to be set to enable those two resources.

krithika369: Super nit: can we name this `knative-configmap.yaml` instead, so it's clear from the filename what it's for?

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -85,10 +85,6 @@ type ClusterServiceBuilder interface {
 		secretName string,
 		fluentdConfig *config.FluentdConfig,
 	) *cluster.KubernetesService
-	NewPluginsServerService(
```
krithika369, line 88: I think this method still features in `api/turing/service/router_deployment_service_test.go` and possibly elsewhere - we can clean it up.

api/turing/service/router_deployment_service.go
```diff
@@ -283,21 +269,6 @@ func (ds *deploymentService) UndeployRouterVersion(
 		}
 	}
 
-	// Delete experiment engine plugins server
-	if routerVersion.ExperimentEngine.PluginConfig != nil {
```
krithika369, line 287: Since we have existing deployments that use this component, I think we should still attempt to delete it during the uninstallation (but ignore any errors in case the resource is not found). Let's add a TODO in the code to remove this logic and we can do so after a few months.

krithika369 (approved):
Thanks for this PR. It's nicely done. Left some small comments.

infra/cluster-init/config-features.yaml
deadlycoconuts: Hmm, or should I just create a folder named `knative-configmaps` and then place this file (still named `config-features.yaml`) inside? Apparently there are like over 10 different Knative configmaps (like `config-autoscaler`, `config-defaults`, `config-logging`, `config-network`, etc.) and it `knative-configmap` might sound a little too generic 👀

api/turing/cluster/servicebuilder/service_builder.go
deadlycoconuts: Whoops, okay let me go hunt them down 😶

api/turing/service/router_deployment_service.go
deadlycoconuts: Ahh that's a very good point; ok I'll add this method back and do the necessary error handling for this :)

deadlycoconuts: I've changed the error handling a little and turned the `NewPluginsServerService` method into a simple function so we don't need to keep the method as part of the service builder interface. When we're ready to remove this deletion logic, we can just remove this code block as well as the file containing the `NewPluginsServerService` function.

infra/cluster-init/config-features.yaml
krithika369: That works too!

deadlycoconuts:
Thanks for the comments and the quick review! I'll be merging this! 🚀

PR #260, opened 2022-10-12
Minor UI Fixes for Autoscaling Policy

krithika369:
- Correct the order of Resources panel vs Autoscaling Policy panel in the form / detail view components so that the Autoscaling policy always appears after the Resources.
- `ui/src/router/components/form/components/autoscaling_policy/typeOptions.js` - Add default target values that make sense for the metric. These will be applied in `ui/src/router/components/form/components/autoscaling_policy/AutoscalingPolicyPanel.js`.

https://user-images.githubusercontent.com/23465343/195338127-3aa293ac-7060-4453-9759-c4402abb362d.mov


leonlnj (approved):
LGTM!

PR #261, opened 2022-10-12
Upgrade gorm dependency in Turing API

krithika369:
Upgrading to Gorm v2: https://gorm.io/

Changes similar to: https://github.com/caraml-dev/xp/pull/43

### Summary of Changes
* `api/turing/service/router_service.go` - Use upsert in the place of Create / Update
* `api/turing/service/router_version_service.go` - Use router version ID to test Create / Update. Where foreign key record creation is involved (Enricher / Ensembler), upsert cannot be used. Previously, the `NewRecord` function was testing the same thing. See: https://github.com/go-gorm/gorm/issues/3400 and https://github.com/go-gorm/gorm/issues/1044
* `Count()` query works with `int64` now, as opposed to `int`.
* `api/turing/service/ensembler_service.go`, `api/turing/service/ensembling_job_service.go` - The concurrent count query has been modified to serial execution as it sometimes causes data race in tests. As such, the gain may not be significant.
* `api/turing/models/pyfunc_ensembler.go` - Update `BeforeCreate` to match the new signature - `gorm.Scope` has been removed.


api/turing/service/router_version_service.go
```diff
@@ -113,7 +113,9 @@ func (service *routerVersionsService) ListRouterVersionsWithStatus(
 func (service *routerVersionsService) Save(routerVersion *models.RouterVersion) (*models.RouterVersion, error) {
 	var err error
 	tx := service.db.Begin()
-	if service.db.NewRecord(routerVersion) {
+	// For create vs update, there is no default gorm API. Upsert will not help for
+	// foreign key associations (enricher / ensembler). So, we naively compare ID==0.
+	if routerVersion.ID == 0 {
```
deadlycoconuts, line 118: Can we leave a comment mentioning what `ID == 0` means? i.e. when this router does not exist or something of the like? 😅

deadlycoconuts (approved):
LGTM! Just left a tiny comment but otherwise everything looks good! Thanks for updating this! 🚀

PR #264, opened 2022-10-14
Bugfix: Fix UI crashes when changing between experiment engines

deadlycoconuts:
## Context
This PR addresses 2 main bugs that have been found with the UI:

1. Currently when editing a router, if a user changes from a custom experiment engine to a standard engine, both of which use Standard Ensemblers, the UI crashes because it supposedly is unable to locate the list of experiments configured for the standard experiment engine.
Screenshot 2022-10-17 at 3 30 05 PM
This in reality, happens only momentarily when an [effect](https://github.com/caraml-dev/turing/blob/345e9b001adaa7da95b0f99757ce15672711c1f0/ui/src/router/components/form/steps/ExperimentStep.js#L27) is triggered to reset the experiment config when a different experiment engine is selected. During this short span of time, the `experiments` field within the experiment config is undefined and hence the UI crashes. Hence, a fallback array is passed to the `StandardEnsemblerPanel` component if that field is undefined. Additionally, an additional step has been added to the effect to reset both the `route_name_path` and `experiment_mappings` fields when a different experiment engine is selected with a standard ensembler already configured.

2. If a user unsets the experiment engine field on the router edit page, the UI similarly crashes because it continues to render the `StandardExperimentFormGroup` component.
Screenshot 2022-10-17 at 3 32 58 PM
This happens because `ensembler.standard_config` of the router config is still present even after the experiment engine has been unset (set to 'None'). Hence, an additional check has been added to ensure that the `StandardExperimentFormGroup` component does not get displayed when no experiment engine is selected (the config nonetheless remains in memory until the user completes the router edit/save operation).

### DIFF
--- a/infra/cluster-init/init.sh
+++ b/infra/cluster-init/init.sh
@@ -47,13 +47,13 @@ function install_knative {
     kubectl apply \
         -f "https://github.com/knative/serving/releases/download/knative-v${KNATIVE_VERSION:-1.0.1}/serving-hpa.yaml"
 
-    kubectl apply \
-        -f knative-configmaps/config-features.yaml
-
     local core_apps=("activator" "autoscaler" "controller" "webhook")
     for app in ${core_apps[@]}; do
         kubectl wait -n knative-serving --for=condition=ready pod -l app=${app} --timeout=5m
     done
+
+    kubectl apply \
+        -f knative-configmaps/config-features.yaml
 
     kubectl apply \
         -f "https://github.com/knative-sandbox/net-istio/releases/download/knative-v${KNATIVE_ISTIO_VERSION:-1.0.0}/net-istio.yaml"
--- a/ui/src/router/components/form/steps/EnsemblerStep.js
+++ b/ui/src/router/components/form/steps/EnsemblerStep.js
@@ -69,7 +69,7 @@ export const EnsemblerStep = ({ projectId }) => {
         />
       )}
 
-      {experiment_engine.type !== "nop" && ensembler.type === "standard" && (
+      {ensembler.type === "standard" && (
         <StandardEnsemblerFormGroup
           projectId={projectId}
           experimentEngine={experiment_engine}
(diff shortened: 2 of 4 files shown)


infra/cluster-init/init.sh
```diff
@@ -47,14 +47,14 @@ function install_knative {
...
-    kubectl apply \
-        -f knative-configmaps/config-features.yaml
-
     local core_apps=("activator" "autoscaler" "controller" "webhook")
     for app in ${core_apps[@]}; do
         kubectl wait -n knative-serving --for=condition=ready pod -l app=${app} --timeout=5m
     done
 
+    kubectl apply \
+        -f knative-configmaps/config-features.yaml
```
deadlycoconuts, line 56: Reordered this `cluster-init` step to ensure that the config map only gets applied when all the core components of knative have been up and running. This was causing flaky e2e cluster set up errors in the past.

ui/src/router/components/form/steps/EnsemblerStep.js
```diff
@@ -69,7 +69,7 @@ export const EnsemblerStep = ({ projectId }) => {
         />
       )}
 
-      {ensembler.type === "standard" && (
+      {experiment_engine.type !== "nop" && ensembler.type === "standard" && (
```
krithika369, line 72: Instead of doing this, can we handle this in `onChangeEngineType` in `ui/src/router/components/form/steps/ExperimentStep.js`? You are currently clearing the standard ensembler properties if the engine type is changed. In addition, can we handle enginetype becoming none?

```javascript
const onChangeEngineType = (newType) => {
    onChange("config.experiment_engine")({ type: newType });
    if (ensemblerType === "standard") {
      // Reset standard ensembler configs
      if (newType === "nop") {
           onChange("config.ensembler.type")("nop");
      }
      onChange("config.ensembler.standard_config.experiment_mappings")([]);
      onChange("config.ensembler.standard_config.route_name_path")("");
    }
}
```

krithika369 (approved):
Thanks for the fix. Left a comment. LGTM!

ui/src/router/components/form/steps/EnsemblerStep.js
deadlycoconuts: Oh that's true that makes more sense 😅 I'll make the change and then test it out again before merging this! Thanks!

PR #266, opened 2022-10-19
Add Fluentd BigQuery Image

krithika369:
This PR moves to GiHub the special fluentd image that's being used internally for forwarding logs to BigQuery, when BQ is chosen as the outcome logging destination for the router. Tags created with the prefix `fluentd/` will publish a new version of the Docker image to GHCR. A sample publish job has been triggered from the fork: https://github.com/krithika369/turing/actions/runs/3279237790

The Turing Helm base values file will be updated in a subsequent PR, to default to the new fluentd image published from this repo.


scripts/fluentd-bigquery/Gemfile
```diff
@@ -0,0 +1,4 @@
+source "https://rubygems.org"
+ruby '~> 2.6.0'
+
+gem "fluent-plugin-bigquery", "~> 2.2.0"
```
terryyylim, line 4: Likewise here, should we upgrade it?

terryyylim (approved):
Thanks @krithika369 , LGTM! Left some food-for-thought questions regarding versions.

scripts/fluentd-bigquery/Dockerfile
```diff
@@ -0,0 +1,35 @@
+FROM fluent/fluentd:v1.10-debian-1
```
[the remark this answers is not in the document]
krithika369: That's a good point. Done.

scripts/fluentd-bigquery/Gemfile
krithika369: I was worried the newer plugin version may have some breaking changes but it seems to work well for the functionality we use. Done.

PR #267, opened 2022-10-20
Refactor E2E tests and fix local E2E tests setup

terryyylim:
**Why we need this PR?**

This PR makes local E2E tests which have been broken for a long time running successfully again. The following changes have been made/added:
- Existing tests have been refactored to use Ginkgo which improves the maintainability of our E2E tests, in `Behavioural Driven Development` (BDD) framework
- Updated README.md to run E2E tests locally


.github/workflows/turing.yaml
```diff
@@ -383,20 +388,33 @@ jobs:
         working-directory: sdk
         run: make setup
 
-      - name: Run End-to-End Test Suite with Turing API
+      # - name: Run End-to-End Test Suite with Turing API
```
terryyylim, line 391: I saw UPI e2e tests using the current structure, but this PR converts test to use ginkgo. Hence, I placed ginkgo tests in `test-e2e` folder to differentiate, and we can decide what to do before this PR is merged. Any thoughts? @krithika369 @pradithya

krithika369: Since the UPI tests are in a branch at the moment, IMO we can go ahead and make the necessary changes for the test framework replacement in this PR and get it merged to `main` first. Then, when the changes from `main` are pulled back to `grpc_upi`, the UPI test cases must be adapted - this task can lie on either team depending on the bandwidth. @pradithya / @leonlnj, do you need us to pick this up once this PR is merged?

.github/workflows/turing.yaml
```diff
@@ -383,20 +388,33 @@ jobs:
...
-          TEST_ID: ${{ github.run_id }}
-          MOCKSERVER_ENDPOINT: http://mockserver
-          API_BASE_PATH: http://turing-gateway.127.0.0.1.nip.io/api/turing/v1
-          MODEL_CLUSTER_NAME: "dev"
-          PROJECT_ID: "1"
-          PROJECT_NAME: default
-          KUBECONFIG_USE_LOCAL: true
-        run: go test -v -parallel=2 ./e2e/... -tags=e2e -run TestEndToEnd
+        run: |
+          go get github.com/onsi/ginkgo/v2/ginkgo/generators@v2.3.0
```
krithika369, line 413: Can this be moved to a make target such as `setup_e2e` ?

api/README.md
```diff
@@ -91,26 +92,85 @@ watch KUBECONFIG=/tmp/kubeconfig kubectl get pod -A
...
+istio-system      istio-ingressgateway-b7ffbd9c6-kfpl2      1/1     Running   0          17h
 ```
 
+#### <b>Prepare proprietary experiment engine</b>
+---
+Next, let's get the proprietary experiment engine plugin ready!
+
+1. Build the proprietary experiment engine plugin binary
+
+When building the binary, make sure you're using the same GOOS and GOARCH that's compatible with your machine. Eg. for MacOS M1 would be `GOOS=darwin GOARCH=arm64`.
```
krithika369, line 123: Rather than documenting the differences here, should we just make the `GOOS` property an ARG to the Dockerfile which is set to `linux` by default? Can the rest of the commands be used as is? Then, the same file could be used for both local and CI envs.

api/README.md
```diff
@@ -123,41 +183,54 @@ pushd ../engines/router
...
 ```
 go run turing/cmd/main.go -config=config-dev.yaml
 ```
 
-Now run E2E tests.
-
+#### <b>Run E2E tests</b>
+---
 ```
-TEST_ID=$(date +%Y%m%d%H%M) \
```
krithika369, line 135: Is it ok to remove `TEST_ID`? I see it's still used here and it also has some README notes: `api/e2e/new-test/config/config.go`

api/config-dev.yaml
```diff
@@ -62,6 +66,11 @@ SparkAppConfig:
   TTLSecond: 86400
 RouterDefaults:
   Image: localhost:5000/turing-router
+  # Uncomment the following for running E2E tests locally OR to use proprietary experiment engine
```
krithika369, line 69: Since we can specify multiple config files on app start up (and the later ones will override the former ones), should we break the experiment engine configs into a separate file and add instructions to include it only if intending to use the experiment engine ?

api/e2e/new-test/e2e_test.go
```diff
@@ -0,0 +1,418 @@
+package e2e
```
krithika369, line 1: Why is the filename different than the internal one? Can we keep the same names so it's easier to compare later? (Similar comment about other files.)

infra/docker-compose/dev/docker-compose.yml
```diff
@@ -83,8 +83,8 @@ services:
         cp /.kube/kubeconfig.yaml /.kube/config
         sed -i 's/127.0.0.1/server/' /.kube/config
         kubectl cluster-info
-        curl -sL https://istio.io/downloadIstioctl | ISTIO_VERSION=1.12.4 sh -
-        /.istioctl/bin/istioctl install -y -f /minimal-operator.yaml
```
krithika369, line 87: Why has this config file application been removed?

infra/docker-compose/dev/knative/config-features.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
krithika369, line 1: Rather than copy the file at `infra/cluster-init/knative-configmaps/config-features.yaml`, can we use a symlink ? May be it's hard to symlink it to work with docker, in which case we can make this the original and `infra/cluster-init/knative-configmaps/config-features.yaml` the symlink.

api/README.md
terryyylim: It's ok because by default, `DefaultConfig()` in `api/e2e/new-test/config/config.go` will be triggered first when loading the configs.

infra/docker-compose/dev/docker-compose.yml
terryyylim: I'll remove `minimal-operator.yaml` file. I don't see this required anymore in the Istio documentation.

infra/docker-compose/dev/knative/config-features.yaml
terryyylim: Referenced `infra/cluster-init/knative-configmaps/config-features.yaml` instead!

api/README.md
```diff
@@ -91,26 +92,82 @@ watch KUBECONFIG=/tmp/kubeconfig kubectl get pod -A
...
+
+When building the binary, make sure you're using the same GOOS and GOARCH that's compatible with your machine.
+
+```bash
+cd engines/experiment/examples/plugins/hardcoded
+# If using MacOS M1
+make proprietary_exp_plugin GOOS=darwin GOARCH=arm64
+```
+
+2. Build the proprietary experiment engine and push it to the local registry.
```
krithika369, line 131: Can we reuse the make target `build-image` for this? We just need to pass down the correct GOOS and GOARCH from there too, like the previous `proprietary_exp_plugin`, right?

engines/experiment/examples/plugins/hardcoded/Makefile
```diff
@@ -15,4 +17,9 @@ vendor:
 build-image: vendor
 	@$(eval IMAGE_TAG = $(if $(DOCKER_REGISTRY),$(DOCKER_REGISTRY)/,)${PLUGIN_NAME}-engine-plugin:${VERSION})
 	@echo "Building docker image: ${IMAGE_TAG}"
-	docker build --tag ${IMAGE_TAG} .
\ No newline at end of file
+	docker build --tag ${IMAGE_TAG} .
+
+.PHONY: proprietary_exp_plugin
+proprietary_exp_plugin:
```
krithika369, line 23: Nit: Can we use hyphen as the separator in the target name, to match others in this file / `api/Makefile` ?

engines/experiment/examples/plugins/hardcoded/Makefile
```diff
@@ -15,4 +17,9 @@ vendor:
...
+	GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o ./bin/example-plugin ./cmd/main.go
+	mv ./bin/example-plugin ../../../../../api/bin/example-plugin
```
krithika369, line 25: Should we add a symlink with `ln -s ...` rather than move the file?

api/e2e/test/routers_suite_test.go
```diff
@@ -0,0 +1,271 @@
...
+		When(fmt.Sprintf("%s router is deployed", testCaseName), func() {
+			// Istio VirtualService configuration is applied asynchronously, so the fact that it
+			// exists in the cluster doesn't mean that it was already being in use.
+			// In the newer version of Istio (starting from v1.6), it is possible to wait for the
+			// changes readiness: https://istio.io/latest/docs/ops/configuration/mesh/config-resource-ready/
+			// However, we are currently on Istio v1.3 and there is no straight-forward
+			// way of achieving this readiness check, hence we just add an arbitrary sleep to give
+			// Istio some time to apply the changes.
+			//
+			// TODO: Remove once Turing is migrated on a newer version of Istio
```
krithika369, line 204: Just noticed this TODO - something to address, but we could do it in a separate MR later.

api/README.md
```diff
@@ -91,26 +92,82 @@ watch KUBECONFIG=/tmp/kubeconfig kubectl get pod -A
...
+
+```bash
+pushd ../experiment/examples/plugins/hardcoded
```
krithika369, line 134: I believe this doesn't match the current working directory, per step # 1.

api/README.md
```diff
@@ -91,26 +92,82 @@ watch KUBECONFIG=/tmp/kubeconfig kubectl get pod -A
...
 ```
 
+2. Build and push the image to the local registry.
+
+```bash
+docker build --no-cache -t localhost:5000/proprietary-experiment-engine-plugin
+docker push localhost:5000/proprietary-experiment-engine-plugin
+```
+
+#### <b>Prepare Turing router</b>
```
krithika369, line 169: I think the image building commands below, for the router, can also be rewritten using the existing `make build-image` target for the router, to simply the manual instructions in the README.

api/README.md
```diff
@@ -123,41 +180,54 @@ pushd ../engines/router
 popd
 ```
 
+#### <b>Start Turing API</b>
+---
 Now fire up the turing API server (as a daemon or tmux or some other terminal process).
 
 ```
-go run turing/cmd/main.go -config=config-dev.yaml
+go run turing/cmd/main.go -config=config-dev.yaml -config=config-dev-exp-engine.yaml
```
krithika369, line 188: In general, it may be nice to double-check all the paths in this README file, to see if they make sense regardless of if the previous steps in the README are run or no (so `pushd/popd` or `cd ${path} && ...` may be a good idea to capture the steps rather than standalone `cd` commands).

krithika369: Do you think we can have a single make target at the repo root that's just `e2e-local` or something which will do all of these actions described in this README at one shot? The individual commands and explanation in this README are useful but it may be a lot easier to get started if we can wrap everything into a single command to be run by the user. wdyt?

api/Makefile
```diff
@@ -81,6 +81,20 @@ version:
 	$(eval VERSION=$(if $(OVERWRITE_VERSION),$(OVERWRITE_VERSION),v$(shell ../scripts/vertagen/vertagen.sh)))
 	@echo "turing-api version:" $(VERSION)
 
+# ==================================
+# E2E tests
+# ==================================
+
+.PHONY: setup_e2e
+setup_e2e:
```
krithika369, line 89: Nit: Can the targets' names use hyphen as a separator as opposed to `_`, for consistency ?

api/README.md
terryyylim: At best, it'll do these as we'll need to run the API server separately after these steps prior to triggering the test, do you think its useful to have this?
```
.PHONY: all-local-test-steps
all-local-test-steps:
	cd ../engines/experiment/examples/plugins/hardcoded && make proprietary-exp-plugin GOOS=darwin GOARCH=arm64
	cd ../engines/experiment/examples/plugins/hardcoded && make build-local-proprietary-exp-plugin-image
	cd ../engines/router && make build-local-router-image
```

krithika369: Can we also add the following to the target:
* Setting up local infra
* Starting of the API server (this can be a background process)
* Running the tests

It would be nice to include a clean-up-local-e2e target as well, to remove any running processes (such as the API server) and to do docker-compose down on the infra.

api/README.md
```diff
@@ -91,73 +92,145 @@ watch KUBECONFIG=/tmp/kubeconfig kubectl get pod -A
...
+
+```bash
+pushd ../engines/experiment/examples/plugins/hardcoded
+{
+    make build-local-proprietary-exp-plugin-image
+}
+popd
 ```
 
+If you're facing issues building the proprietary experiment engine docker image locally, do the following steps.
```
krithika369, line 144: > If you're facing issues building the proprietary experiment engine docker image locally

Is this still possible? Can we simply build using the right GOOS and GOARCH flags? We only need this docker building step for the plugin to be available to the turing-router, right? Then I suppose the default GO env values will work?

krithika369 (approved):
Left some comments on the new make targets. The rest LGTM. Feel free to merge when you've addressed/reviewed them. Thanks! 🎉

.github/workflows/turing.yaml
leonlnj: oh weird i didnt get any notification, just happen to stumble on this. 

Im not too sure on the urgency of other tasks, let me check with Aria

api/README.md
terryyylim: Simplified the local E2E workflow.

leonlnj:
so moving forward will we need to maintain the docker-compose and make sure its up to date with new changes?

.github/workflows/turing.yaml
pradithya: I didn't realize this change.
Can you merge it to `main`? I'll try to adapt upi e2e test afterwards but timebox it to only today. I'll let you know if I need your help.

terryyylim:
> so moving forward will we need to maintain the docker-compose and make sure its up to date with new changes?

Yes.

.github/workflows/turing.yaml
terryyylim: Sure thing @pradithya ! Happy to chime in!

PR #270, opened 2022-10-23
Grpc upi UI

leonlnj:
### Description
This PR adds the UI for create, edit and router view workflow to support UPI protocol routers.

XP remote engine updated to show prediction context instead of headers - [PR](https://github.com/caraml-dev/xp/pull/49)

Key changes
- `ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js`
General Setting panel will support protocol selection (UPI or HTTP).
- `ui/src/router/components/form/components/router_config/RoutesPanel.js`
Routecard with grpc url validation
-  `ui/src/router/components/form/components/router_config/RulesPanel.js`
Traffic rules to allow header + prediction context
- Conditional in enricher/ensembler/experiment engine steps
  - Enricher not supported for UPI
  - Experiment engine supported
  - Ensembler standard type only supported for UPI 

### UI
https://user-images.githubusercontent.com/30390872/198812731-65e0bfb7-9a13-4439-827d-eb5716687cab.mov

https://user-images.githubusercontent.com/30390872/198812742-beedd7fa-0cca-4310-b750-59443ee33e46.mov

Detail view
Screenshot 2022-10-25 at 2 06 04 PM

### Testing
Screenshot 2022-10-25 at 1 35 42 PM
Screenshot 2022-10-25 at 1 35 54 PM


ui/src/router/components/form/validation/schema.js
```diff
@@ -92,7 +92,9 @@ const routerNameRegex = /^[a-z0-9-]*$/,
   kafkaBrokersRegex =
     /^([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+)(,([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+))*$/i,
   kafkaTopicRegex = /^[A-Za-z0-9_.-]{1,249}/i,
-  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
+  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/,
+  upiUrlRegex= /^([a-zA-Z0-9_-]{1,256})(\.[a-zA-Z0-9-]{1,253})*(:\d+)$/,
```
leonlnj: image

ui/src/components/form/endpoint_combo_box/SelectEndpointComboBox.js
```diff
@@ -4,12 +4,13 @@ import { EuiComboBoxSelect } from "../combo_box/EuiComboBoxSelect";
...
 export const SelectEndpointComboBox = ({
   value,
+  protocol,
   onChange,
   options,
   ...props
 }) => {
   const onCreateOption = (value) => {
-    if (!isValidUrl(value)) {
+    if (!isValidUrl(value, protocol)) {
```
leonlnj, line 13: actually i dont grasp the rationale for this behavior. If url is not valid, the router object values is never set. So if it is non empty, means it validated.

I think we can remove this and change and let the validation be done in `schema.js` to prevent two validation logic. 

Do lmk if i missed anything to why this is needed here
```
onCreateOption={onCreateOption}
onCreateOption={e => onChange(e)}
```

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,68 +1,101 @@
...
+      ),
+    },
+    {
+      value: "standard",
+      inputDisplay: "Standard",
+      description: (
+        <Fragment>
+          Turing will select the response from one of the routes, based on the
+          configured mapping between routes and experiment treatments. Only standard 
+          ensembler is support for UPI now.
```
pradithya, line 77: is *supported* for

ui/src/router/components/form/components/experiment_config/EngineTypePanel.js
```diff
@@ -4,8 +4,10 @@ import { EuiForm, EuiFormRow, EuiSuperSelect } from "@elastic/eui";
 import { DescribedFormGroup } from "../../../../../components/form/described_form_group/DescribedFormGroup";
 import { FormLabelWithToolTip } from "../../../../../components/form/label_with_tooltip/FormLabelWithToolTip";
 
-export const EngineTypePanel = ({ type, options, onChange, errors }) => {
+export const EngineTypePanel = ({ type, options, protocol, onChange, errors }) => {
   const selectedOption = options.find((option) => option.value === type);
+  const helpText = protocol === "UPI_V1" ? 
+    "For UPI Router, payload variables should be in Prediction Context" : ""
```
pradithya, line 10: "For UPI Router, payload variables should be stored as variable in Prediction Context field of the request payload."

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
```diff
@@ -80,6 +88,38 @@ export const RouteCard = ({
...
+            <EuiFormRow
+              label="Service Method *"
+              isInvalid={!!get(errors, "service_method")}
+              error={get(errors, "service_method")}
+              fullWidth>
+              <EuiComboBoxSelect
+                fullWidth
+                placeholder="/caraml.upi.v1.UniversalPredictionService/PredictValues"
+                value={route.service_method}
+                options={upiServiceMethodOptions}
```
pradithya, line 103: Is it customizable by user?

ui/src/router/components/form/validation/schema.js
```diff
@@ -103,9 +105,22 @@ const routeSchema = yup.object().shape({
   type: yup.string().oneOf(["PROXY"], "Route Type is required"),
   endpoint: yup
     .string()
-    .required("Valid url is required")
-    .url("Valid url is required"),
+    .when('$protocol', {
+      is: "UPI_V1",
+      then: (schema) => schema
+      .required("Valid grpc endpoint is required eg. {url}:{port}")
```
pradithya, line 111: It should be `{host}:{port}`, `url` implies `scheme` should be part of the input.

pradithya (commented):
I left few minor comment, but overall LGTM

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
leonlnj: it is, should i disallow that?

pradithya: No, I think it's fine this way

ui/src/router/components/form/validation/schema.js
```diff
@@ -92,7 +92,9 @@ const routerNameRegex = /^[a-z0-9-]*$/,
   kafkaBrokersRegex =
     /^([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+)(,([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+))*$/i,
   kafkaTopicRegex = /^[A-Za-z0-9_.-]{1,249}/i,
-  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
+  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/,
+  upiUrlRegex= /^([a-zA-Z0-9_-]{1,256})(\.[a-zA-Z0-9-]{1,253})*(:\d+)$/,
+  serviceMethodRegex= /^\/?(\w+(\.|-)?)+(\/[A-Za-z]+)$/
```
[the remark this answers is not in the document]
pradithya: Why the last one didn't match? Is numeric not allowed in the method name?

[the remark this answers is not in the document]
leonlnj: yup numeric isn't allowed in method name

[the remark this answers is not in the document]
pradithya: I just found out that grpc allows method name containing numeric value.

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,68 +1,101 @@
...
+    }
+  ]
+};
 
-export const ensemblerTypeOptions = (engineProps) => {
+export const ensemblerTypeOptions = (engineProps, protocol) => {
+  const options = typeOptions[protocol]
+  // For UPI, only standard ensembler is available and must be selected when there is experiment engine
+  if (protocol === "UPI_V1") {
+    if (engineProps.name) {
```
krithika369, line 88: `if (engineProps.name)` -> `if (!!engineProps.name)` is more apt.

In any case, why are we doing a special check with the protocol here? Doesn't `const options = typeOptions[protocol]` already ensure we have the right subset? The check that comes after should be enough, right?

```javascript
if (!engineProps.name) {
    // Standard Ensembler is not available when there is no experiment engine
    return options.filter((o) => o.value !== "standard");
}
```

ui/src/router/components/form/components/experiment_config/EngineTypePanel.js
```diff
@@ -4,8 +4,10 @@ import { EuiForm, EuiFormRow, EuiSuperSelect } from "@elastic/eui";
 import { DescribedFormGroup } from "../../../../../components/form/described_form_group/DescribedFormGroup";
 import { FormLabelWithToolTip } from "../../../../../components/form/label_with_tooltip/FormLabelWithToolTip";
 
-export const EngineTypePanel = ({ type, options, onChange, errors }) => {
+export const EngineTypePanel = ({ type, options, protocol, onChange, errors }) => {
   const selectedOption = options.find((option) => option.value === type);
+  const helpText = protocol === "UPI_V1" ?
```
krithika369, line 9: I feel like this help text here can be moved to the field source dropdown tooltip - when UPI is chosen as the protocol, "payload" should not even be visible as an option both for traffic rules and the experiment engine variables config, right? For the custom experiment engines, this actually comes from the experiment engine's microfrontend ([ref](https://github.com/caraml-dev/turing/blob/main/engines/experiment/docs/developer_guide.md#ui-1)) which I believe is currently missing now and should be added to the XP repo, for Turing Experiments. Is this TBD?

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
```diff
@@ -60,6 +63,31 @@ export const GeneralSettingsPanel = ({
...
+          display="row">
+          <EuiSuperSelect
+            fullWidth
+            options={protocolTypeOptions}
+            valueOfSelected={protocol}
+            onChange={(e)=>{
+              onChange("config.protocol")(e)
+              // reset routes and rules when protocol changes
+              onChange("config.routes")([newRoute()])
+              onChange("config.rules")([])
```
krithika369, line 83: This seems a bit harsh and the users will lose all their settings (even if only using headers). Should we try to swap "prediction_context" for "payload" and vice versa when the protocol changes but keep everything else in the traffic rules and experiment variables?

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
```diff
@@ -69,6 +71,12 @@ export const RuleCard = ({
 
   // Hide last row when all routes are selected
   const routeSelectionOptions = (rule.routes.length < routes.length ? [...rule.routes,  "_none_"]: rule.routes);
+  
+  const conditionLabel = protocol === "UPI_V1" ? 
+  <FormLabelWithToolTip
+    label="Conditions *"
+    content='Prediction Context of the UPI proto.  e.g. "model_name" in "model_a"'
```
krithika369, line 78: Why is this here? The condition can also be for the header, right?

Screenshot 2022-10-27 at 9 29 10 AM

IMO it's better for such descriptions to be a tooltip on the option itself, rather than in the parent UI component - it keeps the parent component generic.

ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
```diff
@@ -58,6 +58,20 @@ export const RouterDetailsPageHeader = ({ router }) => {
         },
       },
     },
+    {
+      title: "Protocol",
+      description: (
+        <EuiTextColor size="s" color={isRouterUpdating ? "subdued" : "default"}>
```
krithika369, line 64: Why is this disabled when the router is updating? Like the endpoint, we can show the protocol as well, right?

ui/src/router/components/request_field_source/FieldSourceFormLabel.js
```diff
@@ -9,23 +9,36 @@ import {
...
-  {
-    value: "header",
-    inputDisplay: "Header",
-  },
-  {
-    value: "payload",
-    inputDisplay: "Payload",
-  },
-];
+const fieldSourceOptions = {
```
krithika369, line 12: Screenshot 2022-10-27 at 9 37 58 AM

I encountered this problem when attempting to use an internal standard experiment engine with UPI.

ui/src/router/components/form/validation/schema.js
```diff
@@ -92,7 +92,9 @@ const routerNameRegex = /^[a-z0-9-]*$/,
   kafkaBrokersRegex =
     /^([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+)(,([a-z]+:\/\/)?\[?([0-9a-zA-Z\-%._:]*)\]?:([0-9]+))*$/i,
   kafkaTopicRegex = /^[A-Za-z0-9_.-]{1,249}/i,
-  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
+  trafficRuleNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/,
```
krithika369, line 95: The form validation seems to have a regression - it's no longer scrolling back to the top of the section which has the errors:

https://user-images.githubusercontent.com/23465343/198171085-729198bf-b47a-4640-9305-07bdc674a256.mov

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
```diff
@@ -13,16 +13,23 @@ import { SelectEndpointComboBox } from "../../../../../../components/form/endpoi
...
 
 export const RouteCard = ({
   route,
+  protocol,
   endpointOptions,
   onChange,
   onDelete,
   errors,
 }) => {
   const [endpointsMap, setEndpointsMap] = useState({});
```
krithika369, line 26: Eventually, I imagine we will filter the available endpoints by the chosen protocol either in `ui/src/providers/endpoints/MerlinEndpointsProvider.js` or here, right?

krithika369 (commented):
Thanks for the PR and the details description. Left some questions / comments.

ui/src/components/form/endpoint_combo_box/SelectEndpointComboBox.js
leonlnj: >  checking free-form text upfront in cases like these can be useful.
fair enough, will keep this then. I didn't see other free-form input component having such validation so I thought this was off, but your right to say it could a tiny enhancement to others rather than removing this. Thanks!

ui/src/components/validation.js
```diff
@@ -10,10 +10,16 @@ const jsonPathSchema = yup
     'Valid Request Payload json path is required, e.g. "my_object.session_id"'
   );
 
+const predictionContextSchema = yup
+.string()
+.required(
+  'Valid Prediction Context value is required, e.g. "model_name" in "model_a"'
```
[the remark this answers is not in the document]
leonlnj: oh right. I could just leave it as ` e.g. "model_name"` then. 

I thought it was confusing for myself "X" in "Y", which is the key and value, like `Header x-session-id in 123` so I added the "in" unknowingly to remind myself  😅

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
leonlnj: ah this is a mistake, initially i forgotten that we support `Header` as well and so I written with assumption `Prediction Context` was the only option. 

`tooltip on the option itself` -> right, this sounds better. will change it

ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
leonlnj: Oh, this is because the protocol is part of the `config` object which is `undefined` together with `timeout` during updating state

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
leonlnj: For UPI yes, actually they should all be `"/caraml.upi.v1.UniversalPredictionService/PredictValues"`, only if there is custom GRPC then a free form would be allowed. 

should we fix the route service method and disable user modification? @pradithya do you foresee any issue?

pradithya: Yes, I am fine with that

ui/src/router/components/form/components/ensembler_config/typeOptions.js
leonlnj: right, i can remove it totally. I added this before I added the map and missed updating this 🙏

ui/src/router/components/form/components/router_config/rule_card/RuleCard.js
leonlnj: Not sure if it helps but i shifted it 😅 
image

ui/src/router/components/form/components/experiment_config/EngineTypePanel.js
leonlnj: > "payload" should not even be visible as an option both for traffic rules and the experiment engine variables config, right? 

Yes, it is as such for traffic rules now

> For the custom experiment engines, this actually comes from the experiment engine's microfrontend ([ref](https://github.com/caraml-dev/turing/blob/main/engines/experiment/docs/developer_guide.md#ui-1)) which I believe is currently missing now and should be added to the XP repo, for Turing Experiments. Is this TBD?

No, it was intentional as I thought the experiment engine doesn't need to know about the existence of UPI as the interaction with experiment engine required no changes. If shifted, all experiment engine would have to modify the remote ui to display on HTTP. Also, will this will lead to requiring experiment engine to be explicit about supporting UPI ? They should be able to opt to not support this if they don't use UPI.

This said I do agree there is some inconsistency with the traffic rules. There so much work involved just to add a tooltip below 😰  I will address this in a separate PR if that is fine.

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
```diff
@@ -13,16 +13,23 @@ import { SelectEndpointComboBox } from "../../../../../../components/form/endpoi
...
 
+  const upiServiceMethodOptions = [{
+    label: "/caraml.upi.v1.UniversalPredictionService/PredictValues",
```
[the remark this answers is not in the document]
leonlnj: I did this with thought that it would support custom grpc too, but I probably shouldn't had done it for consistency.

Checked with aria, for UPI we will fix the service method and remove it totally from the card since its not configurable.

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
leonlnj: I thought of it, with some relevance to the earlier convo on the url validation

For the routes, the component values is empty, if it is set it must had passed the url validation regex. 
Just changing the protocol value would had left invalid http endpoint for the UPI router, which is a new state of invalid value set (although it can be caught by yup validation), which lead me to reset it completely.

> Should we try to swap "prediction_context" for "payload" and vice versa when the protocol changes but keep everything else in the traffic rules and experiment variables
Good suggestion, will update accordingly

ui/src/router/components/form/validation/schema.js
leonlnj: I see the same behaviour in dev, not sure what triggers the scrolling back, will check it.

ui/src/router/components/form/validation/schema.js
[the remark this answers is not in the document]
leonlnj: added numeric to regex

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
leonlnj: updated

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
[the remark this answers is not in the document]
leonlnj: removed service_method section and internally modify the obj

ui/src/router/components/form/validation/schema.js
leonlnj: If i understood it correctly, for [StepWizardHorizontal](https://github.com/gojek/mlp/blob/main/ui/packages/lib/src/components/multi_steps_form/StepsWizardHorizontal.js), there is no scrolling unlike the [AccordionForm](https://github.com/gojek/mlp/blob/main/ui/packages/lib/src/components/accordion_form/AccordionForm.js)  which has the [AccordionFormScrollController](https://github.com/gojek/mlp/blob/main/ui/packages/lib/src/components/accordion_form/AccordionFormScrollController.js) which does the scrolling to section `onSubmit`. 

Just that it usually isn't that common for a Step to be that long to notice it all along i guess.

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
krithika369: Actually my question is more about the HTTP vs UPI Merlin endpoints - we currently get all endpoints in the `MerlinEndpointsProvider`, that are active under the same project. We should filter only those endpoints that match the protocol.

> should we fix the route service method and disable user modification?

This sounds good to me too.

ui/src/router/components/form/validation/schema.js
krithika369: I didn't realise that. TIL. Future improvement to the `StepWizardHorizontal`, I suppose. 😅  Thanks for digging into this!

ui/src/router/details/components/router_details_header/RouterDetailsPageHeader.js
krithika369: Ahh ok. The reason they are not set while the router is updating is because they can change between versions. By this logic, may be we should also hide the router endpoint. The downside to this this, there is no other place that can show the endpoint value (eg: router version details) as it's a property of the router. So I suppose we can leave this as it is for now.

ui/src/router/components/request_field_source/FieldSourceFormLabel.js
leonlnj: added a filter in experiment step `ui/src/router/components/form/steps/ExperimentStep.js` as UPI doesn't intend to support standard experiment engine yet.

ui/src/router/components/form/components/experiment_config/EngineTypePanel.js
leonlnj: removed and opened [PR](https://github.com/caraml-dev/xp/pull/49) in XP

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
leonlnj: updated the videos, around 40s the protocol is changed

ui/src/router/components/form/components/router_config/route_card/RouteCard.js
leonlnj: right, added the filter to `ui/src/router/components/form/components/router_config/RoutesPanel.js`. Assuming all grpc endpoint will not start with `http` while http endpoints will always have `http`.

leonlnj:
Updated docs with relevant screenshot changes

docs/how-to/create-a-router/configure-routes.md
```diff
@@ -6,6 +6,8 @@ The request sent to the each route is constructed from the request headers to th
 identifier `Turing-Req-Id` that is uniquely assigned to each request received by the Router), and either the request
 body to the Router (if there is no Enricher) or the response body from the Enricher (if enabled).
 
+For UPI Routers, endpoint must be support [UPI interface](https://github.com/caraml-dev/universal-prediction-interface).
```
terryyylim, line 9: nit: `the endpoint must support [UPI interface]...`

docs/how-to/create-a-router/configure-traffic-rules.md
```diff
@@ -51,10 +51,12 @@ From the above rules, both rule 1 and 2 are not overlapping because there's at l
...
+Each rule should have at least one condition configured on it. If there are multiple conditions configured on the same rule, then this rule will be triggered only if each and every condition is satisfied. 
 
-* **Condition source**: either `Header` or `Payload`
-* **Condition key**: if condition's source is `Header` – then the name of a request Header (example: `X-Session-ID`), or else, if condition's source is `Payload` – a valid JSON path of the property from the request's JSON payload (example: `service_type.id` or `users.0.name`)
+Rule condition can be defined on either request header or request payload (assuming payload is a valid JSON object). For UPI routers, it would be header or [prediction context](https://github.com/caraml-dev/universal-prediction-interface/blob/main/proto/caraml/upi/v1/upi.proto) (name of the variable and value). For each condition you should specify:
```
terryyylim, line 56: nit: missing comma, `For each condition, you should specify:`

ui/src/router/components/form/components/enricher_config/typeOptions.js
```diff
@@ -1,21 +1,31 @@
...
-    disabled: true,
-  },
-];
+export const enricherTypeOptions = {
+  HTTP_JSON: [
+    {
+      value: "nop",
+      inputDisplay: "None",
+      description:
+        "No enrichment gonna happen. Original request will be sent to configured routes",
```
terryyylim, line 7: nit: `gonna` sounds really informal. `There will be no enrichment being done. The original request will be sent to the configured routes.`

ui/src/router/components/form/components/enricher_config/typeOptions.js
```diff
@@ -1,21 +1,31 @@
...
+      value: "docker",
+      inputDisplay: "Docker",
+      description:
+        "Turing will deploy specified image as a pre-processor and will send original request to it for enrichment",
+    },
+    {
+      value: "external",
+      inputDisplay: "External (Coming Soon)",
+      description:
+        "Turing will send original request to the external URL for enrichment",
```
terryyylim, line 19: nit: `send the original request`

ui/src/router/components/form/components/enricher_config/typeOptions.js
```diff
@@ -1,21 +1,31 @@
...
+      disabled: true,
+    },
+  ],
+  UPI_V1: [
+    {
+      value: "nop",
+      inputDisplay: "None",
+      description: "Enricher is not yet supported for the UPI protocol.",
```
terryyylim, line 27: nit: Can we standardise full-stop for all the above options' description?

ui/src/router/components/form/components/enricher_config/typeOptions.js
```diff
@@ -1,21 +1,31 @@
...
+      value: "nop",
+      inputDisplay: "None",
+      description:
+        "No enrichment gonna happen. Original request will be sent to configured routes",
+    },
+    {
+      value: "docker",
+      inputDisplay: "Docker",
+      description:
+        "Turing will deploy specified image as a pre-processor and will send original request to it for enrichment",
```
terryyylim, line 13: nit: `deploy the specified image as a pre-processor and send the original request to it...`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,69 +1,99 @@
...
+          configured mapping between routes and experiment treatments
+        </Fragment>
+      ),
+    },
+    {
+      value: "docker",
+      inputDisplay: "Docker",
+      description: (
+        <Fragment>
+          Turing will deploy specified image as a post-processor and will send
```
terryyylim, line 30: nit: `will deploy the specified image as a post-processor and send it responses from all routes..., for ensembling the final response.`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,69 +1,99 @@
...
+      ),
+    },
+    {
+      value: "pyfunc",
+      inputDisplay: "Pyfunc",
+      description: (
+        <Fragment>
+          Turing will build and deploy the selected pyfunc ensembler and will
+          send to it responses from all routes, together with the treatment
+          configuration, for ensembling
```
terryyylim, line 43: nit: `and send the responses from all routes..., for ensembling the final response.`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,69 +1,99 @@
...
+        </Fragment>
+      ),
+    },
+    {
+      value: "external",
+      inputDisplay: "External (Coming Soon)",
+      description: (
+        <Fragment>
+          Turing will send responses from all routes, together with treatment
+          configuration, to the external URL for ensembling
```
terryyylim, line 53: nit: `Turing will send the responses from all routes, along with the treatment configuration to the external URL, for ensembling the final response.`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,69 +1,99 @@
...
+      disabled: true,
+    },
+  ],
+  UPI_V1: [
+    {
+      value: "nop",
+      inputDisplay: "None",
+      description: (
+        <Fragment>
+          Turing will simply return response from the route marked as the{" "}
```
terryyylim, line 65: nit: `will simply return the response`

ui/src/router/components/form/components/ensembler_config/typeOptions.js
```diff
@@ -1,69 +1,99 @@
...
+          <strong>Final Response</strong>. Other ensembler types are not yet
+          support for the UPI protocol.
```
terryyylim, line 67: nit: `not yet supported`

ui/src/router/components/form/validation/schema.js
```diff
@@ -101,16 +106,32 @@ const timeoutSchema = yup
...
   type: yup.string().oneOf(["PROXY"], "Route Type is required"),
-  endpoint: yup
-    .string()
-    .required("Valid url is required")
-    .url("Valid url is required"),
+  endpoint: yup.string().when("$protocol", {
+    is: "UPI_V1",
+    then: (schema) =>
+      schema
+        .required("Valid grpc endpoint is required eg. {host}:{port}")
```
terryyylim, line 113: nit: `gRPC`

ui/src/router/components/form/validation/schema.js
```diff
@@ -101,16 +106,32 @@ const timeoutSchema = yup
...
+        .matches(
+          upiUrlRegex,
+          "Valid grpc endpoint is required eg. {host}:{port}"
```
terryyylim, line 116: nit: `gRPC`

terryyylim (approved):
LGTM, bunch of nits, thanks @leonlnj !

ui/src/router/components/form/components/experiment_config/ExperimentConfigPanel.js
```diff
@@ -39,6 +41,7 @@ const CustomExperimentEngineConfigGroup = ({
           fallback={<FallbackView text="Loading Experiment Engine config" />}
           projectId={projectId}
           config={config}
+          protocol={protocol}
```
krithika369, line 44: Could you help update the doc for the experiment engine developer guide? Particularly, the UI contract for the custom experiment engine - adding the protocol argument to the `EditExperimentEngineConfig` and `ExperimentEngineConfigDetails` components: https://github.com/caraml-dev/turing/blob/main/engines/experiment/docs/developer_guide.md#ui-1

docs/how-to/create-a-router/configure-traffic-rules.md
```diff
@@ -51,10 +51,12 @@ From the above rules, both rule 1 and 2 are not overlapping because there's at l
...
 
-* **Condition source**: either `Header` or `Payload`
-* **Condition key**: if condition's source is `Header` – then the name of a request Header (example: `X-Session-ID`), or else, if condition's source is `Payload` – a valid JSON path of the property from the request's JSON payload (example: `service_type.id` or `users.0.name`)
+Rule condition can be defined on either request header or request payload (assuming payload is a valid JSON object). For UPI routers, it would be header or [prediction context](https://github.com/caraml-dev/universal-prediction-interface/blob/main/proto/caraml/upi/v1/upi.proto) (name of the variable and value). For each condition, you should specify:
+
+* **Condition source**: either `Header` or `Payload` for HTTP router. `Header` or `Prediction Context` for UPI router.
+* **Condition key**: if condition's source is `Header` – then the name of a request Header (example: `X-Session-ID`), or else, if condition's source is `Payload` – a valid JSON path of the property from the request's JSON payload (example: `service_type.id` or `users.0.name`). If condition source is `Prediction Context` - the name of the variable within Prediction Context proto.
```
krithika369, line 59: Nit: `within Prediction Context proto` -> ` within the Prediction Context proto`

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
```diff
@@ -60,6 +70,59 @@ export const GeneralSettingsPanel = ({
...
+            />
+          }
+          display="row"
+        >
+          <EuiSuperSelect
+            fullWidth
+            options={protocolTypeOptions}
+            valueOfSelected={protocol}
+            onChange={(e) => {
+              onChange("config.protocol")(e);
```
krithika369, line 88: Nit: Can we define this handler separately so it can just be used here like `onChange={onChangeProtocol}` ? It more readable that way.

docs/how-to/create-a-router/configure-routes.md
```diff
@@ -6,6 +6,8 @@ The request sent to the each route is constructed from the request headers to th
 identifier `Turing-Req-Id` that is uniquely assigned to each request received by the Router), and either the request
 body to the Router (if there is no Enricher) or the response body from the Enricher (if enabled).
 
+For UPI Routers, the endpoint must support [UPI interface](https://github.com/caraml-dev/universal-prediction-interface).
```
krithika369, line 9: Should we mention something about the default service method here?

ui/src/router/components/form/steps/ExperimentStep.js
```diff
@@ -40,13 +41,18 @@ export const ExperimentStep = ({ projectId }) => {
     ExperimentEngineContext
   );
   const engineProps = getEngineProperties(experiment_engine.type);
+  // UPI will only support custom experiment engine now
+  const filteredEngine =
```
krithika369, line 45: Nit: `filteredEngine` -> `filteredEngines` in plural

ui/src/router/components/form/validation/schema.js
```diff
@@ -72,7 +74,8 @@ const validateDanglingRoutes = function (items) {
       errors.push(
         this.createError({
           path: `${this.path}[${idx}].id`,
-          message: "This route should be removed since they have no Traffic Rule(s) associated and will never be called.",
+          message:
+            "This route should be removed since they have no Traffic Rule(s) associated and will never be called.",
```
krithika369, line 78: Not a change introduced in this PR but I think it should be `since it has no` instead of `since they have no`.

ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
```diff
@@ -60,6 +70,59 @@ export const GeneralSettingsPanel = ({
...
+              // update service_method for each route if changed to UPI
+              onChange("config.routes")(
+                routes.map((route) => {
+                  return {
+                    ...route,
+                    service_method:
```
krithika369, line 94: I guess we are not showing the service method anywhere. Though it's auto-populated, I think it would be better to show it (as a non-editable field or by [appending](https://elastic.github.io/eui/#/forms/form-controls#prepend-and-append) to the endpoint, to make it clearer to the user. wdyt?

krithika369 (commented):
@leonlnj, a little late but I took another look at this PR have a few small comments. Could you review them and address the relevant ones in a new PR? Thanks! The rest looks great.

PR #273, opened 2022-10-26
Fix and rename make commands for local E2E

terryyylim:
Fix incorrect `make build-local-router-image` where `IMAGE_TAG` is empty when executing the following:
```
docker push $(IMAGE_TAG)
```

Also, better naming of Make command.

### DIFF
--- a/engines/router/Makefile
+++ b/engines/router/Makefile
@@ -97,9 +97,8 @@ swagger-ui:
 
 
 .PHONY: build-local-router-image
-build-local-router-image: vendor
-	$(MAKE) build-image DOCKER_REGISTRY=localhost:5000 OVERWRITE_VERSION=latest
-	docker push $(IMAGE_TAG)
+build-local-router-image: vendor build-image
+	docker push ${DOCKER_REGISTRY}/turing-router:${OVERWRITE_VERSION}
 
 
 #   R E V I E W   R E Q U I R E D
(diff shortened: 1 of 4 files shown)


krithika369 (approved):
LGTM!

engines/router/Makefile
```diff
@@ -97,9 +97,8 @@ swagger-ui:
 
 
 .PHONY: build-local-router-image
-build-local-router-image: vendor
-	$(MAKE) build-image DOCKER_REGISTRY=localhost:5000 OVERWRITE_VERSION=latest
-	docker push $(IMAGE_TAG)
+build-local-router-image: vendor build-image
+	docker push ${DOCKER_REGISTRY}/turing-router:${OVERWRITE_VERSION}
```
leonlnj, line 101: i think this line can remain as `docker push $(IMAGE_TAG)` since the is using `image_tag`

leonlnj (approved):

pradithya:
Another issue that I faced. When running `make proprietary-exp-plugin`, the operation `ln -sf ${PWD}/bin/example-plugin ../../../../../api/bin/example-plugin` failed since `../../../../../api` is not available. Adding directory creation part should solve it. 

```
proprietary-exp-plugin: tidy
	go build -o ./bin/example-plugin ./cmd/main.go
	mkdir -p ../../../../../api/bin
	ln -sf ${PWD}/bin/example-plugin ../../../../../api/bin/example-plugin
```

PR #274, opened 2022-10-26
Deploy experiment engine secrets with router deployment

deadlycoconuts:
## Context
As part of an ongoing process to make the Treatment Service of [XP](https://github.com/caraml-dev/xp) runnable as a plugin for [Turing](https://github.com/caraml-dev/turing/) Router deployments, as opposed to calling a singular Treatment Service that has been deployed to receive treatments, this PR introduces a series of changes to pass secrets and configuration values required in the Treatment Service to the XP plugin. 

During each router deployment, the API will rely on the XP plugin manager to supply the configuration that needs to be deployed together with the XP plugin runner. In order to generate this configuration, the following steps need to be performed before returning the aforementioned experiment configuration to the Turing API (refer to the illustration below):

1. (XP plugin manager) Retrieves values shared with the Management Service via the HTTP GET endpoint at `\treatment-service-plugin-config`
2. (XP plugin manager) Use values stored within the XP plugin manager (these are passed to the plugin through Turing API's config file when the plugin manager gets initialised) and those retrieved in Step 1 to create the experiment configuration
3. (XP plugin manager) Return the created experiment configuration to the Turing API
4. (Turing API) Retrieve the secret(s) required 
5. (Turing API) Inject the required configuration values and secrets into the Turing Router deployment

xpplugin

As the responsibility of Turing lies in performing **Steps 4 and 5**, this PR specifically introduces changes to support those steps, while a separate [PR](https://github.com/caraml-dev/xp/pull/47) in XP has been created to perform "Steps 1, 2 and 3**.

### Main Modifications
- `api/turing/api/deployment_controller.go` - Modification of the deployment controller to retrieve the secret required for the treatment service plugin when deploying a router version
- `api/turing/cluster/controller.go` - Modification of the cluster controller to create a controller to the cluster in which the Turing API will be deployed, as well as methods that are used to retrieve the secret from this cluster
- `api/turing/cluster/servicebuilder/service_builder.go` - Addition of the experiment engine service account key to the secret deployment that happens at each router version deployment
- `api/turing/config/config.go` - Modification of the `ExperimentEnginePluginConfig` struct to contain a new `ServiceAccount` field
- `infra/charts/turing/templates/_helpers.tpl` - Addition of a new field in the helm chart templates to inject the service account name value into the configs of the Turing API deployment


api/turing/api/deployment_controller.go
```diff
@@ -453,3 +434,45 @@ func (c RouterDeploymentController) updateRouterVersionStatusToFailed(
 	}
 	return errors.New(strings.Join(errorsStrings, ". "))
 }
+
+func (c RouterDeploymentController) getExperimentConfig(routerVersion *models.RouterVersion) (json.RawMessage, error) {
```
deadlycoconuts, line 438: Extracted all the logic into a separate method.

api/turing/cluster/controller.go
```diff
@@ -198,6 +201,17 @@ func InitClusterControllers(
...
+	selfCtl, err := newController(
+		clusterConfig{
+			ClusterName:     LocalClusterName,
+			InClusterConfig: true,
+		},
+	)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to initialize cluster controller")
+	}
+	controllers[LocalClusterName] = selfCtl
```
deadlycoconuts, line 213: This new controller allows the Turing API to work directly with the cluster in which it is deployed.

krithika369: What do you think about letting the experiment engine declare the service account key file's path relative to the API, rather than the name of the secret in the API's cluster? We can use the `extraVolumes` and `extraVolumeMounts` fields in the chart to mount the engine's SA data: https://github.com/caraml-dev/turing/blob/main/infra/charts/turing/values.yaml#L85

I think it can help with 2 things:
* Makes the implementation simpler (no need to have the local controller, get the secret from the Turing app's cluster)
* Makes local development consistent - this is something I hadn't considered earlier but with the current implementation, if we were to run the API locally and point to a remote cluster for deploying the routers, this step of "getting a secret from the app's cluster" doesn't seem to fit in. :/ 

Wdyt ?

api/turing/api/deployment_controller.go
```diff
@@ -190,41 +191,20 @@ func (c RouterDeploymentController) deployRouterVersion(
...
-		isClientSelectionEnabled, err := expSvc.IsClientSelectionEnabled(routerVersion.ExperimentEngine.Type)
+		experimentConfig, err = c.getExperimentConfig(routerVersion)
 		if err != nil {
 			return "", c.updateRouterVersionStatusToFailed(err, routerVersion)
 		}
-		if isClientSelectionEnabled {
-			// Convert the config to the standard type
-			standardExperimentConfig, err := manager.ParseStandardExperimentConfig(experimentConfig)
+
+		if routerVersion.ExperimentEngine.PluginConfig.ServiceAccount != nil {
```
krithika369, line 200: I think we should also check for `PluginConfig` != nil ?

On a related note, I wonder if the `ServiceAccount` should be promoted up (directly under [ExperimentEngine](https://github.com/caraml-dev/turing/blob/main/api/turing/models/experiment_engine.go#L17)), as it's not limited to RPC plugins (theoretically it can also be used for the in_proc experiment engines, right?). Along the same lines, when we are taking the configs in for the experiment engines, we could take it in as a separate config, at the same level as `ExperimentEnginePluginConfig`.

 So, the helm values would look like: https://github.com/caraml-dev/turing/blob/main/infra/charts/turing/values.yaml#L107

(Example considers this comment as well: https://github.com/caraml-dev/turing/pull/274/files#r1013555923)

```
Example:
  - name: my-exp-engine
    type: rpc-plugin
    serviceAccountKeyFilePath: /etc/abc.json
    rpcPlugin:
      image: ghcr.io/turing/my-exp-engine:latest
    options:
      key-1: value-1
```

And may be, in `config.go`, we need to change to:
```
ExperimentEnginePlugins map[string]*ExperimentEngineConfig `validate:"dive"`
```

where `ExperimentEngineConfig` is:
```
type ExperimentEngineConfig {
    PluginConfig *ExperimentEnginePluginConfig
    ServiceAccountKeyFilePath *string
}
```

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -287,11 +289,13 @@ func (sb *clusterSvcBuilder) NewSecret(
 	routerServiceAccountKey string,
 	enricherServiceAccountKey string,
 	ensemblerServiceAccountKey string,
+	expEngineServiceAccountKey string,
 ) *cluster.Secret {
 	data := map[string]string{
 		secretKeyNameRouter:    routerServiceAccountKey,
 		secretKeyNameEnricher:  enricherServiceAccountKey,
 		secretKeyNameEnsembler: ensemblerServiceAccountKey,
+		secretKeyNameExpEngine: expEngineServiceAccountKey,
```
krithika369, line 298: This is good. But we also need to mount this key to the router, right (like [this](https://github.com/caraml-dev/turing/blob/8002d13de90f81efbeb545c02c5504bd35e3c132/api/turing/cluster/servicebuilder/router.go#L328))? Will that be done in a different PR?

krithika369 (commented):
Thanks for the detailed PR description - that was really helpful to get an overview of the context and the changes involved.

I think this PR is very clean and I agree with the overall workflow. But after analysing it more closely, I have a comment for simplification and another one for consistency with the in-proc engines. Please review them and LMK what your thoughts are.

api/turing/cluster/servicebuilder/service_builder.go
deadlycoconuts: Oh that's true, I must've missed that out; I'll add this to this PR.

api/turing/cluster/controller.go
deadlycoconuts: Yep that makes sense! On the other hand, if we were to make users specify a `ServiceAccountKeyFilePath` under the experiment engine, the user would now have to potentially specify values in 2 different locations, 1 for the actual `ServiceAccountKeyFilePath` found under `experimentEngines`
```yaml
experimentEngines:
- name: xp
   type: rpc-plugin
   serviceAccountKeyFilePath: /etc/plugins/xp/gcp_service_account/service-account.json
   ...
```
and 1 other for determining which secret (the secret name) to be mounted as a volume in the Turing API pod
```yaml
extraVolumes:
- name: xp-plugin-service-account
  secret:
    secretName: xp-secret
    items:
    - key: service-account.json
       path: service-account.json
```
All other values/configurations can kind of be automatically generated using helm templates automatically. I was thinking if we can go further and make the file path of the service account unconfigurable (where it gets mounted will be determined by the helm template) and allow the users to directly state the name of the secret under the experiment engine. This is pretty much identical to how the original implementation of this PR would require users to specify the `experimentEngines` configuration:
```yaml
experimentEngines:
- name: xp
   type: rpc-plugin
   secretName: xp-secret
   ...
```
We'll then rely on the helm template to automatically generate the volumes and volume mounts and also set the `ServiceAccountKeyFilePath` field under `RouterDefaults.ExperimentEnginePlugins`. What do you think? 🤔

krithika369: The user also needs to set the `extraVolumes `'s `key: service-account.json` to identify what key to look for in the secret, right? Automatically appending to the `extraVolumes` and `extraVolumeMounts` via the Helm templates makes sense to me.

deadlycoconuts: Oh yea, that too. So I guess we'll have 2 fields, one for the `secretName` and `key`. But now as I'm thinking through this, I'm starting to realise that this automatic appending of the `extraVolumes` and `extraVolumeMounts` for the experiment engine plugin might be taking things a little too far: we aren't even doing this for the regular service account keys that the Turing API requires (one has to specify the `extraVolumes`, `extraVolumeMounts` and `extraEnvs` manually to ensure the creds file can be properly located). 

For consistency with how we're expecting users to mount the creds file themselves for the Turing API (do-it-yourself, unknown to the Helm charts, secret-agnostic in some sense since we don't care how the creds file get inserted into the pod as long as it's there), I was thinking maybe the best idea is to just keep the `ServiceAccountKeyFilePath` (sticking to your original suggestion) and assume that the user knows how to get the creds file for the plugin into the Turing API's pod somehow. 

Sorry for flip flopping on this 😵‍💫

api/turing/cluster/servicebuilder/router.go
```diff
@@ -322,6 +322,27 @@ func buildRouterVolumes(
...
+								Key:  secretKeyNameExpEngine,
+								Path: secretKeyNameExpEngine,
+							},
+						},
+					},
+				},
+			})
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      secretVolumeExpEngine,
+				MountPath: secretMountPath,
```
krithika369, line 343: I'm not sure if we can actually mount 2 volumes at the same path. See: https://blog.sebastian-daschner.com/entries/multiple-kubernetes-volumes-directory

Shall we append the secretKeyName to the path to avoid conflict, as in the linked article?

krithika369 (approved):
Left one comment - the rest LGTM! Thanks for the changes and the refactoring. Please feel free to merge once you've addressed that comment.

api/turing/cluster/servicebuilder/router.go
deadlycoconuts: Ohh yea I forgot about that; I'll be appending the `exp-engine` to the `MountPath` when we're setting up the experiment engine secret volume mount (the `MountPath` becomes `secretMountPath + "exp-engine/"` or `"/var/secret/exp-engine"`). The volume mount for the router's secret remains the same. 

I read a little more about using the `subPath` field but apparently doing so makes the secrets specified within unable to get updated during runtime (the only way for the pod to get its secrets updated is to restart it) 🥲https://www.golinuxcloud.com/k8-mount-multiple-secrets-to-same-directory/

krithika369: Thanks for sharing this, I wasn't aware about the limitation of using the `subPath`.
Although, I'm skeptical about the current implementation too where one mount path is nested under the other: https://github.com/kubernetes/kubernetes/issues/57421

Shall we simply separate them?

deadlycoconuts: 😱 Didn't expect this to have so many quirks; okay then, I'll make the new mount paths as follows:
- router secret: `"secretMountPath + "router/"`
- exp engine secret: `"secretMountPath + "exp-engine/"`

deadlycoconuts:
Thanks a lot @krithika369 for the review and for the detailed comments! 🙏 I've slipped in 1 or 2 more commits that were used to fix certain unexpected issues caused by the main changes (changes to the OpenAPI specs and also fixes to the ensembling job code because the linters were failing it).  I'll be merging this as soon as the pipeline passes! 🚀

PR #275, opened 2022-10-27
Add default pod scaling limits

deadlycoconuts:
## Context
In preparation to support automatic resource estimation for users of Turing Routers (and their associated components) through a new 'default autoscaling policy', as opposed to having them configure resource requests and autoscaling policies manually, this PR introduces API, SDK, and UI changes to support the selection and deployment of the said 'default autoscaling policy'.

Without going too deep into details of how this default autoscaling policy works, as this will be elaborated on in another PR, the new workflow for deploying this policy requires a user to initially estimate the size of payloads to be received by a certain component (Turing Router, enricher or ensembler). This payload size will then be used to estimate a certain `rps` (the process of which has yet to be determined and instead temporarily fulfilled by a dummy function), that gets configured as the KPA `rps` target by the Turing API. A VPA will then additionally be deployed (to be added in a future PR) alongside the existing KPA that we currently already do to perform resource recommendations over time.

## Workflow Changes
In order to store data of this user-estimated payload size value, the existing `models.AutoscalingPolicy` struct has been modified to feature a new nullable-field `PayloadSize`. When this field is non-`nil`, it implicitly means that the component for which this struct is part of is configured with the 'default autoscaling policy', and as such, the `Metric` and `Target` fields should accordingly be expected to be `nil`. These values would be stored as such in the DB though they would be configured with initial values during deployment.

```go
type AutoscalingPolicy struct {
	Metric *AutoscalingMetric `json:"metric"`
	// Target is the target value of the metric that should be reached to add a new replica.
	// It is expected that the autoscaling target is an absolute value for concurrency / rps
	// while it is a % value (of the requested value) for cpu / memory.
	Target *string `json:"target"`
	// PayloadSize indicates the request payload
	PayloadSize *resource.Quantity `json:"payload_size"`
}
```
Hence, the existing `Metric` and `Target` fields have been made nullable in this PR. Furthermore, when the default autoscaling policy is enabled (`PayloadSize` is non-`nil`), we should also expect the `ResourceRequest` field to be `nil` since a user would no longer have to configure the values in it anymore; a set of initial values would still be used during deployment but these would not be configurable by the user (see the section 'API Changes' below). This change has also been made in this PR.

## SDK Changes
The SDK has been updated to feature the new changes mentioned above.

### 🚨 Breaking Changes
- Due to some refactoring with the Save Router Version endpoint of the API server (see below for more info), users need to update their SDK versions after this change gets deployed.
- The `autoscaling_policy` field in enricher, ensembler and router configs are now a non-nullable argument at initialisation. No default autoscaling policy will be applied if that argument is not specified.

## UI Changes 
Router Create/Edit Page
ezgif com-gif-maker (2)

Router Config Page
Screenshot 2022-11-01 at 11 13 30 AM

## API Changes
The initial values for the min and max pod replicas when a user enables the default autoscaling policy are to now be retrieved from the Turing API config, under the `DefaultAutoScalingPolicyConfig` field:

```yaml
DeployConfig:
  EnvironmentType: dev
  DefaultAutoScalingPolicyConfig:
    Enabled: true
    DefaultMinReplica: 0
    DefaultMaxReplica: 30
```

### 🚨 Breaking Changes
Due to some refactoring with the Save Router Version endpoint of the API server (see below for more info), users that communicate directly with the Turing API need to update their request payloads after this change gets deployed.

## Minor Refactoring
The current Save Router Version endpoint expects a `RouterConfig` in the request payload while the Update Router endpoint expects this `RouterConfig` nested in another struct that contains the router name and the deployment environment name, which is inconsistent because both operations technically perform almost exactly the same thing, except for actually deploying the new version (or not). Furthermore the router name and the environment name are both redundant fields since the API does not actually allow those fields to be changed. Hence, this PR modifies the Update Router endpoint (as well as the SDK and the UI) to simply expect a `RouterConfig` in the payload.


api/api/openapi.bundle.yaml
```diff
@@ -619,7 +619,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/RouterConfig'
+              $ref: '#/components/schemas/RouterVersionConfig'
```
deadlycoconuts, line 622: Standardising the expected payload required for the edit router operation.

api/api/specs/routers.yaml
```diff
@@ -117,7 +117,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/RouterConfig"
+              $ref: "#/components/schemas/RouterVersionConfig"
```
deadlycoconuts, line 120: Standardising the expected payload required for the edit router endpoint.

api/api/specs/routers.yaml
```diff
@@ -737,7 +738,7 @@ components:
       type: "object"
       required:
         - image
-        - resource_request
+        - autoscaling_policy
```
deadlycoconuts, line 741: Making `resource_request` optional while `autoscaling_policy` becomes required, as detailed in the PR description.

api/turing/api/request/request.go
```diff
@@ -12,9 +12,9 @@ import (
 	routerConfig "github.com/caraml-dev/turing/engines/router/missionctl/config"
 )
 
-// CreateOrUpdateRouterRequest structure defines the format of the request payload
-// when creating or updating routers
-type CreateOrUpdateRouterRequest struct {
+// CreateRouterRequest structure defines the format of the request payload
+// when creating routers
+type CreateRouterRequest struct {
```
deadlycoconuts, line 17: Renamed this struct as it is no longer used for router updates.

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -305,6 +331,56 @@ func (sb *clusterSvcBuilder) NewSecret(
...
+	return string(*autoscalingPolicy.Metric)
+}
+
+func (sb *clusterSvcBuilder) getAutoscalingTargetOrDefault(autoscalingPolicy models.AutoscalingPolicy) string {
+	// TODO: default autoscaling policy is used; hence calculate target RPS from payload size
+	if autoscalingPolicy.PayloadSize != nil {
+		return "500"
+	}
+	return *autoscalingPolicy.Target
+}
```
deadlycoconuts, line 382: This autoscaling target has temporarily been set as 500 rps when the default autoscaling policy is selected. A future PR will refine this method to set the initial target depending on the initial payload size.

api/turing/validation/validator.go
```diff
@@ -313,58 +314,179 @@ func validateRouterConfig(sl validator.StructLevel) {
...
-			if router.DefaultTrafficRule != nil {
-				allRules := append(router.TrafficRules, &models.TrafficRule{
-					Name:   "default-traffic-rule",
-					Routes: router.DefaultTrafficRule.Routes,
-				})
-				validateDefaultRouteTrafficRules(sl, "TrafficRules", allRules, router.DefaultRouteID)
-				for _, route := range router.DefaultTrafficRule.Routes {
-					allRuleRoutesSet.Insert(route)
-				}
+		validateTrafficRules(sl, instance, router, routeIdsStr)
```
deadlycoconuts, line 318: Refactored the huge chunk of code into this helper function because of excessive cyclomatic complexity reported by one of our linters.

api/turing/models/autoscaling_policy.go
```diff
@@ -21,11 +22,13 @@ const (
 )
 
 type AutoscalingPolicy struct {
-	Metric AutoscalingMetric `json:"metric" validate:"required,oneof=concurrency rps cpu memory"`
+	Metric *AutoscalingMetric `json:"metric"`
```
krithika369, line 25: With this change, we are not testing the oneOf, right? We shouldn't remove everything and I think `omitempty` instead of `required` can help:

`json:"metric" validate:"omitempty,oneof=concurrency rps cpu memory"`

This could help trim down some of the special validation logic in `api/turing/validation/validator.go`

sdk/turing/router/router.py
```diff
@@ -144,7 +144,7 @@ def update(self, config: RouterConfig) -> "Router":
         self._config = config
         updated_router = Router.from_open_api(
             turing.active_session.update_router(
-                router_id=self.id, router_config=config.to_open_api()
+                router_id=self.id, router_version_config=config.to_open_api().config
```
krithika369, line 147: This is a breaking change to the users. Could you document it in the PR description? Ideally we can keep the API backward-compatible but in this instance, I feel that it may be ok to force the users to update the SDK after the app is released, so let's remember to document it in the release notes later.

ui/src/router/components/form/components/DefaultAutoscalingPolicyPanel.js
```diff
@@ -0,0 +1,69 @@
...
+          onChange={toggleUseDefaultAutoscalingPolicy}
+        />
+
+        {useDefaultAutoscalingPolicy &&
+          <>
+            <EuiSpacer size={"m"}/>
+
+            <DescribedFormGroup description={
+              "The value specified here will be used to perform autoscaling automatically. There is no need to specify " +
+              "and resource requests values or custom autoscaling values."
```
krithika369, line 42: I think the `and` here should be `any`, no?

sdk/turing/router/config/router_config.py
```diff
@@ -136,7 +138,9 @@ def routes(self) -> List[Route]:
 
     @routes.setter
     def routes(self, routes: Union[List[Route], List[Dict[str, str]]]):
-        if isinstance(routes, list) and all(isinstance(route, dict) for route in routes):
```
krithika369, line 139: For the router / enricher / ensembler config, we are setting the default autoscaling policy this way:
https://github.com/caraml-dev/turing/blob/main/sdk/turing/router/config/router_config.py#L91

If the user wants to set the payload size instead, what is the expectation? Should they explicitly clear the metric and target? Or shall we clear the metric/target if payload size is set, in the `to_open_api()` function ?

krithika369 (approved):
Thanks for the changes and the extensive refactoring! Left some comments - please feel free to merge to the feature branch when you've reviewed / addressed them!

api/api/openapi.bundle.yaml
```diff
@@ -2511,10 +2527,7 @@ components:
           $ref: '#/components/schemas/FieldSource'
         field:
           description: |
-            For HTTP_JSON protocol, the valid `field_source` are `header` and `payload`. Whereas, for UPI_V1 protocol the valid `field_source` are `header` and `prediction_context`.
-            If `field_source` is `header`, then `field` should contain the name of the request header.
-            If `field_source` is `payload`, then `field` should be a valid json path.
-            If `field_source` is `prediction_context`, then `field` should contain variable name stored in `prediction_context` field of the incoming request.
+            For HTTP_JSON protocol, the valid `field_source` are `header` and `payload`. Whereas, for UPI_V1 protocol the valid `field_source` are `header` and `prediction_context`. If `field_source` is `header`, then `field` should contain the name of the request header. If `field_source` is `payload`, then `field` should be a valid json path. If `field_source` is `prediction_context`, then `field` should contain variable name stored in `prediction_context` field of the incoming request.
```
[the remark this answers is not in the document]
deadlycoconuts: Ahaha I think this was removed when the OpenAPI generator regenerated the `bundle.yaml` file 😅 I'll manually edit it to put it back :)

ui/src/router/components/form/components/DefaultAutoscalingPolicyPanel.js
deadlycoconuts: Thanks for catching this! 😅

api/db-migrations/000016_make_resource_requests_not_null.down.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE router_versions
```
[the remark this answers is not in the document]
deadlycoconuts: Oh my bad, thanks for pointing them out; I've added additional commands for the migration scripts:

Up: 
- Added command to make the `resource_requests` column of the `enrichers` non-null (just like for the `router_versions` table)
- No changes for the `ensembler_configs` table as the `docker_config` and `pyfunc_config` columns are in jsonb (they don't impose any non-null constraints for the `resource_requests` field in them

Down:
- Added command to make the `resource_requests` column of the `enrichers` non-null and to introduce a `DEFAULT` values for null `resource_requests` (just like for the `router_versions` table)
- Update the `docker_config` and `pyfunc_config` columns of the `ensembler_configs` table by replacing their values with the union of their original value and a default `resource_requests` value

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -306,6 +332,56 @@ func (sb *clusterSvcBuilder) NewSecret(
 	}
 }
 
+func (sb *clusterSvcBuilder) getMinReplicaOrDefault(resourceRequest *models.ResourceRequest) (*int, error) {
```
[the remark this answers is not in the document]
deadlycoconuts: Ah I think my initial consideration was to address scenarios if the `defaultMinReplica` and `defaultMaxReplica` values were unset in the configs, but yeah as you mentioned, that case can simply be handled by setting default values. So yep, you're right, the pointers are unnecessary; will revert them to regular integers :)

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -306,6 +332,56 @@ func (sb *clusterSvcBuilder) NewSecret(
...
+	if resourceRequest == nil {
+		if sb.DefaultAutoscalingPolicyMinReplica == nil {
+			return nil, errors.New("API does not have min replica for default autoscaling policy set")
+		}
+		return sb.DefaultAutoscalingPolicyMinReplica, nil
+	}
+	return &resourceRequest.MinReplica, nil
+}
+
+func (sb *clusterSvcBuilder) getMaxReplicaOrDefault(resourceRequest *models.ResourceRequest) (*int, error) {
```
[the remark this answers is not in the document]
deadlycoconuts: That's true, that works too! Thanks for the suggestion!

api/turing/config/config.go
```diff
@@ -205,11 +205,18 @@ type SparkAppConfig struct {
...
+	EnvironmentType                string                         `validate:"required"`
+	Timeout                        time.Duration                  `validate:"required"`
+	DeletionTimeout                time.Duration                  `validate:"required"`
+	MaxCPU                         Quantity                       `validate:"required"`
+	MaxMemory                      Quantity                       `validate:"required"`
+	DefaultAutoScalingPolicyConfig DefaultAutoScalingPolicyConfig `validate:"required"`
+}
+
+type DefaultAutoScalingPolicyConfig struct {
+	Enabled           bool
```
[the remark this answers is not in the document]
deadlycoconuts: Yep it was to serve as a feature flag but yes I agree, it does seem unnecessary and convoluted to have to handle this value. Removing this while adding the default min/max replicas values as you've suggested.

sdk/turing/router/config/router_config.py
deadlycoconuts: Hmm I've decided to remove the `DEFAULT_AUTOSCALING_POLICY` that exists within the SDK and to make the router / enricher / ensembler configs all require an `AutoscalingPolicy` object as a required argument during initialisation. This is to make the expectation that this field should not be nil consistent with that of the API.

> If the user wants to set the payload size instead, what is the expectation? Should they explicitly clear the metric and target? Or shall we clear the metric/target if payload size is set, in the to_open_api() function ?

As for this, there are no checks that the SDK performs to ensure that only the `metric`/`target` fields or the `payload_size` field is set. However, any configs in requests sent to the API will ultimately be validated by its own validator, which contains checks to ensure that not all the fields are set, failing which the API would return an error to the SDK during execution. This is similar to the current behaviour of the standard ensembler whereby the only checks that take place to ensure only the `route_name_path` or `experiment_mappings` field is set occur in the API.

PR #277, opened 2022-11-01
WIP upgrade knative to 1.3.2

leonlnj:


krithika369 (commented):
Not sure if any other code changes are needed for this PR, but the current changes LGTM. One question - is the API dependency change backward compatible? I.e., is it possible to keep Knative serving and net-istio components at the older version `1.0.1` / `1.0.0 ` and upgrade the API dependency alone?

Perhaps the e2e test in GitHub can be run with the old infra + new API to determine this.

leonlnj:
>  is the API dependency change backward compatible? I.e., is it possible to keep Knative serving and net-istio components at the older version 1.0.1 / 1.0.0 and upgrade the API dependency alone?

Which API dependencies are you referring to? Are you referring for the new changes to API server compatible with 1.0.1?

krithika369:
> Not sure if any other code changes are needed for this PR, but the current changes LGTM. One question - is the API dependency change backward compatible? I.e., is it possible to keep Knative serving and net-istio components at the older version `1.0.1` / `1.0.0` and upgrade the API dependency alone?
> 
> Perhaps the e2e test in GitHub can be run with the old infra + new API to determine this.

Sorry for not being clear. It seems I accidentally removed one of the sentences. My question was - shall we update the serving client as well ([the dependency in the API mod file](https://github.com/caraml-dev/turing/blob/main/api/go.mod#L54))? And if so, we should check that it's compatible both with serving controller `1.0.1` and the new `1.3.2`.

But I remembered that kserve's dependency on knative is also at 0.x: https://github.com/kserve/kserve/blob/v0.8.0/go.mod#L37

So we can revisit this later.

leonlnj:
checked that we wont be upgrading the knative in this upgrade, as the current one is still working and we should be in sync with merlin. 

Anyway will close this and address together in https://github.com/caraml-dev/turing/pull/282

PR #278, opened 2022-11-02
Modify error rate's alert query that use UPI protocol

tiopramayudi:
Due to knative revision metrics treated all grpc response code as 200, we need to modify alert query regarding error rates. Modification is by using `istio_requests_total` metric if `grpc` protocol being used and using existing `revision_request_count` query if `http` protocol being  used


krithika369:
> if grpc protocol being used and using existing revision_request_count query if http protocol being used

Since we are creating an Istio VS for the router from the Turing codebase, I think we can assume the Istio metric will always be there. Given this, shall we default to using the Istio metrics for both http and grpc instead of selectively handling this?

tiopramayudi:
> > if grpc protocol being used and using existing revision_request_count query if http protocol being used
> 
> Since we are creating an Istio VS for the router from the Turing codebase, I think we can assume the Istio metric will always be there. Given this, shall we default to using the Istio metrics for both http and grpc instead of selectively handling this?

Update this using metric from `istio_requests_count` but still need to do the condition since we rely on different labels for status code between grpc and http

krithika369:
Not important to do this, but for avoiding any confusing, can we update the sample alert in https://github.com/caraml-dev/turing/blob/main/api/turing/service/alert_service_test.go#L34, to no longer use `revision_request_count` ?

api/turing/models/alert.go
```diff
@@ -146,20 +146,28 @@ func getAlertExpr(metric Metric, env string, rev string, threshold float64) stri
   service_name=~"%s-[0-9]*"
 }[1m])) by (le)) %s %f
 `, env, rev, getAlertOperator(metric), threshold)
-
 	case MetricErrorRate:
-		return fmt.Sprintf(`sum(rate(revision_request_count{
```
krithika369, line 151: What about the `revision_request_latencies_bucket` metric with UPI? Is it expected to still work, or does it have to be changed?

krithika369 (approved):
Left some comments. The rest LGTM. Thanks, @tiopramayudi !

tiopramayudi:
> Not important to do this, but for avoiding any confusing, can we update the sample alert in https://github.com/caraml-dev/turing/blob/main/api/turing/service/alert_service_test.go#L34, to no longer use `revision_request_count` ?

The alert that must use `istio` metrics is only the error rate due to revision metric could not differentiate  ok or fail status code when using grpc, hence the only query that change only the error rate for throughput and latency we can still rely on the `revision` metrics

PR #283, opened 2022-11-11
Fix some of grpc client that get stuck when access grpc server

tiopramayudi:
Since we use cmux to multiplex connection to grpc or http, there is match rule that will redirect request to correct server. Current match rule that we use won't work for some of grpc client due to some of grpc client will blocks until it receives `SETTINGS` frame from the server. In this PR we change the match rule so it can handle all the grpc client

ref: [https://github.com/soheilhy/cmux#limitations](https://github.com/soheilhy/cmux#limitations)


leonlnj (approved):
thanks for this!

PR #284, opened 2022-11-11
Support Lazy routing with Standard Ensembler

krithika369:
This PR introduces an option to call the experiment engine before the routing step, when using a standard ensembler. This makes it possible to only invoke the route that will be returned as the final response, based on the treatment retuned by the experiment engine.

##  Illustration

(With the [XP experiment engine](https://github.com/caraml-dev/xp) as the example)

### Edit Router

standard_ensembler_lazy_routing

### View Router

Screenshot 2022-11-12 at 1 44 50 AM

## Summary of Changes
* `api/api/specs/routers.yaml`, `api/turing/models/ensembler.go` - Add `lazy_routing` as a field in the Standard Ensembling config
* `api/turing/cluster/servicebuilder/router.go` - Select Fiber `LAZY_ROUTER` when the `lazy_routing` property of the standard ensembling is selected.
* `sdk/**` - Update generated code, add `lazy_routing` property to `EnsemblerStandardConfig` and `StandardRouterEnsemblerConfig` data classes (`sdk/turing/router/config/router_ensembler_config.py`) and add unit tests.
* `ui/src/router/components/configuration/components/standard_config_section/RoutingOrderConfigSection.js` - Add Routing Order details panel for standard ensembler
* `ui/src/router/.../StandardEnsemblerRoutingOrderPanel.js` - Add routing oder panel to the router form


sdk/e2e/11_deploy_router_upi_traffic_split_test.py
```diff
@@ -15,10 +15,14 @@
 from turing.router.config.resource_request import ResourceRequest
 from turing.router.config.log_config import LogConfig, ResultLoggerType
 from turing.router.config.router_config import RouterConfig, Protocol
-from turing.router.config.router_ensembler_config import EnsemblerStandardConfig, RouterEnsemblerConfig, \
-    StandardRouterEnsemblerConfig
+from turing.router.config.router_ensembler_config import StandardRouterEnsemblerConfig
```
krithika369, line 18: This file and the other SDK-based e2e tests have simply been re-formatted (and some unused imports removed).

sdk/tests/conftest.py
```diff
@@ -119,6 +120,17 @@ def standard_router_ensembler_config_with_route_name_path():
         experiment_mappings=None,
         route_name_path="route_name",
         fallback_response_route_id="route-1",
+        lazy_routing=False,
```
krithika369, line 123: `EnsemblerStandardConfig` should be initialised with `lazy_routing`. It's not necessary to set this parameter when using `StandardRouterEnsemblerConfig` which will be initialised by the users (default value for the parameter is `False`).

sdk/turing/router/config/router_config.py
```diff
@@ -30,6 +30,7 @@
 NAME_INDEX = 0
 VALUE_INDEX = 1
 
+
```
krithika369, line 33: This file has simply been auto-formatted.

sdk/turing/router/config/route.py
```diff
@@ -3,6 +3,7 @@
 from urllib.parse import urlparse
 from dataclasses import dataclass
 
+
```
krithika369, line 6: This file has simply been autoformatted.

leonlnj (approved):
LGTM. 

Conceptually just one flag to that change the eager->lazy in fiber yaml 😅 everywhere, the size made it look like there were more to it. 

No comment other than it is abit spacious here
image

Perhaps we can use the select box and have `Standard` and `Lazy` option to have a more consistent view
image

No strong opinion, its LGTM overall, thanks!

ui/package.json
```diff
@@ -47,7 +47,7 @@
     "start": "PORT=3001 craco start --verbose",
     "build": "craco build",
     "test": "craco test",
-    "lint": "eslint --max-warnings 0 src"
+    "lint": "eslint --max-warnings 0 --ext .js,.jsx --fix src"
```
deadlycoconuts, line 50: What's this for :O

deadlycoconuts (approved):
LGTM! 🚀 Thanks for adding this option in! Don't have any comments except maybe @leonlnj's that the checkbox is a bit... strange in the current way it looks 😅 Can we use `EuiSpacer`s to realign the checkbox to the middle?

ui/package.json
krithika369: * The `--ext` flag is not really necessary, it's just more explicit what kind of files to try to scan.
* The `--fix` flag will auto-format whatever possible: https://masteringjs.io/tutorials/eslint/fix

deadlycoconuts: Ahhh I see, thanks for adding them in! I was just wondering what the motivation for adding them was :)

krithika369:
Thanks for the review @deadlycoconuts and @leonlnj. I've updated the layout in my latest commit. It looks like [this](https://github.com/caraml-dev/turing/blob/b6b44b720e73c9f7a2d6e57c7efb9a6c7a21cd5f/docs/.gitbook/assets/standard_ensembler_lazy_routing.png) now. I realised the `DescribedFormGroup` reserves a lot more space for the form components than needed, for all its current uses. So I updated the base component. If no other opinions on this, I will merge this PR once the CI passes.

PR #287, opened 2022-11-16
Bugfix: Add env var for exp engine secret path

deadlycoconuts:
## Context 
In PR #274, changes were made to the way experiment engine secrets were passed to the plugin manager of the experiment engine. One of these secrets is the Google service account token (stored in a json file), which gets mounted as a volume within a Turing Router deployment. The experiment engine plugin (run as a subprocess of the Turing Router and hence is run within the same container as it) is then supposed to access this file using a file path stored in an environment variable. 

However, it is possible for the Turing Router and the experiment engine to have separate Google service accounts instead of the same one. Unfortunately, PR #274 has missed out specifying a file path to the service account token of the experiment engine. While the file path of the router's service account token remains stored in the env variable `GOOGLE_APPLICATION_CREDENTIALS`, there is a need to create another environment variable that points to the location of the experiment engine service account.

This PR does just that and injects an additional environment variable `EXP_GOOGLE_APPLICATION_CREDENTIALS` for the said purpose.


api/turing/cluster/servicebuilder/router.go
```diff
@@ -25,38 +25,39 @@ import (
...
+	envFluentdTag                      = "APP_FLUENTD_TAG"
+	envKafkaBrokers                    = "APP_KAFKA_BROKERS"
+	envKafkaTopic                      = "APP_KAFKA_TOPIC"
+	envKafkaSerializationFormat        = "APP_KAFKA_SERIALIZATION_FORMAT"
+	envKafkaMaxMessageBytes            = "APP_KAFKA_MAX_MESSAGE_BYTES"
+	envKafkaCompressionType            = "APP_KAFKA_COMPRESSION_TYPE"
+	envRouterConfigFile                = "ROUTER_CONFIG_FILE"
+	envRouterProtocol                  = "ROUTER_PROTOCOL"
+	envGoogleApplicationCredentials    = "GOOGLE_APPLICATION_CREDENTIALS"
+	envExpGoogleApplicationCredentials = "EXP_GOOGLE_APPLICATION_CREDENTIALS"
```
krithika369, line 58: Super nit: can this be `GOOGLE_APPLICATION_CREDENTIALS_EXPERIMENT_ENGINE` ? I think it's nicer if the name starts with `GOOGLE_APPLICATION_CREDENTIALS` and replacing `EXP` with `EXPERIMENT_ENGINE` would be clearer.

krithika369 (approved):
LGTM. Thanks for this fix.

PR #288, opened 2022-11-17
Support configurable worker count for Fluentd

terryyylim:
This PR adds configuration to support worker counts in Fluentd. With more traffic, Fluentd tends to be more CPU bound. In this case, we should consider using [multi-worker](https://github.com/deployment/multi-process-workers) feature.


krithika369 (approved):
LGTM!

krithika369:
One thought - should we set the `FLUENTD_WORKER_COUNT` env var to be `1` from the API for now? If not, the latest fluentd image may not be compatible with the router.

PR #291, opened 2022-11-25
Update image builder gcloud auth

krithika369:
In clusters where workload identity is enabled, `gsutil` fails to use the user-defined service account key file even if set as `GOOGLE_APPLICATION_CREDENTIALS`. To address this, the service account must be explicitly activated using `gcloud auth activate-service-account` command.


leonlnj (approved):
LGTM, thanks!

PR #292, opened 2022-11-25
Update Batch Ensembler job Runner's Image builder environment

krithika369:
This PR makes it possible to use a non user-facing environment for building the Ensembler images. Typically, there are 2 types of clusters - one for the deployment of the Turing app and another for deploying user resources (i.e., the routers, batch workloads, etc.). When a Pyfunc ensembler (real-time / batch) is used, it needs Kaniko, for being able to build the image.

### Summary of Changes
* `api/turing/config/config.go` - Rename `EnsemblerServiceBuilderConfig`'s `DefaultEnvironment` to `ClusterName`. This property will capture the actual cluster name as opposed to a user-facing environment name that's registered in Merlin.
* `api/turing/api/appcontext.go` - Add image-builder cluster info to the cluster controllers. Create image-builder controller using the correct cluster configuration.
* `infra/charts/turing/` - Update chart


terryyylim (approved):

api/turing/api/appcontext.go
```diff
@@ -114,19 +114,22 @@ func NewAppContext(
 			mlpSvc,
 		)
 
-		batchClusterController, ok := clusterControllers[cfg.BatchEnsemblingConfig.JobConfig.DefaultEnvironment]
+		imageBuildingController, ok := clusterControllers[cfg.EnsemblerServiceBuilderConfig.ClusterName]
```
deadlycoconuts, line 117: I find it a little strange that the `batchEnsemblingJobRunner` would have its `imageBuildingController` be built using a cluster controller that's derived from the `EnsemblerServiceBuilderConfig`. Not a huge priority or a big problem but would it be better to move the `ClusterName` config to be under `ImageBuildingConfig`?

i.e. 
```yaml
BatchEnsemblingConfig:
  ImageBuildingConfig: &imageBuildingConfig
    ClusterName: build-cluster
    BuildNamespace: build-namespace
    ...
```
such that we get the same cluster name automatically copied over to the ensembler service builder config too?  
```yaml
EnsemblerServiceBuilderConfig:
  ImageBuildingConfig:
  <<: *imageBuildingConfig
  ...
```

api/turing/api/appcontext.go
```diff
@@ -114,19 +114,22 @@ func NewAppContext(
...
 			*cfg.BatchEnsemblingConfig.ImageBuildingConfig,
 		)
 		if err != nil {
 			return nil, errors.Wrapf(err, "Failed initializing ensembling image builder")
 		}
 
+		batchClusterController, ok := clusterControllers[cfg.BatchEnsemblingConfig.JobConfig.DefaultEnvironment]
+		if !ok {
+			return nil, errors.Wrapf(err, "Failed getting the batch ensembling job controller")
+		}
```
deadlycoconuts, line 132: Thanks for this crucial change!

deadlycoconuts (approved):
LGTM except for that one tiny comment! Thanks for getting the changes out so quickly! 🚀

api/turing/api/appcontext.go
krithika369: Given the current usage, I agree that the `ClusterName` can be embedded into the `ImageBuildingConfig` which also makes the configs more concise. Will address this in the next MR. Thanks for raising.

PR #297, opened 2022-11-29
Add max replica limits to Turing API server 

deadlycoconuts:
## Context
The Turing API server currently does not have any limits set on the max replica values that users can use for the Turing Router, Enricher and Ensembler pods that they deploy, unlike in the SDK or in the UI. This PR introduces changes to standardise these limits across the three user-facing interfaces by performing the following:
1. Add checks to the API server to ensure that any max replica values set by the user for any of the components (Turing Router, Enricher and Ensembler) fall below a limit that is specified within a config file
2. Remove the existing max replica checks from the SDK (we will now rely on the API response to return any errors with the max replica validation)


api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -323,6 +326,9 @@ func (sb *clusterSvcBuilder) validateKnativeService(
 	if svc.MemoryRequests.Cmp(sb.MaxMemory) > 0 {
 		return nil, errors.New("Requested Memory is more than max permissible")
 	}
+	if svc.MaxReplicas > sb.MaxAllowedReplica {
+		return nil, errors.New("Requested max replica is more than max permissible")
```
terryyylim, line 330: nit: `"Requested Replica is more than max permissible"`

terryyylim (approved):
LGTM, thanks @deadlycoconuts !

deadlycoconuts:
Thanks a lot for the quick approval! Will be merging this when the pipeline passes!

leonlnj:
CMMIW 
Before -> After 
API limit 0 -> 30 (configurable)
UI 30 -> 30
SDK 30 -> 0

PR #298, opened 2022-11-30
UI remote component and fiber lazy routing bugfix

leonlnj:
## Description

1. UI Bugfix for standard experiment engine to propagate protocol field
2. Traffic routes were not showing up for nested lazy routers (traffic rules with selective option) due to label map being overwritten and not [propagated to the caller type ](https://github.com/caraml-dev/turing/blob/main/engines/router/missionctl/fiberapi/interceptors.go#L149).
Fiber version bump for bugfix - https://github.com/gojek/fiber/pull/10

## Testing
1. Variables panel showing
image

2. Labels showing up traffic rules when calling `/metrics` endpoint locally
Screenshot 2022-12-01 at 6 48 31 PM


deadlycoconuts (approved):
LGTM! 🚀 Thanks for investigating this!

PR #299, opened 2022-12-02
revert fiber to 2.0

leonlnj:
## Description

In fiber 2.0, when Traffic Rules are used with `Selective` routing strategy in standard ensembler setup. Underlying, the router are composed of two nested `lazy router`. With the default `Exhaustive` routing strategy, it is composed of a parent `lazy router` (which is used for traffic splitting) and child `eager router` (to call all routers in parallel).

Fiber 2.1-rc1 fixes the above problem with route request duration metric but introduce another bug to route component metric To explain in context

**Fiber 2.0**
Step 1 Parent layer - lazy router (for both routing strategy)
1. Lazy router calls `TrafficSplittingStrategy` and obtain the traffic rule labels
2. Invoke child common with traffic labels added to context

Step 2a Second layer - eager router (exhaustive strategy)
1. Calls all routes with context passed (contain traffic label, which logs metric correctly)
2. Calls `DefaultTuringRoutingStrategy` and get empty label
3. Return with no labels

Step 2b Second layer - lazy router (selective strategy)
1. Calls `DefaultTuringRoutingStrategy` and get empty label
2. Calls all routes with context passed (contain NO traffic label, as it is pass the label from `DefaultTuringRoutingStrategy` which logs metric incorrectly)
3. Return with no labels

Step 3 Back to first parent layer -  lazy router (for both routing strategy)
1. Both child pass back response with no label
2. Return response with traffic labels from `TrafficSplittingStrategy`

**Fiber 2.1-rc1**
Step 1 Parent layer - lazy router (for both routing strategy), no change from the above
1. Lazy router calls `TrafficSplittingStrategy` and obtain the traffic rule labels
2. Invoke child common with traffic labels added to context

Step 2a Second layer - eager router (exhaustive strategy)
1. Calls all routes with context passed (contain traffic label, which logs metric correctly)
2. Calls `DefaultTuringRoutingStrategy`, get empty label and merge labels with parent's label
4. Return with merged labels

Step 2b Second layer - lazy router (selective strategy)
1. Calls `DefaultTuringRoutingStrategy` , get empty label and merge labels with parent's label
2. Calls all routes with context passed (contain traffic label, as it merged with parent label, which logs metric correctly)
3. Return  with merged labels

Step 3 Back to first parent layer -  lazy router (for both routing strategy)
1. Both child pass back response with merged label (problem occurs here)
2. Return response with traffic labels from `TrafficSplittingStrategy` by merging back return label, causing double labelling

Fundamentally, this behaviour occur due to eager and lazy router calling routes with different sequence.
- eager router call route with parent context, then call strategy
- lazy router call strategy, then call route with strategy context

Hence this PR reverts Fiber 2.1-rc1, until a solution fixes both metrics.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-playground/validator v9.31.0+incompatible
 	github.com/go-playground/validator/v10 v10.9.0
-	github.com/gojek/fiber v0.2.1-rc1
+	github.com/gojek/fiber v0.2.0
 	github.com/gojek/merlin v0.0.0
 	github.com/gojek/mlp v1.4.7
 	github.com/golang-collections/collections v0.0.0-20130729185459-604e922904d3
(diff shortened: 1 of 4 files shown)


deadlycoconuts:
Thanks for the deep investigation into this! We'll have to find another fix somehow...🤔 Temporarily approving this so we can cut the UI fixes for `v1.8.1`.

deadlycoconuts (approved):

PR #300, opened 2022-12-05
Add NodeSelector and TolerationName to ImageBuildingConfig

deadlycoconuts:
## Context
This PR introduces two additional configuration to the `ImageBuildingConfig`:

1. `NodeSelector` which allows operators of the Turing API server to set labels such that the image building jobs (and the pods they spin up) only scheduled on nodes with those labels and,
2. `TolerationName` which allows operators of the Turing API server to set a toleration name such that image building jobs will be deployed with the specified toleration name, allowing them to be scheduled on nodes with the matching taint.

### DIFF
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -159,6 +159,8 @@ type ImageBuildingConfig struct {
 	BaseImageRef map[string]string `validate:"required"`
 	// KanikoConfig contains the configuration related to the kaniko executor image builder.
 	KanikoConfig KanikoConfig `validate:"required"`
+	// NodeSelector restricts the running of image building jobs to nodes with the specified labels
+	NodeSelector map[string]string
 }
 
 // Resource contains the Kubernetes resource request and limits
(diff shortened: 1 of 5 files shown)


api/turing/config/config.go
```diff
@@ -159,6 +159,10 @@ type ImageBuildingConfig struct {
 	BaseImageRef map[string]string `validate:"required"`
 	// KanikoConfig contains the configuration related to the kaniko executor image builder.
 	KanikoConfig KanikoConfig `validate:"required"`
+	// TolerationName allow the scheduler to schedule image building jobs with the matching name
+	TolerationName *string
```
deadlycoconuts, line 163: This is somewhat different from how Merlin saves this particular config (https://github.com/gojek/merlin/blob/83199a79831687bb9b230f5a3029716f05cd448e/api/config/config.go#L156) but I decided to store it like this because it's how we currently storing tolerations in the `SparkAppConfig` here https://github.com/caraml-dev/turing/blob/main/api/turing/config/config.go#L197.

leonlnj (approved):
thanks!

deadlycoconuts:
Thanks for the quick approval! 🙏

PR #302, opened 2022-12-15
Add method to register metrics for experiment engines

deadlycoconuts:
## Context
The experiment engine plugin, as of present, offers a simple interface that allows plugin writers to specify the logic needed to generate a treatment for each request a Turing Router would receive. Additional functionalities required by any implementation of such plugins to generate a treatment (such as the logging of metrics) would need to be implemented separately by the plugin writer and are run independently from the Turing Router. 

However, some of these functionalities, in particular metric logging, are also common to the Turing Router; instead of running them independently in a separate child process, the plugin could instead call upon the Turing Router's implementation of these functionalities within the main process. In the case of metric logging, the experiment engine plugin could use the existing metric logging functionality of the Turing Router (its metric logger, methods, etc.) instead of having to initialise and manage its own metric logging service. 

This PR hence introduces an additional method to the `ExperimentRunner` interface - `RegisterMetricsCollector`. This method allows the experiment runner to register a metrics collector (see the `Collector` interface [here](https://github.com/gojek/mlp/blob/main/api/pkg/instrumentation/metrics/metrics.go#L19)) that the Turing Router uses and to interact with it directly. 

The aforementioned method also has a `MetricsRegistrationHelper` argument that allows the experiment plugin to register additional metrics by sending their details to the Router.

## Workflow
metricsworkflow2

1. After initialising an `ExperimentRunner`, the Turing Router then calls the `RegisterMetricsCollector` method, with its own metrics collector as well as an implementation of `MetricsRegistrationHelper` as arguments, to the experiment engine plugin.
2. The experiment engine then does something (to be implemented by the plugin writer) with the given metrics collector (a reference to it can be stored somewhere). Additional metrics can be registered using the `MetricsRegistrationHelper`.
3. Subsequently, any logging done by the experiment engine will utilise the actual implementations on the Turing Router.

## Additional Refactoring
The implementations of the `PrometheusClient` class and the `Collector` interface used in the Turing Router have been **abandoned** in favour of those from [gojek/mlp](https://github.com/gojek/mlp/tree/main/api/pkg/instrumentation/metrics). This standardisation of the structs and interfaces used for logging is in line with other CaraML products (e.g. XP).

## ⚠️ Breaking Changes 🚨🚨🚨
All experiment engines need to be updated to implement the new `ExperimentRunner` interface which has a new `RegisterMetricsCollector` method:
```go
// ExperimentRunner is the generic interface for generating experiment configs
// for a given request
type ExperimentRunner interface {
	// ...

	// RegisterMetricsCollector is a method that should only be called on startup of the experiment runner to register
	// the metrics collector of the Turing Router (to be called just after the experiment runner has been initialised)
	RegisterMetricsCollector(
		collector metrics.Collector,
		metricsRegistrationHelper MetricsRegistrationHelper,
	) error
}
```
If there is no need to use the Turing Router's metrics collector within the plugin, a dummy implementation of `RegisterMetricsCollector` can simply be used to avoid unnecessary changes:
```go
func (r *ExperimentRunner) RegisterMetricsCollector(_ metrics.Collector, _ runner.MetricsRegistrationHelper) error {
	return nil
}
```


api/e2e/test/cluster/cluster.go
```diff
@@ -57,7 +57,7 @@ func GetModelClusterCredentials(clusterName string, vaultCfg config.VaultConfig)
 		Address: vaultCfg.Address,
 		Token:   vaultCfg.Token,
 	}
-	vaultClient, err := vault.NewVaultClient(vaultConfig)
+	vaultClient, err := vault.NewClient(vaultConfig)
```
deadlycoconuts, line 60: This change is made due to an update in the `gojek/mlp` version which feature new method names for the vault package.

api/turing/api/appcontext.go
```diff
@@ -44,7 +44,7 @@ func NewAppContext(
 	db *gorm.DB,
 	cfg *config.Config,
 	authorizer *middleware.Authorizer,
-	vaultClient vault.VaultClient,
+	vaultClient vault.Client,
```
deadlycoconuts, line 47: This change is made due to an update in the gojek/mlp version which feature new struct names in the vault package.

api/turing/api/deployment_controller.go
```diff
@@ -161,7 +161,7 @@ func (c RouterDeploymentController) deployRouterVersion(
 
 	if routerVersion.LogConfig.ResultLoggerType == models.BigQueryLogger {
 		routerServiceAccountKey, err = c.MLPService.GetSecret(
-			models.ID(project.Id),
+			models.ID(project.ID),
```
deadlycoconuts, line 164: This change is made due to an update in the `gojek/mlp` version which features new attribute names for the `Project` struct.

engines/router/missionctl/config/config_test.go
```diff
@@ -61,9 +61,9 @@ var optionalEnvs = map[string]string{
 	"APP_JAEGER_COLLECTOR_ENDPOINT":  "http://localhost:5000",
 	"APP_JAEGER_REPORTER_HOST":       "localhost",
 	"APP_JAEGER_REPORTER_PORT":       "5001",
-	"SENTRY_ENABLED":                 "true",
-	"SENTRY_DSN":                     "test:dsn",
-	"SENTRY_LABELS":                  "sentry_key1:value1,sentry_key2:value2",
+	"APP_SENTRY_ENABLED":             "true",
+	"APP_SENTRY_DSN":                 "test:dsn",
+	"APP_SENTRY_LABELS":              "sentry_key1:value1,sentry_key2:value2",
```
deadlycoconuts, line 66: The additional `APP_` prefix is needed because the `envconfig` tags have been removed from post `v1.4.7` of `gojek/mlp`: see https://github.com/gojek/mlp/pull/29/files and https://github.com/gojek/mlp/compare/v1.4.7...v1.4.8.

sdk/requirements.txt
```diff
@@ -4,7 +4,7 @@ fire
 google-auth>=1.11.0
 google-cloud-storage>=1.19.0
 mlflow>=1.2.0,<=1.23.0
-numpy
+numpy<1.24.0
```
deadlycoconuts, line 7: This change is needed due to a breaking change from numpy v1.24.0 that is incompatible with our pinned versions of mlflow due to the deprecation of several common numpy aliases (see the last bullet point here: https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations).

engines/experiment/plugin/rpc/runner/plugin.go
```diff
@@ -12,10 +14,38 @@ type ExperimentRunnerPlugin struct {
 	Impl ConfigurableExperimentRunner
 }
 
-func (p *ExperimentRunnerPlugin) Server(*plugin.MuxBroker) (interface{}, error) {
-	return &rpcServer{Impl: p.Impl}, nil
+func (p *ExperimentRunnerPlugin) Server(b *plugin.MuxBroker) (interface{}, error) {
+	return &rpcServer{Impl: p.Impl, MuxBroker: b}, nil
```
krithika369, line 18: The MuxBroker is added now for the bidirectional call, correct?

engines/experiment/runner/runner.go
```diff
@@ -23,4 +32,8 @@ type ExperimentRunner interface {
 		payload []byte,
 		options GetTreatmentOptions,
 	) (*Treatment, error)
+	RegisterMetricsCollector(
```
krithika369, line 35: We need to update the dev docs for the RPC plugin ([here](https://github.com/caraml-dev/turing/blob/main/engines/experiment/docs/rpc_plugins.md)) with the details of the new method. This can be done in a separate PR too. But, in this one, it will be good to add some inline comments for the 2 runner methods to capture their summary, especially that the `RegisterMetricsCollector` method is only called on startup.

krithika369 (approved):
Thanks for the detailed PR description and thorough changes. I only left some small comments. LGTM. 🎉

krithika369:
With the addition of the new runner interface method, existing plugin implementations will need to be updated to provide at least a dummy implementation for the new method, right? Could you help make a note of this in the PR description? I'm also thinking about using the labels more consciously to classify the PRs. I added a new "breaking" label to the labels list - shall we set that on this PR too?

sdk/requirements.txt
deadlycoconuts: Added that in the file :)

engines/experiment/plugin/rpc/runner/plugin.go
deadlycoconuts: Yep that's right.

engines/experiment/runner/runner.go
deadlycoconuts: Yep that makes sense, given all the changes we're making to the experiment engine interfaces. I've just added inline comments to the methods and structs introduced to this file. I'll add additional dev docs to capture more details in a separate PR.

deadlycoconuts:
> With the addition of the new runner interface method, existing plugin implementations will need to be updated to provide at least a dummy implementation for the new method, right? 
Yep that's right. I've updated the examples to reflect the necessary updates but yes we'll need to make this requirement more explicit.

> Could you help make a note of this in the PR description? I'm also thinking about using the labels more consciously to classify the PRs. I added a new "breaking" label to the labels list - shall we set that on this PR too?
Ok, I'll edit the description to include a note and use the new label 🙂

deadlycoconuts:
Thanks a lot for the looking through this long PR! 🙏 I'll be merging this once the pipeline passes! 🚀

PR #304, opened 2023-01-18
fix json formatting of logs proto

leonlnj:
## Description 
This PR add marshal option to the protojson marshaller to use proto name (separate words by underscore instead of camelCase) and a test for logging that will assert the format. Also to address a previous [PR](https://github.com/caraml-dev/turing/pull/301/files)  `engines/router/missionctl/server/upi/logutils.go` whereby the request was logged as response without noticing.

Using the default marshaller, the proto keys are camelcase which is not consistent
Screenshot 2023-01-18 at 6 34 34 PM

### DIFF
--- a/engines/router/missionctl/server/upi/logutils_test.go
+++ b/engines/router/missionctl/server/upi/logutils_test.go
@@ -4,16 +4,23 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"time"
 
+	"github.com/caraml-dev/turing/engines/router/missionctl/config"
+	"github.com/caraml-dev/turing/engines/router/missionctl/log"
+	"github.com/caraml-dev/turing/engines/router/missionctl/log/resultlog"
 	upiv1 "github.com/caraml-dev/universal-prediction-interface/gen/go/grpc/caraml/upi/v1"
 	fiberProtocol "github.com/gojek/fiber/protocol"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zaptest/observer"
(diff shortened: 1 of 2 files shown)


engines/router/missionctl/server/upi/logutils_test.go
```diff
@@ -52,3 +59,94 @@ func Test_logTuringRouterRequestSummary(t *testing.T) {
...
+			logger := zap.New(core)
+			log.SetGlobalLogger(logger.Sugar())
+			err := resultlog.InitTuringResultLogger(&config.AppConfig{
+				ResultLogger: config.ConsoleLogger,
+			})
+			assert.NoError(t, err)
+
+			//Add response to channel and invoke logging of summary
+			respCh := make(chan grpcRouterResponse, 1)
+			respCh <- grpcRouterResponse{header: tt.reqHeader, body: tt.resBody, key: resultlog.ResultLogKeys.Router}
```
krithika369, line 127: Shouldn't this be setting `resHeader` instead of `reqHeader` ?

engines/router/missionctl/server/upi/logutils_test.go
```diff
@@ -52,3 +59,94 @@ func Test_logTuringRouterRequestSummary(t *testing.T) {
...
+		expectedResHeader string
+		expectedResBody   string
+	}{
+		{
+			name:      "ok",
+			reqHeader: header,
+			reqBody: &upiv1.PredictValuesRequest{
+				PredictionTable: table,
+			},
+			resHeader: header,
```
krithika369, line 102: Would be better if the `resHeader` and `reqHeader` are different values, to also test that they are not wrongly captured in the logs, no?

krithika369 (approved):
Left some small comments. LGTM.

PR #305, opened 2023-01-19
Upgrade Knative Serving component to 1.7.4

krithika369:
This PR bumps up the default Knative infrastructure dependencies installed by the cluster init chart, as well as the dependencies used in the e2e tests as follows:
* `KNATIVE_VERSION`: `1.3.2` -> `1.7.4`
* `KNATIVE_ISTIO_VERSION`: `1.3.0` -> `1.7.1`

Therefore, the minimum Kubernetes version required is now `1.22`.

### Motivation
The current version of Knative has a bug in the liveness probe, fixed by https://github.com/knative/serving/pull/13229, which is only available in Knative `1.7.x` ([release notes](https://github.com/knative/serving/releases)). This bug currently affects Merlin (corresponding PR: https://github.com/gojek/merlin/pull/309).

**Note:** The latest version of Knative serving is `1.8.x` but it has a handful of breaking changes that may make the update harder. It also has a min K8s requirement of `1.23`. So, we're sticking with `1.7.x` for now.


.github/actions/run-cluster-init/action.yaml
```diff
@@ -31,7 +31,7 @@ runs:
   using: composite
   steps:
     - name: "Setup local k8s cluster"
-      uses: AbsaOSS/k3d-action@v2.0.0
+      uses: AbsaOSS/k3d-action@v2.3.0
```
krithika369, line 34: This bumps the k8s version installed from `1.21.5` to `1.22.7`

deadlycoconuts (approved):
LGTM! 🚀 Thanks for testing the changes out too!

krithika369:
> LGTM! 🚀 Thanks for testing the changes out too!

@deadlycoconuts the e2e tested it. 😛 Thanks for the quick review!

PR #306, opened 2023-01-20
Update Postgres Dependency in Helm charts

krithika369:
This PR aims to address the recent [CI error on main branch](https://github.com/caraml-dev/turing/actions/runs/3964075353/jobs/6792602502).

The Postgres chart being used no longer exists. Upgrading the dependency to the latest chart version `12.1.9` (which upgrades the Postgres image to `15.11.0-debian-11-r20` by default, but has been overridden in the values file, to `12.13.0`, primarily for the compatibility with the `migrate/migrate` image used by MLP to run the DB migrations : [ref](https://artifacthub.io/packages/helm/bitnami/postgresql?modal=values)).

Many of the existing values for Postgres configs are outdated with the new chart and accordingly, the default values and Helm docs are updated.


infra/charts/turing/subcharts/merlin/values.yaml
```diff
@@ -1,5 +1,6 @@
 global:
   merlin: {}
+  mlp: {}
```
krithika369, line 3: Adding this for completeness. When installing the Turing chart, the master values file will define this, so it is not a problem. But if trying to dry-run just the Merlin chart, the absence of this field is problematic.

infra/charts/turing/values.yaml
```diff
@@ -253,10 +253,6 @@ mlp:
   postgresql:
     # -- Name of MLP's Postgresql deployment
     nameOverride: postgresql-mlp
-    # -- Password for MLP Postgresql database
-    postgresqlPassword: mlp
```
krithika369, line 257: This is now moved under a nested `auth` field and is anyway defined directly in the subchart with the same value. I believe it was added here as well as an example, can be cleaned up.

infra/charts/turing/values.yaml
```diff
@@ -253,10 +253,6 @@ mlp:
...
-    tls:
```
krithika369, line 258: This is already the default config now, removing.

.github/actions/setup-test-cluster/action.yaml
```diff
@@ -193,6 +193,10 @@ runs:
         timeout: 600 # seconds
         max-restarts: -1
 
+    - name: Kubernetes namespace report
```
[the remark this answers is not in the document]
deadlycoconuts: How did that report get generated in your example if you've used the `if: ${{ failure() }}` conditional? 😱

infra/charts/turing/subcharts/merlin/templates/_helpers.tpl
```diff
@@ -27,10 +27,6 @@
 {{- printf "%s.%s.svc.cluster.local" (include "merlin.postgresql.name" .) .Release.Namespace -}}
 {{- end -}}
 
-{{- define "merlin.mlflow.backendStoreUri" -}}
-{{- printf "postgresql://%s:%s@%s:5432/%s" .Values.postgresql.postgresqlUsername .Values.postgresql.postgresqlPassword (include "merlin.db.host" .) .Values.postgresql.postgresqlDatabase -}}
-{{- end -}}
```
deadlycoconuts, line 32: Thanks for removing this!

deadlycoconuts (approved):
LGTM! Thanks a lot for updating this! 🚀

.github/actions/setup-test-cluster/action.yaml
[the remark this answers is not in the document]
krithika369: Haha yes, that's right!

PR #309, opened 2023-01-25
Bugfix: Throw error when a router is created with an nop exp engine and a standard ensembler

deadlycoconuts:
## Context 
As of present, Turing Routers created/updated with an `nop` experiment engine and a standard ensembler configured are not blocked by the Turing API server. Such router configurations are invalid because standard ensemblers are not meant to be utilised without an experiment engine. Hence, an appropriate error needs to be thrown whenever a user tries to create/edit routers with such a configuration. 

While the UI currently prevents such a scenario from occurring due to how it manages the router creation/edit workflow, an additional API validation check needs to be introduced so as to prevent such a scenario from occurring if users were to send requests to the API directly or via the SDK. This PR introduces such a check that throws an error if the aforementioned invalid configuration is used to create/update a Turing Router.

## Additional Modifications
- `api/go.mod` - Removal of explicit dependency on `github.com/go-playground/validator` in favour of `github.com/go-playground/validator/v10` (this package reappears as an indirect dependency but this needs to be removed from the `gojek/mlp` package for it to be completely removed)

### DIFF
--- a/api/turing/validation/validator.go
+++ b/api/turing/validation/validator.go
@@ -287,6 +287,19 @@ func validateConditionOrthogonality(
 	}
 }
 
+func ValidateStdEnsemblerNotConfiguredForNopExpEngine(
+	sl validator.StructLevel,
+	ensembler *models.Ensembler,
+	experimentEngine *request.ExperimentEngineConfig,
+) {
+	if ensembler != nil && ensembler.Type == models.EnsemblerStandardType {
+		if experimentEngine.Type == models.ExperimentEngineTypeNop {
+			sl.ReportError(experimentEngine.Type, "Type", "Type",
+				"should not be nop when a standard ensembler is configured", "")
+		}
+	}
+}
+
 func validateRouterConfig(sl validator.StructLevel) {
 	router := sl.Current().Interface().(request.RouterConfig)
 	instance := sl.Validator()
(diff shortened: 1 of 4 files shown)


api/turing/validation/validator.go
```diff
@@ -287,6 +287,19 @@ func validateConditionOrthogonality(
 	}
 }
 
+func ValidateStdEnsemblerNotConfiguredForNopExpEngine(
```
krithika369, line 290: Can we flip the check and the error message around to make sure that if the experiment engine is nop, standard ensembler cannot be selected? This way, it will match the typical router configuration workflow more closely (i.e., fixing the experiment engine as it is chosen first).

krithika369: Also, this function can be private (name can start with lower case letter) as there is no need for it to be called explicitly from another package. Even the tests are using `validate.Struct` to invoke this function.

krithika369 (approved):
Left some small comments. Thanks for this fix!

api/turing/validation/validator.go
deadlycoconuts: Okay, I've flipped the error and the check around now, i.e.:
```go
	if experimentEngine.Type == models.ExperimentEngineTypeNop {
		if ensembler != nil && ensembler.Type == models.EnsemblerStandardType {
			sl.ReportError(ensembler.Type, "Type", "Type",
				"should not be standard if an nop experiment engine is configured", "")
		}
	}
```
Thanks for catching the part on the function being public too!

deadlycoconuts:
Thanks for the quick review! I'll be merging this soon! 🚀

PR #311, opened 2023-01-26
Add plugin monitoring by pinging it async

deadlycoconuts:
## Context
Given the recent changes to [XP](https://github.com/caraml-dev/xp)'s experiment engine plugin to refactor its entire Treatment Service logic into a plugin, it has been observed that there would be an increased likelihood of the plugin failing due to its increased complexity. While this only affects XP's plugin, it has been decided that an experiment engine plugin failure, in general, should be handled differently by the Turing Router.

As of present, a Turing Router merely receives an error message when calling the `GetTreatmentForRequest` method of an experiment engine that has crashed, which conflates with regular error messages returned by the experiment engine plugin even when it is running normally. Hence, a separate way needs to be introduced to determine the health of the plugin, which this PR introduces through the use of `Ping()` in an asynchronous manner periodically.

Upon detecting that the plugin has crashed, the Router will be made to `panic()`. This was a conscious decision to ensure that any Routers with a faulty plugin that keeps crashing is bound to crash, as opposed to making the Router restart the plugin manually (asynchronously or not) which would allow the Router to either continue serving requests while relying on its fallback route or cause the Router's queue proxy to continue receiving requests while the Router attempts to restart and reconnect to the plugin. 

In both cases, there is reduced visibility of any actual fault of the plugin since the Router would repeatedly attempt to restart the plugin without giving any explicit alerts of a faulty plugin.

## Main Modifications
- `api/turing/cluster/servicebuilder/router.go` - Addition of a line to insert `LivenessPeriodSeconds`, the period at which a router should `Ping()` its plugin into a Router's configuration
- `api/turing/config/config.go` - Addition of the `LivenessPeriodSeconds` field to an experiment's plugin configuration
- `engines/router/missionctl/experiment/experiment.go` - Addition of a monitoring method to `Ping()` the experiment engine plugin repeatedly

### DIFF
--- a/infra/charts/turing/templates/_helpers.tpl
+++ b/infra/charts/turing/templates/_helpers.tpl
@@ -202,7 +202,7 @@ RouterDefaults:
 {{ if eq (toString $expEngine.type) "rpc-plugin" }}
       PluginConfig:
         Image: {{ $expEngine.rpcPlugin.image }}
-        LivenessPeriodSeconds: {{ $expEngine.rpcPlugin.livenessPeriodSeconds }}
+        LivenessPeriodSeconds: {{ $expEngine.rpcPlugin.livenessPeriodSeconds | default 10 }}
 {{ end }}
       ServiceAccountKeyFilePath: {{ $expEngine.serviceAccountKeyFilePath }}
 {{- end -}}
--- a/infra/charts/turing/templates/_helpers.tpl
+++ b/infra/charts/turing/templates/_helpers.tpl
@@ -202,6 +202,7 @@ RouterDefaults:
 {{ if eq (toString $expEngine.type) "rpc-plugin" }}
       PluginConfig:
         Image: {{ $expEngine.rpcPlugin.image }}
+        LivenessPeriodSeconds: {{ $expEngine.rpcPlugin.livenessPeriodSeconds }}
 {{ end }}
       ServiceAccountKeyFilePath: {{ $expEngine.serviceAccountKeyFilePath }}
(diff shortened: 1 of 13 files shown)


infra/charts/turing/templates/_helpers.tpl
```diff
@@ -202,6 +202,7 @@ RouterDefaults:
 {{ if eq (toString $expEngine.type) "rpc-plugin" }}
       PluginConfig:
         Image: {{ $expEngine.rpcPlugin.image }}
+        LivenessPeriodSeconds: {{ $expEngine.rpcPlugin.livenessPeriodSeconds }}
```
krithika369, line 205: Shall we set a default value here so that the Turing operator can potentially skip setting this value during deployment (`| default 10`) ?

infra/e2e/turing.values.yaml
```diff
@@ -18,6 +18,7 @@ turing:
     type: rpc-plugin
     rpcPlugin:
       image:  # Value will be set at install time using helm --set command
+      livenessPeriodSeconds: 10
```
krithika369, line 21: If introducing the liveness default at the `infra/charts/turing/templates/_helpers.tpl` level, we can just comment this one out here and note that the value can be overridden here.

krithika369 (approved):
Left some small comments. The PR LGTM. Thanks, @deadlycoconuts!

deadlycoconuts:
Thanks a lot for the quick review! I'll be merging this! 🚀

PR #315, opened 2023-02-06
Bugfix: Fix plugin metric registration

deadlycoconuts:
## Context
When a Turing Router is configured with multiple traffic rules, it uses a fiber config with multiple routes. Each route has its own route strategy, which will initialise its own experimentation policy. Experiment engines that register their metrics on the Turing Router will thus attempt to register the same metrics multiple times on the Turing Router, causing the Router to panic since those metrics would have already been registered before.

This PR introduces a fix to store the list of names of registered plugin metrics in a set, which is then used to verify any subsequent registration of metrics, which happens when the other fiber routes initialise their experimentation policies.

## Additional Changes
- `engines/experiment/examples/plugins/hardcoded/Dockerfile` - Update the `alpine` image used by the hardcoded experiment engine plugin


engines/router/missionctl/instrumentation/prometheus.go
```diff
@@ -32,6 +32,8 @@ var requestLatencyBuckets = []float64{
 	2000, 5000, 10000, 20000, 50000, 100000,
 }
 
+var additionalRegisteredMetricNames = set.New(nil)
```
leonlnj, line 35: dont have full context of the code, would having `registeredMetrics` with `ExperimentEngineRequestMs`, `RouteRequestDurationMs` and `TuringComponentRequestDurationMs` be a safer option, in case these get registered again?

leonlnj:
let a small comment, LGTM

leonlnj (approved):

engines/router/missionctl/instrumentation/prometheus.go
krithika369, line 35: Could you add a comment above `additionalRegisteredMetricNames` why it's needed? Just a one-liner explaining that it's meant to prevent re-registrations if the Register method is called multiple times will be useful.

krithika369 (approved):
Thanks for the fix, @deadlycoconuts!

engines/router/missionctl/instrumentation/prometheus.go
deadlycoconuts: True, that's a good point, but then again those are Router metrics and if any of the plugins attempts to register those metric names (for themselves), I'd prefer the Router to panic to let the plugin user/builder know that they are attempting an illegal operation instead of just pretending nothing's happening.

deadlycoconuts:
Thanks everyone for the quick reviews! Merging this!

PR #317, opened 2023-02-07
Fix marshalling issue in config.Load()

shydefoo:
### Summary
* This PR fixes a marshelling issue when reading in `ClusterConfig.EnsemblingServiceK8sConfig`, which results in some `k8sConfig.Cluster` fields not being read.
* This error was not detected in the e2e tests, as none of the routers created in e2e tests use the pyfunc ensemblers

### Modifications
`config.Load()` has been modified to further unmarshall `EnsemblingServiceK8sConfig` if it is present, using `"sigs.k8s.io/yaml"`.

### DIFF
--- a/api/turing/config/config.go
+++ b/api/turing/config/config.go
@@ -496,14 +496,21 @@ func Load(filepaths ...string) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to unmarshal config values: %s", err)
 	}
+	config, err = loadEnsemblingSvcConfig(config, v.AllSettings())
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load ensemblingservicek8sconfig, err %s", err)
+	}
 
+	return config, nil
+}
+
+func loadEnsemblingSvcConfig(config *Config, v map[string]interface{}) (*Config, error) {
 	// NOTE: This section is added to parse any fields in EnsemblingServiceK8sConfig that does not
 	// have yaml tags.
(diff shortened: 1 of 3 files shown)


api/turing/config/config.go
```diff
@@ -496,6 +497,33 @@ func Load(filepaths ...string) (*Config, error) {
 		return nil, fmt.Errorf("failed to unmarshal config values: %s", err)
 	}
 
+	// NOTE: This section is added to parse any fields in EnsemblingServiceK8sConfig that does not
+	// have yaml tags.
+	// For example `certificate-authority-data` is not unmarshalled
+	// by vipers unmarshal method.
```
deadlycoconuts, line 503: Since this block of code is performing a specific operation on a specific part of the `config` object (while the rest of the code in `Load` acts directly on `config` itself), would it be better to refactor this block into a separate helper function? 😅

Something like: 
```go
func Load(filepaths ...string) (*Config, error) {
	// ...
	config.ClusterConfig = loadClusterConfigs(v.AllSettings())
	return config, nil
}
```
It would also help make it easier to interpret why we're doing `return config, nil` prematurely whenever the `if !ok` statements return true.

deadlycoconuts (approved):
Thanks a lot for digging into this! 🚀 Just left one comment but it's a pretty minor one!

api/turing/config/config.go
shydefoo: Yep that makes sense, i'll make the change

PR #318, opened 2023-02-08
Add tzdata dependency to Turing Router

deadlycoconuts:
## Context
This PR introduces the `tzdata` installation within the Turing Router image so as to allow it and other experiment engine plugins to utilise timezone related tools.

### DIFF
--- a/engines/router/Dockerfile
+++ b/engines/router/Dockerfile
@@ -35,6 +35,7 @@ FROM alpine:latest
 ARG BIN_NAME=turing-router
 ENV BIN_NAME ${BIN_NAME}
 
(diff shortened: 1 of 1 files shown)


engines/router/Dockerfile
```diff
@@ -35,6 +35,7 @@ FROM alpine:latest
 ARG BIN_NAME=turing-router
 ENV BIN_NAME ${BIN_NAME}
 
+RUN apk update && apk add tzdata
```
krithika369, line 38: A comment above this line will be useful - to explain that this dependency is not required by the router, but it's a basic dependency for the experiment engine.

krithika369 (approved):
LGTM. Thanks for the fix!

PR #319, opened 2023-02-16
add validation and new upi logger type

leonlnj:
# Description
This PR aims at
- Adding some validation to UPI routers for invalid component configuration
- Bugfix existing experiment response that doesn't include experiment and treatment name defined in [runner interface](https://github.com/caraml-dev/turing/blob/main/engines/experiment/runner/runner.go#L15)
- New `ResultLogger=UPI` and removed `ConsoleLogger` from API schema 
  - Adding RouterLog publishing to UPI router (existing implementation to BQ/Kafka is removed)
    - Currently UPI Router will only publish to Kafka

# Task
- [x] API Server
- [x] Router
- [x] SDK
- UI is not scoped in this PR

# Key Modification

## API Schema
`api/api/specs/routers.yaml` - added UPI logger type and removed console

## Validation
`api/turing/validation/validator.go` - Standard ensembler and UPI logger validation for UPI Router

## API Server
`api/turing/cluster/servicebuilder/router.go` - construction of env var for Router. Logic to which kafka topic is predefined here.

## Patching Experiment response
`engines/router/missionctl/experiment/experiment.go` - where the experiment and treatment name was never set before

## Patching e2e
`api/Makefile` - fix cli to include env creds
`api/config-dev-exp-engine.yaml` - patch config to incorporate recent plugin changes + fix treatment config
`api/e2e/test/config-local.yaml` - add description of expected payload for cred config
`api/e2e/test/router_test.go` - patching to fit the correct treatment name and route name

## UPI Logger
`engines/router/missionctl/log/resultlog/upi_result_log.go` - new `UPIResultLogger` struct and `UPILogger` interface. Kakfa and nop logger added new method to satisfy `UPILogger` interface. This is also the main logic on how `RouterLog` is constructed and the test validate that the fields are set in expected manner.  

`engines/router/missionctl/log/resultlog/kafka.go` - other than adding new method for `UPILogger` interface, existing method is moved to `writeToKafka` which takes in a generic proto, how the message is published remain unchanged. Both `write` (for http) and `WriteUPIRouterLog` (for upi) call the same underlying method.  

`engines/router/missionctl/log/resultlog/resultlog.go` - one of the key changes is to redefined `TuringResultLogEntry` (its weird how it was defined to be `turing.TuringResultLogMessage` and then added pointer receiver). The proto is added as field and passed as it is when required.
```
type TuringResultLogEntry turing.TuringResultLogMessage
type TuringResultLogEntry struct {
	resultLogMessage turing.TuringResultLogMessage
}
```
The previous defined, will not recognize the type as a proto, probably due to [type tracing of custom-type (Syntax: Type Definitions and Concept: Underlying Types)](https://go101.org/article/type-system-overview.html)

The original proto can assigned to a proto.message but not `TuringResultLogEntry` 
Screenshot 2023-02-23 at 4 07 40 PM

Because of this, the type was never recognized as a proto, and to accomodate this change, `json.marshal` is modified to `protojson` marshaller. The output are intact (as per unit test), only the marshaller is required to change to produce the same output.

`engines/router/missionctl/server/application.go`- Construct the logger and pass it to `UPILogger`. 

`engines/router/missionctl/server/upi/server.go` - Note that `UPILogger` has its own `ResultLogger` and does not refer to the global module as per `HTTP`. The logger is injected and the resultlogger is part of the `missionControl`, this structural change is to allow dependency injection of logger implementation (no need for monkey patch) and it allows the removal of `logUtils` and global reference of `ResultLog` module (currently depending on it without being explicitly anywhere). Result logging is using the ResultLogger injected now. Added implementation to get experiment response using fiber context

`engines/router/missionctl/server/upi/logutils.go` - deleted and beloing to part of `engines/router/missionctl/log/resultlog/upi_result_log.go`, in the rightful log module.

## SDK
Validation and UPI Logger init


api/config-dev-exp-engine.yaml
```diff
@@ -19,15 +21,15 @@ Experiment:
...
         variants:
           - name: control
-          - name: treatment-1
+          - name: treatment-a
         variants_configuration:
           control:
             traffic: 0.85
             treatment_configuration:
-              foo: bar
-              route_name: treatment-a
```
leonlnj, line 28: previously the treatment is `control` but it return a `route_name` of `treatment-a`, vice versa with treatment-1.

Minor refactoring to fix that and simplify the config

api/e2e/test/routers_suite_test.go
```diff
@@ -34,6 +34,7 @@ var cfg config.Config
 var defaultDeploymentIntervals = []interface{}{"10m", "5s"}
 var defaultDeletionIntervals = []interface{}{"20s", "2s"}
 var arbitraryUpdateIntervals = []interface{}{"10s", "1s"}
+var istioVirtualServiceIntervals = []interface{}{"60s", "5s"}
```
leonlnj, line 37: increased and shifted, as 30s is too short for local setup and 60s works better

api/e2e/test/router_test.go
```diff
@@ -31,24 +31,24 @@ var _ = DeployedRouterContext("testdata/create_router_std_ensembler_proprietary_
...
 
 					When("called with {client: {id: 4}} in the payload", func() {
-						It("responds with data from the `treatment-a` route", func() {
+						It("responds with data from the `control` route", func() {
 							routerE.POST("/v1/predict").
 								WithHeaders(defaultPredictHeaders).
 								WithJSON(json.RawMessage(`{"client": {"id": 4}}`)).
 								Expect().
 								Status(http.StatusOK).
-								JSON().Equal(json.RawMessage(`{"version": "treatment-a"}`))
```
leonlnj, line 40: this was actually flawed previously. "4" calls treatment "control" which has `route_name=treatment-a`

Since the exp config is patched to pair route and treatment consistently, `id=4` will call control and return `control`

api/turing/validation/validator_test.go
```diff
@@ -22,6 +23,15 @@ import (
 	routerConfig "github.com/caraml-dev/turing/engines/router/missionctl/config"
 )
 
+var dummyConfig json.RawMessage
+
+func getDefaultValidator() (*validator.Validate, error) {
```
leonlnj, line 28: extracting common mock construction for re-use across whole test

api/turing/validation/validator_test.go
```diff
@@ -287,15 +301,22 @@ type routerConfigTestCase struct {
 	trafficRules       models.TrafficRules
 	autoscalingPolicy  *models.AutoscalingPolicy
 	expectedError      string
+	logConfig          *request.LogConfig
 }
 
-func (tt routerConfigTestCase) RouterConfig(protocol routerConfig.Protocol) *request.RouterConfig {
```
leonlnj, line 292: no idea why protocol is in method parameter (written by me 😅  previously i think), removed and refer to input args instead for consistency

engines/router/missionctl/config/config.go
```diff
@@ -36,6 +36,8 @@ const (
 	ConsoleLogger ResultLogger = "CONSOLE"
```
leonlnj, line 36: note: console logger is only removed from API, the implementation isn't remove as it can be potentially useful for debugging

engines/router/missionctl/experiment/experiment.go
```diff
@@ -92,6 +94,8 @@ func NewResponse(expPlan *runner.Treatment, expPlanErr error) *Response {
 		experimentResponse.Error = expPlanErr.Error()
 	} else {
 		experimentResponse.Configuration = expPlan.Config
```
leonlnj, line 96: treatment and experiment name was never returned prior

engines/router/missionctl/server/upi/server.go
```diff
@@ -38,17 +37,18 @@ type Server struct {
...
+func NewUPIServer(mc missionctl.MissionControlUPI, rl *resultlog.UPIResultLogger) *Server {
 	return &Server{
 		missionControl: mc,
+		resultLogger:   rl,
 	}
 }
 
 func (us *Server) Run(listener net.Listener) {
 	s := grpc.NewServer()
-	//TODO: the unmarshalling can be done more efficiently by using partial deserialization
```
leonlnj, line 51: initially optimization could had been done to speed up calls, however it would be required for logging purposes, hence there is no more need to explore this

sdk/e2e/09_deploy_router_with_std_ensembler_test.py
```diff
@@ -116,5 +116,5 @@ def test_deploy_router_with_std_ensembler():
         json={"client": {"id": 4}},
     )
     assert response.status_code == 200
-    expected_response = {"version": "treatment-a"}
+    expected_response = {"version": "control"}
```
leonlnj, line 119: patching due to correction of the experiment engine config

sdk/e2e/11_deploy_router_upi_traffic_split_test.py
```diff
@@ -153,7 +155,7 @@ def test_deploy_router_upi_traffic_split():
             variable_pb2.Variable(
                 name="client_id",
                 type=type_pb2.TYPE_STRING,
-                string_value="1",
```
leonlnj, line 156: although named `treatment` this was actually `control`, if traffic rules malfunction and a control via default route, would had pass this anyway

sdk/e2e/11_deploy_router_upi_traffic_split_test.py
```diff
@@ -64,7 +64,7 @@ def test_deploy_router_upi_traffic_split():
                     field_source=FieldSource.PREDICTION_CONTEXT,
                     field="client_id",
                     operator="in",
-                    values=["1"],
+                    values=["7"],
```
leonlnj, line 67: 1 would return control, which is the default route and doesn't test traffic rules

sdk/e2e/11_deploy_router_upi_traffic_split_test.py
```diff
@@ -175,7 +177,7 @@ def test_deploy_router_upi_traffic_split():
             variable_pb2.Variable(
                 name="client_id",
                 type=type_pb2.TYPE_STRING,
-                string_value="1234",
+                string_value="12",
```
leonlnj, line 180: both 7 and 12 returns treatment. Hence how traffic rules is tested, is that two request both should get treatment, but the one not fulfilling traffic rule will get control, which is `12`.  `1234` previously returns treatment which calls control route

sdk/e2e/12_deploy_router_upi_std_ensembler_test.py
```diff
@@ -126,8 +126,9 @@ def test_deploy_router_upi_std_ensembler():
     channel = grpc.insecure_channel(retrieved_router.endpoint)
     stub = upi_pb2_grpc.UniversalPredictionServiceStub(channel)
 
-    logging.info("send request that satisfy treatment-a")
+    logging.info("send request that satisfy control")
```
leonlnj, line 129: other than the traffic rules splt which has some flaw and the others are merely adapting to the new corrected config

sdk/turing/router/config/router_version.py
```diff
@@ -47,8 +47,9 @@ def __init__(
         self.status = RouterStatus(status)
         self.name = name
         self.monitoring_url = monitoring_url
-        self.log_config = RouterVersionLogConfig(**kwargs.get("log_config"))
```
leonlnj, line 50: because `LogConfig` validates against the `Protocol`, the `RouterConfig` has to be initiated first

engines/router/missionctl/log/resultlog/kafka.go
```diff
@@ -125,33 +130,50 @@ func (l *KafkaLogger) write(turLogEntry *TuringResultLogEntry) error {
...
 
+func (l *KafkaLogger) write(turLogEntry *TuringResultLogEntry) error {
+	return l.writeToKafka(
+		&turLogEntry.resultLogMessage,
+		turLogEntry.resultLogMessage.TuringReqId,
+		turLogEntry.resultLogMessage.EventTimestamp,
+	)
+}
+
+func (l *KafkaLogger) WriteUPIRouterLog(routerLog *upiv1.RouterLog) error {
```
pradithya, line 141: It feels a bit out of place. Can we instead have a separate `KafkaLogger` (e.g. `UPIKafkaLogger`) and embed the original `KafkaLogger` within it and override the `write` method?

pradithya (changes requested):
Thanks @leonlnj I left some comments below

engines/router/missionctl/log/resultlog/upi_result_log.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+		RouterOutput: &upiv1.RouterOutput{
+			PredictionContext: upiResp.GetPredictionContext(),
+			Headers:           convertToUPIHeader(grpcResp.Header),
+		},
+		RequestTimestamp:   upiReq.GetMetadata().GetRequestTimestamp(),
+		TableSchemaVersion: convertorTableSchema,
+	}
+
+	if grpcResp.Err != "" {
```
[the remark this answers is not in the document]
leonlnj: will change to `if grpcResp.ErrCode != int(codes.OK) `

sdk/e2e/12_deploy_router_upi_std_ensembler_test.py
```diff
@@ -168,4 +169,4 @@ def test_deploy_router_upi_std_ensembler():
     logging.info(f"received response {response}")
 
     assert response.prediction_result_table == request.prediction_table
-    assert response.metadata.models[0].name == "control"
```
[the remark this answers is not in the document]
leonlnj: this is not possible with the current proprietary engine, it doesn't return those fields.

engines/router/missionctl/log/resultlog/kafka.go
leonlnj: hmm the challenge with this, is that a UPI wrapper for each logger needs to be created, including the `nop` logger, since write is redeclared twice as the parameter is different. I can change the interface to a proto.message but all the loggers for HTTP need some refactor and type assert back to `TuringResultLog`. I prefer to do this is another MR if necessary. 

```
// TuringResultLogger is an abstraction for the underlying result logger
type TuringResultLogger interface {
	write(*TuringResultLogEntry) error
}
```

pradithya (approved):
LGTM!

engines/router/missionctl/log/resultlog/kafka.go
krithika369: > is that a UPI wrapper for each logger needs to be created, including the nop logger

I don't think this is needed. Since the logging orchestrator (UPIResultLogger) in itself in a separate implementation, the `logger` in the below initialization need not adhere to the `TuringResultLogger` interface anymore, right? That logger can be the `UPIKafkaLogger` which implements only the `UPILogger` interface, but makes use of the existing Kafka logger under the hood to write the given messge.

```
resultLogger, err := resultlog.InitUPIResultLogger(cfg.AppConfig.Name, logger)
```

krithika369 (approved):
LGTM. Thanks, @leonlnj ! Left some small comments.

api/turing/config/config.go
```diff
@@ -340,6 +342,12 @@ type KafkaConfig struct {
 	CompressionType string
 }
 
+// UPIConfig captures the defaults used by UPI Router
+type UPIConfig struct {
+	// KafkaBrokers is broker which all Router will write to when UPI logging is enabled
+	KafkaBrokers string
```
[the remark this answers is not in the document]
leonlnj: Hmm I saw `kafkaConfig` which contain `MaxMessageBytes` and `CompressionType` as defaults applied to all Kafka, the broker is applied specifically for UPI Routers only, so I didnt reuse it. 

> Are we expecting this UPIConfig to hold more values eventually? 
Yes, its one of possible feature to enable finer-grain logging control, so we can reduce the amount of the logs if user doesn't want them.

engines/router/missionctl/log/resultlog/kafka.go
leonlnj: yep `UPIKafkaLogger` is fine, but I would need `UPINopLogger` else there be this clash. Will just add it anyway 😅 

```
// write is a nop method that satisfies the NopLogger interface
func (*NopLogger) write(turLogEntry *TuringResultLogEntry) error {
	return nil
}

// write is a nop method that satisfies the UPILogger interface
func (*NopLogger) write(routerLog *upiv1.RouterLog) error {
	return nil
}
```

PR #321, opened 2023-03-02
UI RouterLogs for UPI Router

leonlnj:


ui/src/router/components/form/components/router_config/GeneralSettingsPanel.js
```diff
@@ -52,6 +52,7 @@ export const GeneralSettingsPanel = ({
             value === "UPI_V1"
               ? "/caraml.upi.v1.UniversalPredictionService/PredictValues"
               : "",
+          endpoint: "",
```
leonlnj, line 55: reset endpoint as well when protocol cause, since the endpoints are not re-usable for both protocol

ui/src/router/components/form/validation/schema.js
```diff
@@ -454,13 +454,19 @@ const schema = (maxAllowedReplica) => [
...
     config: yup.object().shape({
       log_config: yup.object().shape({
-        result_logger_type: yup
-          .mixed()
-          .required("Valid Results Logging type should be selected")
-          .oneOf(
-            ["nop", "bigquery", "kafka"],
-            "Valid Results Logging type should be selected"
-          ),
+        result_logger_type: yup.string().when("$protocol", {
```
leonlnj, line 457: actually i think this entire chunk (and the old) is not required, as the UI is a dropdown (HTTP) or switch (UPI), other value of inputs will not be passed

leonlnj:
closing this as we will put UI on hold

PR #323, opened 2023-03-09
Enable TuringResultLog on upi router and refactor global ResultLog

leonlnj:
## Description
- Remove experiment and treatment name from experiment response in json that was introduced in [previous MR](https://github.com/caraml-dev/turing/pull/319) to keep interface as per `v1.10.0`. UPI Response will continue to be populated with experiment and treatment name

- Refactor global `ResultLog` to be injectable to server

- Add `TuringResultLogMessage` publishing to UPI Router as a option

note* - Leaving fiber as v2.1.0-rc2 for now till later date.

## Modification
Persisting UPI kafka information in `RouterVersion`
- `api/turing/api/request/request.go` - Add kafka config for UPI logger while building router version, so that it is persisted 
- `api/turing/cluster/servicebuilder/router.go` - the formatting logic is moved to `request.go`, env var will be created as per Kafka loggers

Supporting `TuringResultLogMessage` 
- `api/turing/validation/validator.go` - remove validation for UPI logger, as all type can be supported

Removing experiment/treatment name from experiment response
- `engines/router/missionctl/experiment/experiment.go` - omit from json

Deletion of `upi_nop_logger`
- Reusing `nop_logger` of http for the same

Refactoring of ResultLogger
At a high level, there are `ResultLogger` struct for `TuringResultLogMessage` and `UPILogger` for `RouterLog`. The result logger owns the logic on how the log proto message (TuringResult/RouterLog) is being constructed.  
Each result logger composed of an underlying logger middleware that implement and hold the logic of how the respective proto is written (to bq/kafka etc)

`upi_result_log` and `result_log` are structured to be symmetrical as much as possible and dependencies are injected, which remove the need for monkey patching and allow mock logger to be injected and validate the log creation logic.

- `engines/router/missionctl/log/resultlog/resultlog.go` 
  - Removal of global logger
  - New ResultLogger struct
  - Removing of `TuringResultLogEntry`
  - Merged `logutils.go` into this file and make the method part of result logger, as per upi logger
  - `InitTuringResultLogger` does not create the underlying middleware anymore, the application owns the logic of what is the middleware to inject

- `engines/router/missionctl/log/resultlog/upi_result_log.go`
  - new result logger and logger type field - base on the logger type, the result logger will determine which underlying logger to use
  - regardless of the chosen logger type, the server will call `LogTuringRouterRequestSummary` of the result logger, passing in the UPI req/res and be unaware which log proto is used nor how is it produced.

## Testing
Verified bq logging from v1.10 (left) that response format are still intact after refactoring
image

BQ testing for http experiment response
image


engines/router/missionctl/log/resultlog/kafka_test.go
```diff
@@ -165,27 +163,14 @@ func TestKafkaLoggerWrite(t *testing.T) {
 		topic:               "test-topic",
 		producer:            mp,
 	}
-	testKafkaLogEntry := []byte(`{"key": "value"}`)
-	turingResLogEntry := &TuringResultLogEntry{
-		resultLogMessage: turing.TuringResultLogMessage{},
-	}
-
-	// Patch newKafkaLogEntry
-	monkey.Patch(
```
leonlnj, line 174: no reason to patch `newJSONKafkaLogEntry`, can just use its existing implementation and the assertion below will validate it

engines/router/missionctl/log/resultlog/resultlog.go
```diff
@@ -42,131 +54,141 @@ var ResultLogKeys = struct {
...
+func (rl *ResultLogger) LogTuringRouterRequestSummary(
+	predictionID string,
+	logger log.Logger,
+	timestamp time.Time,
+	reqHeader http.Header,
+	reqBody []byte,
+	mcRespCh <-chan RouterResponse) {
 
-// Value returns the TuringResultLogEntry in a loggable format
-func (logEntry *TuringResultLogEntry) Value() (map[string]interface{}, error) {
```
leonlnj, line 56: this is only used by console logger

engines/router/missionctl/log/resultlog/resultlog.go
```diff
@@ -42,131 +54,141 @@ var ResultLogKeys = struct {
...
+	ch <- RouterResponse{
+		key:    key,
+		header: r.Header(),
+		body:   data,
 	}
 }
 
-// NewTuringResultLogEntry returns a new TuringResultLogEntry object with the given context
+func (rl *ResultLogger) logEntry(log *turing.TuringResultLogMessage) error {
+	log.RouterVersion = rl.appName
```
leonlnj, line 159: routerversion is being appended at this layer as the resultlogger owns the logic on how it should be populated and not the server.

krithika369:
@leonlnj I am yet to review this PR but will do it today. Thanks.

engines/router/missionctl/log/resultlog/upi_result_log.go
```diff
@@ -48,36 +53,139 @@ var convertorTableSchema uint32 = converter.TableSchemaV1
...
+
+func (ul *UPIResultLogger) logEntry(log *upiv1.RouterLog) error {
+	return ul.upiLogger.write(log)
+}
+
+// LogTuringRouterRequestError logs the given turing request id and the error data
+func (ul *UPIResultLogger) LogTuringRouterRequestError(ctx context.Context, err *errors.TuringError) {
+	logger := log.WithContext(ctx)
+	defer func() {
+		_ = logger.Sync()
```
pradithya, line 118: What's the purpose of calling logger.Sync() here?

engines/router/missionctl/log/resultlog/bigquery.go
```diff
@@ -26,17 +26,16 @@ import (
...
 type bqLogEntry struct {
-	*TuringResultLogEntry
+	resultLog *turing.TuringResultLogMessage
 }
 
 // Save implements the ValueSaver interface on bqLogEntry, for saving the data to BigQuery
 func (e *bqLogEntry) Save() (map[string]bigquery.Value, string, error) {
 	var kvPairs map[string]bigquery.Value
-	bytes, err := json.Marshal(e)
+	bytes, err := protoJSONMarshaller.Marshal(e.resultLog)
```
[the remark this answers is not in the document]
pradithya: It doesn't seems so, but just to confirm.

engines/router/missionctl/log/resultlog/upi_result_log.go
```diff
@@ -48,36 +53,139 @@ var convertorTableSchema uint32 = converter.TableSchemaV1
...
-	resultLogger.routerName = routerNameWithVersion[:i]
+	upiResultLogger.routerName = routerNameWithVersion[:i]
 	// do not include '-'
-	resultLogger.routerVersion = routerNameWithVersion[i+1:]
+	upiResultLogger.routerVersion = routerNameWithVersion[i+1:]
 
-	return resultLogger, nil
-}
+	// check the respective logger has been provided for the config
+	switch loggerType {
```
pradithya, line 80: nit: I am thinking if it's necessary to check again, `InitUPIResultLogger` is called internally by Turing so all its arguments will likely to be valid, do I miss anything?

pradithya (commented):
Thanks! It's a neat refactoring. I have few comments below to clarify some of my doubts.

engines/router/missionctl/log/resultlog/bigquery.go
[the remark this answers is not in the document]
leonlnj: no, it was needed to change because its directly marshalling on the proto (there is some difference if json marshal is used on the proto)

engines/router/missionctl/log/resultlog/console.go
```diff
@@ -1,24 +1,37 @@
...
-	// Get context-specific logger
+func (*ConsoleLogger) write(turLogEntry *turing.TuringResultLogMessage) error {
+	// Get context-specific upiLogger
 	logger := log.Glob()
 
-	// Get the loggable data
-	kvPairs, err := turLogEntry.Value()
+	var kvPairs map[string]interface{}
+	// Marshal into bytes
+	bytes, err := protoJSONMarshaller.Marshal(turLogEntry)
```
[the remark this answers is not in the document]
leonlnj: its done this way to put the proto into a `map[string]interface{}` from proto i believe. 

we can do `.String()`

engines/router/missionctl/log/resultlog/upi_result_log.go
leonlnj: my understanding is to flush any buffered log entries, im not sure if we actually need it, but its all over the codebase to do a sync for many logging

engines/router/missionctl/log/resultlog/kafka_test.go
```diff
@@ -165,27 +163,14 @@ func TestKafkaLoggerWrite(t *testing.T) {
...
-			assert.Equal(t, &turingResLogEntry.resultLogMessage, message)
-			return testKafkaLogEntry, nil
-		},
-	)
-	defer monkey.UnpatchAll()
+	testKafkaLogEntry := []byte(`{"turing_req_id":"123"}`)
+	msg := &turing.TuringResultLogMessage{TuringReqId: "123"}
 
 	// Set up Produce
 	mp.On("Produce", mock.Anything, mock.Anything).Return(nil)
```
tiopramayudi, line 170: instead of `mock.Anything` I think you can use the expected `kafka.Message` in the parameter

engines/router/missionctl/log/resultlog/upi_result_log.go
leonlnj: hmm yes it will actually be valid for all cases and redudant, I can remove it, no opinion on this. 

I was thinking more of a package pov, when a client want to init this result logger, these checks are necessary to ensure valid configuration to make the rest of the methods work and should not assume on the client (application.go in this case) to pass in correct config

engines/router/missionctl/log/resultlog/upi_result_log.go
```diff
@@ -142,49 +250,11 @@ func (ul *UPIResultLogger) LogTuringRouterRequestSummary(
 
 	// Log the responses. If an error occurs in logging the result to the
 	// configured result log destination, log the error.
-	if err := ul.LogEntry(routerLog); err != nil {
+	if err := ul.logEntry(routerLog); err != nil {
 		log.Glob().Errorf("Result Logging Error: %s", err.Error())
```
tiopramayudi, line 254: should we log the `routerLog` so we know which message that needs to manually retry?

engines/router/missionctl/log/resultlog/kafka.go
```diff
@@ -129,11 +129,11 @@ func (l *KafkaLogger) writeToKafka(
 	return nil
 }
 
-func (l *KafkaLogger) write(turLogEntry *TuringResultLogEntry) error {
+func (l *KafkaLogger) write(turLogEntry *turing.TuringResultLogMessage) error {
 	return l.writeToKafka(
```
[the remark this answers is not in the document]
leonlnj: it make sense if its just kafka alone, but for other logger (bq) there isn't a key so changing the `TuringResultLogger` interface to cater to this is a problem.

alternatively the key generation could be moved to this method from `writeToKafka` but I dont have any opinion on this, since it could be also viewed that `writeToKafka` control how the key is generated from Turing across all implementation.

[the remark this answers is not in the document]
tiopramayudi: I mean to change signature of `writeToKafka` itself since it is only implement kafka, I thought just in case key for `routerLog` and `resultLog` is not the same, you have to change `writeToKafka` in both places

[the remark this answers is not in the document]
leonlnj: hmm its a neat idea but i think that can be done when the key is different, if not it will just be duplicating the same logic in both logger. `writeToKafka` can maintain a consistent way of creating the key till there is a requirement for it to be different

engines/router/missionctl/log/resultlog/console.go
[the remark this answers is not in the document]
pradithya: Let's keep it this way, I just see the original code is like that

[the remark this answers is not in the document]
krithika369: Yes, the zap logger API `Infow` works with KV pairs. No one is using the console logger in prod now, but this format may be preferable to `String()` if we end up enabling it at some point (eg: forward console logs to NewRelic).

api/turing/validation/validator.go
```diff
@@ -400,11 +400,6 @@ func validateUPIRouter(
 		sl.ReportError(router.Ensembler.Type, "Ensembler.Type", "Type",
 			"only standard ensembler is supported for UPI", "")
 	}
-	if !(router.LogConfig.ResultLoggerType == models.NopLogger || router.LogConfig.ResultLoggerType == models.UPILogger) {
```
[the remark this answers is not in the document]
leonlnj: Yes.

Kafka logger (via ResultLogger) uses `TuringResultLogMessage` proto for logging. UPI logger will use `RouterLog`

engines/router/missionctl/log/resultlog/upi_result_log.go
krithika369: Per the Zap logger [docs](https://pkg.go.dev/go.uber.org/zap#Logger.Sync),

> Applications should take care to call Sync before exiting

We have that here: https://github.com/caraml-dev/turing/blob/main/engines/router/missionctl/server/application.go#L38

I think we can remove the other instances of `.Sync()` where we are logging errors (unless there are test cases that are validating that which will require an immediate flush, which doesn't seem to be the case).

krithika369 (approved):
Thanks for the extensive refactoring of the result logging, @leonlnj! LGTM.

krithika369:
Btw, I have one related question, when / how is the `UPILogger` option chosen? As I understand, it only supports the Kafka Result logger now. But at the moment, is the UI exposing any way for the user to choose the RouterLog vs the usual ResultLog?

leonlnj:
If the `APP_RESULT_LOGGER=upi`. Currently its not enabled on the UI and won't be.

PR #327, opened 2023-03-28
Update reference for Merlin dependency

terryyylim:

As a follow-up for moving Merlin repository from `gojek` to `caraml-dev` organisation (https://github.com/caraml-dev/merlin/pull/371), its dependency should be updated for Turing as well.

In addition, this PR upgrades all Turing codebase's modules Go versions to `1.20`.

`knative.dev/pkg` needs to be pinned via the replace directive, to avoid it being upgraded to `v0.0.0-20230320014357-4c84b1b51ee8` which requires `github.com/google/gnostic v0.5.7-v3refs` (https://github.com/knative/pkg/blob/main/go.mod#L78) that will cause 2 dependencies registering the same .proto file twice, and causing the Turing binary to panic on start:
```
2023-04-03T06:35:40.6071783Z panic: proto: file "extensions/extension.proto" is already registered
2023-04-03T06:35:40.6072468Z 	previously from: "github.com/googleapis/gnostic/extensions"
2023-04-03T06:35:40.6073196Z 	currently from:  "github.com/google/gnostic/extensions"
2023-04-03T06:35:40.6073911Z See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict 
```

It should also be noted that the Merlin dependency is pointing to this branch - https://github.com/caraml-dev/merlin/commits/update-import-refs, since the latest Merlin version's dependency requires `knative.dev/pkg` to be updated and will result in the above conflict.


leonlnj (approved):
I was wondering how come its `+185 −372`, turns out to be go.sum. LGTM

api/go.mod
```diff
@@ -207,15 +208,15 @@ replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.4.0+incompatible
 	github.com/Sirupsen/logrus => github.com/sirupsen/logrus v1.9.0
 
+	github.com/caraml-dev/merlin => github.com/caraml-dev/merlin/api v0.0.0-20230403061454-b55c4bd665c0
```
krithika369, line 211: Where is this commit? We should add it to a branch in the Merlin repo if it isn't already there. Could you also update the PR description with the details of the Merlin commit being referenced?

Screenshot 2023-04-05 at 8 05 05 AM

api/go.mod
```diff
@@ -244,4 +245,5 @@ replace (
 	k8s.io/mount-utils => k8s.io/mount-utils v0.22.7
 	k8s.io/pod-security-admission => k8s.io/pod-security-admission v0.22.7
 	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.22.7
+	knative.dev/pkg => knative.dev/pkg v0.0.0-20220222221138-929d328ad73c
```
krithika369, line 248: Why is this added?

krithika369 (approved):
LGTM. Thanks for bumping the Go version and upgrading the CI actions. Left some questions/comments.

api/go.mod
krithika369: Oops, I misread the PR description - this one is the fix.

api/go.mod
terryyylim: My bad, forgot to mention that it is based off this branch - https://github.com/caraml-dev/merlin/commits/update-import-refs. The commit doesnt match the latest commit in that branch because I've been force-pushing so the commit changed.

PR #328, opened 2023-03-28
Update Google client initialisation to use MLP helper method

deadlycoconuts:
## Context
With the introduction of a helper method in MLP (https://github.com/gojek/mlp/pull/78) to abstract away the creation of a Google client to append ID tokens to the headers of outgoing requests, that is authenticated with service or user accounts, this PR imports the said helper method and uses it to initialise its Google client.

This ensures that all outgoing requests from the Turing API server contain ID tokens in their headers.


krithika369 (approved):

PR #331, opened 2023-04-12
Bugfix: Fix pyfunc ensembler undeployment bug

deadlycoconuts:
## Context
After the Gorm library has been updated to v2 in #261, there is a bug discovered in how we're using the Gorm methods to save router versions to the DB whereby nested fields (such as the ensembler configs) do not get updated in their respective tables. This has caused deployed pyfunc ensemblers to not have their corresponding docker configs updated in the `ensembler_configs` table. Since these are left as `NULL`, what then happens downstream is that the cluster controller fails to build KnativeService objects that correspond to the deployed ensembler instances, causing it to be unable to deploy ensemblers that have been deployed (either as part of a router update operation or an undeployment operation).

~~This PR introduces a config to the Gorm DB layer which adds a new field `FullSaveAssociations: true` to the existing Gorm DB configs, which would ensure that all tables get updated whenever Gorm is used to update a router version (or any other object for that matter of fact). See https://github.com/go-gorm/gorm/issues/6162 for more details.~~

This PR introduces additional steps to update the ensembler and enricher configs of a router version (these are stored in their respective tables) and were previously not getting updated whenever a router version is updated:
```go
err = tx.Save(routerVersion).Error
if err != nil {
	return nil, err
}

if routerVersion.Enricher != nil {
	err = tx.Save(routerVersion.Enricher).Error
	if err != nil {
		return nil, err
	}
}
if routerVersion.Ensembler != nil {
	err = tx.Save(routerVersion.Ensembler).Error
	if err != nil {
		return nil, err
	}
}
```

## Totally Unrelated Changes - Unused Arguments Clean Up
The linter's raising a lot of warnings about unused arguments so I've decided to remove them completely wherever possible, which is the case for the methods like `NewRouterEndpoint`, `NewEnricherService` and `NewEnsemblerService` since the existing implementations of the the `clusterSvcBuilder` don't even use them. And since these methods are concrete implementations of the interface `ClusterServiceBuilder`'s methods, I've also removed the arguments from them too, which then requires some minor refactoring of the methods of `mockClusterServiceBuilder` (which also implements the interface) and the tests where it is used.

Most notably, I've removed all instances of the following:
```go
			Labels: map[string]string{
				"env": envType,
			},
``` 
that's specified in the expected and mocked responses (in `api/turing/service/router_deployment_service_test.go`) since in no way is the actual implementation of the `clusterSvcBuilder` ever using the `envType` argument to set labels in the `KnativeService` struct returned. It's weird to continue mocking responses that set those labels and then configure the expected responses to include those labels.

In all other cases (the majority of them), I've simply renamed the remaining unused arguments to `_` so those should be pretty easy to look through.

### DIFF
--- a/api/e2e/test/router_e2e_test.go
+++ b/api/e2e/test/router_e2e_test.go
@@ -184,7 +184,6 @@ var _ = DeployedRouterContext("testdata/create_router_nop_logger_proprietary_exp
 
 							Eventually(func(g Gomega) {
 								router = api.GetRouter(apiE, routerCtx.ProjectID, routerCtx.ID)
-								// TODO: Why is it pending? - g.Expect(router.Value("status").Raw()).To(Equal(api.Status.Pending))
 								g.Expect(router.Value("status").Raw()).To(Equal(api.Status.Undeployed))
 							}, arbitraryUpdateIntervals...).Should(Succeed())
 						})
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -206,7 +206,7 @@ func (ds *deploymentService) DeployRouterVersion(
 
 	// Deploy or update the virtual service
 	eventsCh.Write(models.NewInfoEvent(models.EventStageUpdatingEndpoint, "updating router endpoint"))
-	routerEndpoint, err := ds.svcBuilder.NewRouterEndpoint(routerVersion, project, ds.environmentType, endpoint)
+	routerEndpoint, err := ds.svcBuilder.NewRouterEndpoint(routerVersion, project, endpoint)
 	if err != nil {
 		eventsCh.Write(models.NewErrorEvent(
 			models.EventStageUpdatingEndpoint, "failed to update router endpoint: %s", err.Error()))
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -453,7 +453,6 @@ func (ds *deploymentService) buildEnsemblerServiceImage(
 		Endpoint:          PyFuncEnsemblerServiceEndpoint,
 		Port:              PyFuncEnsemblerServicePort,
 		Env:               routerVersion.Ensembler.PyfuncConfig.Env,
-		ServiceAccount:    "",
 	}
 
 	return nil
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -367,7 +367,6 @@ func (ds *deploymentService) createServices(
 		ensemblerSvc, err := ds.svcBuilder.NewEnsemblerService(
 			routerVersion,
 			project,
-			envType,
 			secretName,
 			knativeQueueProxyResourcePercentage,
 			userContainerLimitRequestFactor,
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -352,7 +352,6 @@ func (ds *deploymentService) createServices(
 		enricherSvc, err := ds.svcBuilder.NewEnricherService(
 			routerVersion,
 			project,
-			envType,
 			secretName,
 			knativeQueueProxyResourcePercentage,
 			userContainerLimitRequestFactor,
--- a/api/turing/service/router_deployment_service_test.go
+++ b/api/turing/service/router_deployment_service_test.go
@@ -36,7 +36,6 @@ type mockClusterServiceBuilder struct {
 func (msb *mockClusterServiceBuilder) NewRouterEndpoint(
 	routerVersion *models.RouterVersion,
 	project *mlp.Project,
-	envType string,
 	versionEndpoint string,
 ) (*cluster.VirtualService, error) {
 	return &cluster.VirtualService{
--- a/api/turing/service/router_deployment_service_test.go
+++ b/api/turing/service/router_deployment_service_test.go
@@ -34,9 +34,9 @@ type mockClusterServiceBuilder struct {
 }
 
 func (msb *mockClusterServiceBuilder) NewRouterEndpoint(
-	routerVersion *models.RouterVersion,
-	project *mlp.Project,
-	versionEndpoint string,
+	_ *models.RouterVersion,
+	_ *mlp.Project,
+	_ string,
 ) (*cluster.VirtualService, error) {
 	return &cluster.VirtualService{
 		Name:      "test-svc-turing-router",
@@ -94,7 +94,7 @@ func (msb *mockClusterServiceBuilder) NewEnricherService(
 func (msb *mockClusterServiceBuilder) NewEnsemblerService(
 	rv *models.RouterVersion,
 	project *mlp.Project,
-	secretName string,
+	_ string,
 	queueProxyResourcePercentage int,
 	userContainerLimitRequestFactor float64,
 ) (*cluster.KnativeService, error) {
@@ -118,7 +118,7 @@ func (msb *mockClusterServiceBuilder) NewRouterService(
 	rv *models.RouterVersion,
 	project *mlp.Project,
 	envType string,
-	secretName string,
+	_ string,
 	expConfig json.RawMessage,
 	routerDefaults *config.RouterDefaults,
 	sentryEnabled bool,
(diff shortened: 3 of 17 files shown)


api/turing/service/router_deployment_service_test.go
```diff
@@ -84,9 +82,6 @@ func (msb *mockClusterServiceBuilder) NewEnricherService(
 		BaseService: &cluster.BaseService{
 			Name:      fmt.Sprintf("%s-enricher-%d", rv.Router.Name, rv.Version),
 			Namespace: project.Name,
-			Labels: map[string]string{
-				"env": envType,
-			},
```
deadlycoconuts, line 89: This should no longer be here the actual `clusterServiceBuilder` does not even use an argument `envType` to set the labels.

api/turing/service/router_deployment_service_test.go
```diff
@@ -271,9 +259,6 @@ func TestDeployEndpoint(t *testing.T) {
 		BaseService: &cluster.BaseService{
 			Name:      fmt.Sprintf("%s-enricher-%d", routerVersion.Router.Name, routerVersion.Version),
 			Namespace: testNamespace,
-			Labels: map[string]string{
-				"env": envType,
-			},
```
deadlycoconuts, line 276: Removed these from the expected responses since we were just 'expecting' a mocked response though in reality the actual method doesn't even actually set the `envType` as the `env` label.

api/turing/service/router_deployment_service.go
```diff
@@ -251,6 +251,15 @@ func (ds *deploymentService) UndeployRouterVersion(
 		return err
 	}
 
+	// Create a new docker config for the ensembler with the newly generated image
+	if routerVersion.Ensembler != nil && routerVersion.Ensembler.Type == models.EnsemblerPyFuncType {
+		// An empty image ref is passed since it is not used by the Kubernetes controller for undeployment
+		routerVersion.Ensembler.DockerConfig = generateDockerConfigFromPyfuncConfig(
```
krithika369, line 257: I'd have expected the `routerVersion.Ensembler.DockerConfig` to be set at this point already, as we are passing the routerVersion reference to the [DeployRouterVersion](https://github.com/caraml-dev/turing/blob/main/api/turing/service/router_deployment_service.go#L110) method (which will alter the docker config after the image is built) and the `UndeployRouterVersion` is called downstream using the same routerVersion ref: https://github.com/caraml-dev/turing/blob/main/api/turing/api/deployment_controller.go#L76

What am I missing?

(Your approach to fix this bug will work too, although it's making the assumption that there will be a docker config attached to the Pyfunc config always, even in cases where the image building may have timed out. Nothing wrong with this but  wondering why we are unable to propagate the docker config down from the deploy step.)

krithika369 (approved):
Thanks for tracking this down, @deadlycoconuts ! Left some questions.

api/turing/service/router_deployment_service.go
```diff
@@ -673,6 +674,22 @@ func updateKnServices(ctx context.Context, services []*cluster.KnativeService,
...
+) *models.EnsemblerDockerConfig {
+	return &models.EnsemblerDockerConfig{
+		Image:             imageRef,
+		ResourceRequest:   pyfuncConfig.ResourceRequest,
+		AutoscalingPolicy: pyfuncConfig.AutoscalingPolicy,
+		Timeout:           pyfuncConfig.Timeout,
+		Endpoint:          PyFuncEnsemblerServiceEndpoint,
+		Port:              PyFuncEnsemblerServicePort,
+		Env:               pyfuncConfig.Env,
+		ServiceAccount:    "",
```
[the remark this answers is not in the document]
deadlycoconuts: Thanks for pointing that out, I've just removed that :)

api/turing/service/router_deployment_service.go
deadlycoconuts: Ah that's true if the `UndeployRouterVersion` method is only called when updating routers (i.e. deploying a new router version and then deleting the old one) via `deployOrRollbackRouter`, but the  `UndeployRouterVersion` method is also used to undeploy routers using the undeploy endpoint via `undeployRouter` [here](https://github.com/caraml-dev/turing/blob/main/api/turing/api/deployment_controller.go#L266). In the second case, the `routerVersion.Ensembler.DockerConfig`(-s) wouldn't have been set since they would've been [retrieved](https://github.com/caraml-dev/turing/blob/main/api/turing/api/deployment_controller.go#L292) from the DB. 

Although that said, it's possible to introduce more logic into the conditional statement above to only re-generate the docker config if it's missing (i.e. when we are only performing an undeployment) and instead use the existing docker config if it exists in the `routerVersion` (i.e. when we are performing a router update).

> although it's making the assumption that there will be a docker config attached to the Pyfunc config always, even in cases where the image building may have timed out

It's not a pretty assumption but I guess it's partly linked to 3 things, 1) that we are rebuilding entire `KnativeService` structs for undeploying Knative services, which is a little redundant since all we need is the Knative service name, 2) that we aren't storing any information in the `routerVersion` struct/data/schema about the build success/failure of its pyfunc ensembler and 3) that we an 'ensembler' Knative service and the methods that correspond to it are ultimately tightly coupled to the presence of a docker config and it isn't generic enough to operate on a pyfunc config. 

So in order to re-use the existing workflow for undeploying the pyfunc ensemblers, we'll need to generate a docker config temporarily, so that we can build an ensembler Knative service out of it, so that the controller's able to undeploy it using the Knative name contained within the `KnativeService` struct. That's like the approach I thought of but maybe you were seeing another approach that I may have missed out 😅

krithika369: > but the UndeployRouterVersion method is also used to undeploy routers using the undeploy endpoint via undeployRouter [here](https://github.com/caraml-dev/turing/blob/main/api/turing/api/deployment_controller.go#L266)

In this scenario, the docker_config should have been saved to the DB (we also display it in the UI for pyfunc ensemblers). 🤔 

There are 3 undeploy scenarios in all:
* Call `/undeploy` endpoint
* Update a router version successfully (then the existing router version is undeployed)
* Update a router version but it's unsuccessful (then, the new router version will be undeployed immediately)

In scenarios 1 and 2, we should have the docker_config in the DB and in scenario 3, it should be in the routerVersion object ref that's passed around. 😕

deadlycoconuts: Okay I think we've found the true bug behind this to be Gorm, so the ultimate fix to this is to update the Gorm configs to allow all nested fields to be updated whenever we perform a `Save` operation on the DB. I've updated the PR description with the new details; thanks for helping to dig into this! :D

krithika369: Thanks for the fix, @deadlycoconuts (and also for updating all the details in the PR description). LGTM!

api/e2e/test/router_e2e_test.go
```diff
@@ -184,7 +184,6 @@ var _ = DeployedRouterContext("testdata/create_router_nop_logger_proprietary_exp
 
 							Eventually(func(g Gomega) {
 								router = api.GetRouter(apiE, routerCtx.ProjectID, routerCtx.ID)
-								// TODO: Why is it pending? - g.Expect(router.Value("status").Raw()).To(Equal(api.Status.Pending))
```
deadlycoconuts, line 187: Removed this as it is no longer needed (I agree that we shouldn't be expecting the `pending` state).

sdk/e2e/01_create_router_test.py
```diff
@@ -77,9 +77,9 @@ def test_create_router():
     ensembler = DockerRouterEnsemblerConfig(
         image=os.getenv("TEST_ECHO_IMAGE"),
         resource_request=ResourceRequest(
-            min_replica=2, max_replica=2, cpu_request="100m", memory_request="256Mi"
+            min_replica=1, max_replica=1, cpu_request="250m", memory_request="256Mi"
         ),
-        endpoint="anything",
+        endpoint="ensemble",
```
deadlycoconuts, line 82: Made these changes to the SDK tests so these configs resemble the test configs here: https://github.com/caraml-dev/turing/blob/main/api/e2e/test/testdata/create_router_nop_logger_proprietary_exp.json.tmpl.

krithika369 (approved):
LGTM. Thanks for sorting this out and all the test case analysis, @deadlycoconuts !

PR #334, opened 2023-04-18
Add safe-to-evict annotation to image builder job

ariefrahmansyah:
The image-building jobs in Turing and Merlin are timing out. After some investigation, we found that one of the root causes is the node pool got scaled down resulting in the image building pods to be rescheduled. This PR adds "cluster-autoscaler.kubernetes.io/safe-to-evict": "false" to avoid the pod get killed and rescheduled.

Related Merlin's PR: https://github.com/caraml-dev/merlin/pull/380

### DIFF
--- a/api/config-dev.yaml
+++ b/api/config-dev.yaml
@@ -19,6 +19,7 @@ BatchEnsemblingConfig:
     BuildNamespace: default
     BuildTimeoutDuration: 20m
     DestinationRegistry: ghcr.io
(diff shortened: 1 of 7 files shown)


krithika369 (approved):
LGTM. Thanks, @ariefrahmansyah ! 🚀

krithika369:
@ariefrahmansyah shall we merge this PR? We can include it for the upcoming release. :)

PR #337, opened 2023-05-17
Update mlp dependency

deadlycoconuts:
## Context
This PR updates the version of MLP uses in order to incorporate this change [here](https://github.com/caraml-dev/mlp/pull/84), which allows Turing to use workload identity-enabled service account to authenticate all its outgoing resources.


krithika369 (approved):
LGTM!

PR #338, opened 2023-05-19
Remove decryption when retrieving secret from MLP

pradithya:
MLP's Secret API removes the application-level encryption when [returning the secret data.](https://github.com/caraml-dev/mlp/pull/85). Moving forward, we'll rely on TLS to secure the secrets while in transit. 
This PR updates the MLP client used by Turing and removes the decryption process when retrieving a secret from MLP. 

- [X] Tested in dev by deploying router which requires secret


infra/charts/turing/templates/_helpers.tpl
```diff
@@ -186,7 +186,7 @@ MLPConfig:
   MLPEncryptionKey: {{ include "turing.mlp.encryption.key" . | quote }}
 {{ if .Values.tags.mlp }}
   MerlinURL: {{ printf "http://%s:8080/v1" (include "turing.merlin.name" .) }}
-  MLPURL: {{ printf "http://%s:8080/v1" (include "turing.mlp.name" .) }}
+  MLPURL: {{ printf "http://%s:8080" (include "turing.mlp.name" .) }}
```
krithika369, line 189: Some docs / sample values need to be updated with the same change:
* https://github.com/caraml-dev/turing/blob/main/README.md?plain=1#L233
* https://github.com/caraml-dev/turing/blob/main/api/.env.development#L30
* https://github.com/caraml-dev/turing/blob/main/api/config-dev.yaml#L91

krithika369 (approved):
LGTM. Thanks, @pradithya !

krithika369 (approved):
LGTM. Thanks, @pradithya !

infra/charts/turing/templates/_helpers.tpl
pradithya: Updated

PR #339, opened 2023-05-20
skip: Introduce semver to workflows

terryyylim:
This PR introduces `Semantic Versioning` when releasing `turing`.

Specifically, the following updates have been made to the workflow:
- All PRs require a prefix that would auto-label the PR
- Auto-labeling has been added, which caters to the following categories for `release-drafter`:

Version Bump Category | PR Title <> matching label
--- | ---
MAJOR | `breaking: ...` <> `breaking`
MINOR | `feat: ...` <> `feature`
PATCH | `fix: ...` <> bug<br/>`docs: ...` <> documentation<br/>`chore: ...` <> maintenance<br/>`build: ...` <> dependencies<br/>`security: ...` <> security<br/>`others: ...` <> others
NONE | `skip: ...` <> `skip-changelog`

Relevant Documentation:
- https://github.com/amannn/action-semantic-pull-request
- https://www.conventionalcommits.org/

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,72 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+categories:
+  - title: 'Breaking 🔨'
+    label: 'type: breaking'
+  - title: 'New 🚀'
+    label: 'type: feature'
+  - title: 'Bug Fixes 🐛'
+    label: 'type: bug'
+  - title: 'Maintenance ⚙️'
+    label: 'type: maintenance'
+  - title: 'Documentation 📄'
+    label: 'type: documentation'
+  - title: 'Other changes'
+  - title: 'Dependency Updates 💪'
+    label: 'type: dependencies'
+    collapse-after: 5
+
+version-resolver:
+  major:
+    labels:
+      - 'type: breaking'
+  minor:
+    labels:
+      - 'type: feature'
+  patch:
+    labels:
+      - 'type: bug'
+      - 'type: maintenance'
+      - 'type: documentation'
+      - 'type: dependencies'
+      - 'type: security'
+      - 'type: others'
+  default: patch
+
+exclude-labels:
+  - 'skip-changelog'
+
+autolabeler:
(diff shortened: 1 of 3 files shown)


.github/release-drafter.yml
```diff
@@ -0,0 +1,72 @@
+name-template: 'v$RESOLVED_VERSION'
+tag-template: 'v$RESOLVED_VERSION'
+categories:
+  - title: 'Breaking 🔨'
+    label: 'type: breaking'
```
terryyylim, line 5: Do we want to have `type: ` prefix for labels? If so, we need to add new labels to the repository.

.github/release-drafter.yml
```diff
@@ -0,0 +1,72 @@
...
+version-resolver:
+  major:
+    labels:
+      - 'type: breaking'
+  minor:
+    labels:
+      - 'type: feature'
+  patch:
+    labels:
+      - 'type: bug'
```
terryyylim, line 28: Do these categories suffice or are they overly-granular? Open to discussion here 🙏🏻

krithika369: I think they look fine. I was thinking it would be nice to have a `refactor:` PR prefix and matching `cleanup` label, for code refactoring changes without any real functionality change. Wdyt?

.github/release-drafter.yml
krithika369: Per your PR, the `type:` prefix exists, right? It's perhaps not necessary now, with such a short list, but could be useful to keep the prefix in case we have more labels down the road that need to be categorised among themselves.

Are you able to update the labels in the repo?

krithika369 (approved):
LGTM. Left some suggestion. Thanks for this @terryyylim !

.github/release-drafter.yml
terryyylim: Ah yeah, I have that categorised under `chore: ` PR prefix and matches `type: maintenance` label. Would that suffice?

.github/release-drafter.yml
terryyylim: Updated the labels!

PR #340, opened 2023-05-31
Add PodDisruptionBudget to router components

ariefrahmansyah:
- Add PDB config
- Add component label
- Modify updateFunc to use generic

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -39,8 +39,8 @@ vendor:
 .PHONY: test
 test: tidy
 	@echo "Running tests..."
-	go test -v -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
-	go tool cover -func cover.out
+	GOARCH=amd64 go test -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
+	GOARCH=amd64 go tool cover -func cover.out
 
 .PHONY: build
 build: tidy
--- a/api/turing/cluster/labeller/labeller.go
+++ b/api/turing/cluster/labeller/labeller.go
@@ -13,12 +13,16 @@ const (
 	teamLabel = "team"
 	// orchestratorLabel refers to the orchestrator that deployed the Kubernetes object
 	orchestratorLabel = "orchestrator"
+	// componentLabel refers to the component that deployed the Kubernetes object
+	componentLabel = "component"
 	// AppLabel refers to application of the Kubernetes Object
 	AppLabel = "app"
 )
 
-var prefix string
-var environment string
+var (
+	prefix      string
+	environment string
+)
 
 // InitKubernetesLabeller builds a new KubernetesLabeller Singleton
 func InitKubernetesLabeller(p, e string) {
@@ -28,9 +32,10 @@ func InitKubernetesLabeller(p, e string) {
 
 // KubernetesLabelsRequest helps to build the Kubernetes labels needed for Kubernetes objects.
 type KubernetesLabelsRequest struct {
-	Stream string
-	Team   string
-	App    string
+	Stream    string
+	Team      string
+	App       string
+	Component string
 }
 
 // GetLabelName prefixes the label with the config specified label and returns the formatted label name
@@ -45,6 +50,7 @@ func BuildLabels(r KubernetesLabelsRequest) map[string]string {
 		GetLabelName(streamLabel):       r.Stream,
 		GetLabelName(teamLabel):         r.Team,
 		GetLabelName(AppLabel):          r.App,
+		GetLabelName(componentLabel):    r.Component,
 		GetLabelName(environmentLabel):  environment,
 	}
 }
--- a/api/turing/cluster/pdb.go
+++ b/api/turing/cluster/pdb.go
@@ -1,0 +1,51 @@
+package cluster
+
+import (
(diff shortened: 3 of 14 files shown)


api/turing/service/router_deployment_service.go
```diff
@@ -626,15 +666,15 @@ func deleteKnServices(
...
 }
 
-// updateKnServices is a helper method for deployment / deletion of services that runs the
+// updateResources is a helper method for deployment / deletion of resources that runs the
 // given update function on the given services simultaneously and waits for a response,
 // within the supplied timeout.
-func updateKnServices(ctx context.Context, services []*cluster.KnativeService,
-	updateFunc uFunc, eventsCh *EventChannel) error {
-
+func updateResources[V *cluster.KnativeService | *cluster.PodDisruptionBudget](ctx context.Context, services []V,
```
ariefrahmansyah, line 675: I'm refactoring updateKnServices to be updateResources that use generic so it can accept both KnativeServie and PodDisruptionBudget. Let me know what do you think about this. @krithika369

api/turing/cluster/labeller/labeller.go
```diff
@@ -28,9 +32,10 @@ func InitKubernetesLabeller(p, e string) {
 
 // KubernetesLabelsRequest helps to build the Kubernetes labels needed for Kubernetes objects.
 type KubernetesLabelsRequest struct {
-	Stream string
-	Team   string
-	App    string
+	Stream    string
+	Team      string
+	App       string
+	Component string
```
ariefrahmansyah, line 38: For PodDisruptionBudget's LabelSelector to work correctly, I added component label here. @krithika369

api/Makefile
```diff
@@ -39,8 +39,8 @@ vendor:
 .PHONY: test
 test: tidy
 	@echo "Running tests..."
-	go test -v -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
-	go tool cover -func cover.out
+	GOARCH=amd64 go test -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
```
krithika369, line 42: Should we make the `GOARCH` value configurable so it's easy to build for other environments too ? Like: https://github.com/caraml-dev/turing/blob/main/engines/experiment/examples/plugins/hardcoded/Dockerfile#L8

api/turing/cluster/labeller/labeller.go
krithika369: Hmm.. Adding lots of labels has a cost too (like with metrics scraping, unless we selectively drop some labels downstream). Instead, can we rely on the default `app` label added by the k8s Deployment controller, the same way that the TopologySpreadConstraints work? https://github.com/caraml-dev/turing/blob/main/api/turing/cluster/knative_service.go#L165

api/turing/cluster/pdb.go
```diff
@@ -0,0 +1,51 @@
...
+	Name           string                `json:"name"`
+	Namespace      string                `json:"namespace"`
+	Labels         map[string]string     `json:"labels"`
+	MaxUnavailable string                `json:"max_unavailable"`
+	MinAvailable   string                `json:"min_available"`
+	Selector       *metav1.LabelSelector `json:"selector"`
+}
+
+func (cfg PodDisruptionBudget) BuildPodDisruptionBudget() (*apipolicyv1.PodDisruptionBudget, error) {
+	if cfg.MaxUnavailable == "" && cfg.MinAvailable == "" {
```
krithika369, line 21: Since this check will only kick in much later (when user attempts to deploy a router), can we move it to the config validator instead? https://github.com/caraml-dev/turing/blob/main/api/turing/config/config.go#L619

This way, if one of these 2 values is not set, Turing app can't start up.

api/turing/service/router_deployment_service.go
```diff
@@ -677,3 +717,133 @@ const (
...
+			errCh <- err
+			return
+		}
+
+		eventsCh.Write(models.NewInfoEvent(
+			models.EventStageDeployingServices, "deleting pdb %s", pdb.Name))
+
+		err := controller.DeletePodDisruptionBudget(ctx, pdb.Namespace, pdb.Name)
+		if err != nil {
+			err = errors.Wrapf(err, "Failed to deploy pdb %s", pdb.Name)
```
krithika369, line 832: This should be "undeploy"

api/turing/service/router_deployment_service.go
```diff
@@ -677,3 +717,133 @@ const (
...
+			eventsCh.Write(models.NewErrorEvent(
+				models.EventStageDeployingServices, "failed to delete pdb: %s", err.Error()))
+			errCh <- err
+			return
+		}
+
+		eventsCh.Write(models.NewInfoEvent(
+			models.EventStageDeployingServices, "deleting pdb %s", pdb.Name))
+
+		err := controller.DeletePodDisruptionBudget(ctx, pdb.Namespace, pdb.Name)
```
krithika369, line 830: In previous releases, we wouldn't have deployed PDBs - so we cannot delete them either. Will issuing the delete call return an error? If so, we should ignore such error as it's a valid scenario.

ariefrahmansyah:
I'm closing this since I'm raising similar PR from my fork: #346

PR #341, opened 2023-06-05
Bugfix: remove validation for resultlogger of upi router in sdk

leonlnj:
# Description
Short: Remove validation in SDK for UPI Router that was missed earlier on.

When UPI Logger was [first introduced](https://github.com/caraml-dev/turing/pull/319/files#diff-49fa6d9d67eac9ae9ac62b762e4db2c2b72de2d407f4a5aa1930e94a0f187a97), it was never intended to work with BQ/Kafka ResultLogger. 

Subsequently, BQ/Kafka logger are enabled for UPI Router in this [PR](https://github.com/caraml-dev/turing/pull/323/files#diff-2bc7d30f97d4ecb41bd4e1ac0a1727086edf6557f30f409991db83e1d7278bcd) 
using `TuringResultLog` instead of `RouterLog` and the validation for these were removed for the API Server but not SDK.

This PR removes the SDK validations that should had been removing with the previous PR.

### DIFF
--- a/sdk/tests/router/config/router_config_test.py
+++ b/sdk/tests/router/config/router_config_test.py
@@ -198,41 +198,6 @@ def test_default_router_autoscaling_policy(request):
 
 
 @pytest.mark.parametrize(
-    "base_config, log_config, expected",
-    [
-        pytest.param(
-            "minimal_upi_router_config",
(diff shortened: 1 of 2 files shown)


deadlycoconuts (approved):
LGTM 🚀 Thanks for removing this!

PR #342, opened 2023-06-06
Bump fiber version

tiopramayudi:
Bump fiber library to use the latest (v0.2.2) that already support gRPC connection pool


leonlnj (approved):
thanks!

PR #344, opened 2023-06-14
feat: Add authz caching config

krithika369:
This PR largely replicates the changes for the API in https://github.com/caraml-dev/mlp/pull/87, to have the option of caching authz responses. As a part of this implementation, the API dependency on MLP has been upgraded.


deadlycoconuts (approved):
LGTM! 🚀 Thanks for the changes!

PR #346, opened 2023-06-21
Add PodDisruptionBudget to router components

ariefrahmansyah:
This PR will add PodDisruptionBudget resources to Turing components (router, enricher, and ensembler).

Modifications:

1. Add PodDisruptionBudget config
2. Add Create and Delete PodDisruptionBudget on Cluster Controller
3. Refactor updateKnServices to updateResources so it can accept both KnativeService and PodDisruptionBudget resources
4. Introduce `gotest` to Makefile and GitHub actions
   a. In GitHub actions, add `make setup` to install `gotest`


api/turing/config/config.go
```diff
@@ -231,6 +231,16 @@ type DeploymentConfig struct {
...
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint
+	PodDisruptionBudget       PodDisruptionBudgetConfig
+}
+
+// PodDisruptionBudgetConfig are the configuration for PodDisruptionBudgetConfig for
+// Turing services.
+type PodDisruptionBudgetConfig struct {
+	Enabled bool
+	// Can specify only one of maxUnavailable and minAvailable
+	MaxUnavailable string
```
krithika369, line 242: Shall we instead name this as `MaxUnavailablePercent` and use `*int` ? Or, if we are sticking with the string type, should we check that the value is of the expected format (like "20%")? Would be nice to add some checks around this as well because otherwise, the earlier we can discover a problem with this config is when we deploy a router, which may be too late.

api/turing/service/router_deployment_service.go
```diff
@@ -546,12 +566,22 @@ func deployKnServices(
...
-		svc *cluster.KnativeService,
+		resource any,
 		wg *sync.WaitGroup,
 		errCh chan<- error,
 		eventsCh *EventChannel,
 	) {
 		defer wg.Done()
+
+		svc, ok := resource.(*cluster.KnativeService)
+		if !ok {
```
krithika369, line 577: Thanks for adding this check. Curious - what was the motivation for adding it?

api/turing/service/router_deployment_service.go
```diff
@@ -677,3 +717,133 @@ const (
...
+		return errors.New("timeout deploying pdb")
+	default:
+		return updateResources(ctx, pdbs, deployFunc, eventsCh)
+	}
+}
+
+func deletePodDisruptionBudgets(ctx context.Context,
+	controller cluster.Controller, pdbs []*cluster.PodDisruptionBudget, eventsCh *EventChannel,
+) error {
+	// Define deploy function
```
krithika369, line 809: Nit: Should this read "delete function" ?

krithika369 (approved):
Left some questions / comments. The rest LGTM. Thanks, @ariefrahmansyah !

api/turing/service/router_deployment_service.go
ariefrahmansyah: deployFunc is refactored to accept any (interface) instead of structural type such as *cluster.KnativeService so the deployFunc can be reused to deploy other resources such as PDB. This line is added to safely cast the resource interface to the concrete structural type.

krithika369 (approved):
LGTM. Thanks for the changes, @ariefrahmansyah ! Please feel free to merge once the linter errors are addressed.

PR #347, opened 2023-06-22
fix: Trim path prefix when checking for authorization

krithika369:
This PR corrects the path manipulation when checking for authorization, to trim the path prefix, which was accidentally removed in: https://github.com/caraml-dev/turing/pull/344

### DIFF
--- a/api/turing/middleware/authorization.go
+++ b/api/turing/middleware/authorization.go
@@ -104,7 +104,7 @@ func getResourceFromPath(path string, prefix string) string {
 	// if a user has READ/WRITE permissions on /projects/{project_id}, they would also have the same
 	// permissions on all its sub-resources. Thus, trimming the resource identifier to aid quicker
 	// authz matching and to efficiently make use of the in-memory authz cache, if enabled.
-	parts := strings.Split(strings.TrimPrefix(path, "/"), "/")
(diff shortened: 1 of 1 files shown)


api/turing/middleware/authorization.go
```diff
@@ -104,7 +104,7 @@ func getResourceFromPath(path string, prefix string) string {
 	// if a user has READ/WRITE permissions on /projects/{project_id}, they would also have the same
 	// permissions on all its sub-resources. Thus, trimming the resource identifier to aid quicker
 	// authz matching and to efficiently make use of the in-memory authz cache, if enabled.
-	parts := strings.Split(strings.TrimPrefix(path, "/"), "/")
+	parts := strings.Split(strings.TrimPrefix(strings.TrimPrefix(path, prefix), "/"), "/")
```
deadlycoconuts, line 107: I thought this change was intentional 👀

deadlycoconuts (approved):

api/turing/middleware/authorization.go
krithika369: No, unfortunately, I had overlooked this. Turing stores the path prefix `/v1` which must be removed from the resource prior to authz checks. I had accidentally removed this trimming.

krithika369:
Thanks for the quick review, @deadlycoconuts !

PR #349, opened 2023-06-30
Fix crash due to mismatch number of columns and values in a row

tiopramayudi:
Turing router crash when trying to convert invalid UPI request. The root cause is due to bug in the UPI utility function to convert UPI table in struct. Following are the modifications:
* Bump version of UPI libarary
* Add panic recovery interceptors

### DIFF
--- a/engines/router/missionctl/log/resultlog/upi_result_log_test.go
+++ b/engines/router/missionctl/log/resultlog/upi_result_log_test.go
@@ -194,10 +194,10 @@ func TestUPIResultLogger_LogTuringRouterRequestSummary_logRouterLog(t *testing.T
 		resultLogger *UPIResultLogger
 	}
 	tests := []struct {
-		name              string
-		args              args
-		want              *upiv1.RouterLog
-		skipPublishingLog bool
+		name           string
+		args           args
+		want           *upiv1.RouterLog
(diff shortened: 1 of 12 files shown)


engines/router/missionctl/log/resultlog/upi_result_log_test.go
```diff
@@ -317,6 +319,56 @@ func TestUPIResultLogger_LogTuringRouterRequestSummary_logRouterLog(t *testing.T
...
+					TargetName:        "target-name",
+					PredictionContext: predictionContext,
+					Metadata: &upiv1.RequestMetadata{
+						PredictionId:     "123",
+						RequestTimestamp: time,
+					},
+				},
+				resultLogger: &UPIResultLogger{upiLogger: &mockUPILogger{}, loggerType: config.UPILogger},
+			},
+			skipPublishingLog: true,
```
leonlnj, line 370: maybe rename to `expectLogError`?

leonlnj (approved):
thanks LGTM

engines/router/missionctl/log/resultlog/upi_result_log_test.go
tiopramayudi: updated

PR #350, opened 2023-07-08
Set initial scale during update of a router deployment

krithika369:
### Summary

This PR includes the addition of the [Knative Intial Scale](https://knative.dev/docs/serving/autoscaling/scale-bounds/#initial-scale) annotation to the components deployed for the router. When deploying a router version, if there is an existing version of the router already deployed, their respective components' current desired scale will be copied over as the initial scale for the new deployment. This way, the new deployment will start with a number of replicas that matches the old deployment, rather than at min replicas.

### Changes
* `api/turing/api/deployment_controller.go` - Pass down the current router version to the deployment service
* `api/turing/cluster/controller.go` - Add `GetKnativeServiceDesiredReplicas` to retrieve the currently deployed revision's desired replicas, if exists
* `api/turing/cluster/knative_service.go` - Add `InitialReplicas` to the `KnativeService` struct and copy its value to the K8s object's annotations, if set
* `api/turing/cluster/servicebuilder/*` - Propagate the initial scale down to the `KnativeService` struct
* `api/turing/service/router_deployment_service.go` - `createServices` has the main logic changes to compare current version's properties and set the initial scale on the new deployment
* Updated unit tests


api/turing/cluster/controller.go
```diff
@@ -301,6 +302,26 @@ func (c *controller) DeleteKnativeService(
...
+
+	rev, err := revisions.Get(ctx, getDefaultRevisionName(svcName), metav1.GetOptions{})
+	if err != nil {
+		return 0, err
+	}
+	if rev.Status.DesiredReplicas != nil {
+		return int(*rev.Status.DesiredReplicas), nil
+	}
+
+	return 0, fmt.Errorf("Desired Replicas for %s/%s is not set", namespace, svcName)
```
ariefrahmansyah, line 322: nit: I think it's better to switch it so that we handle and return invalid case first:

```
	if rev.Status.DesiredReplicas == nil {
		return 0, fmt.Errorf("Desired Replicas for %s/%s is not set", namespace, svcName)
	}

	return int(*rev.Status.DesiredReplicas), nil
```

api/turing/cluster/knative_service.go
```diff
@@ -118,12 +119,16 @@ func (cfg *KnativeService) buildSvcSpec(
 		"autoscaling.knative.dev/class":    autoscalingMetricClassMap[cfg.AutoscalingMetric],
 	}
 
+	if cfg.InitialReplicas != nil {
+		annotations["autoscaling.knative.dev/initial-scale"] = strconv.Itoa(*cfg.InitialReplicas)
```
ariefrahmansyah, line 123: nit: Since the annotation is `initial-scale` I think it's better to rename `InitialReplicas` to `InitialScale` to avoid more confusion and make it consistent across Turing components.

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -136,6 +139,7 @@ func (sb *clusterSvcBuilder) NewEnricherService(
 	secretName string,
 	knativeQueueProxyResourcePercentage int,
 	userContainerLimitRequestFactor float64,
+	initialReplicas *int,
```
ariefrahmansyah, line 142: The interface uses `initialScale` but the implementation uses `initialReplicas`. Similar to my comment above, I think we should make it consistent.

api/turing/service/router_deployment_service.go
```diff
@@ -349,12 +357,26 @@ func (ds *deploymentService) createServices(
...
+			currRouterVersion.Enricher != nil {
+			currEnricherSvc, err := ds.svcBuilder.NewEnricherService(
+				currRouterVersion, project, secretName,
+				knativeQueueProxyResourcePercentage, userContainerLimitRequestFactor, nil,
+			)
+			if err != nil {
+				return services, fmt.Errorf("Unable to build the details for the currently deployed enricher: %w", err)
+			}
+			currReplicas, err := controller.GetKnativeServiceDesiredReplicas(ctx,
+				currEnricherSvc.Name, currEnricherSvc.Namespace)
```
ariefrahmansyah, line 372: Can we avoid creating this currEnricherSvc object and directly used the GetComponentName() and GetNamespace() function? Because, if I understand it correctly, we don't need other field from NewEnricherService().

api/turing/service/router_deployment_service.go
```diff
@@ -364,12 +386,30 @@ func (ds *deploymentService) createServices(
...
+				secretName,
+				knativeQueueProxyResourcePercentage,
+				userContainerLimitRequestFactor,
+				nil,
+			)
+			if err != nil {
+				return services, fmt.Errorf("Unable to build the details for the currently deployed ensembler: %w", err)
+			}
+			currReplicas, err := controller.GetKnativeServiceDesiredReplicas(ctx,
+				currEnsemblerSvc.Name, currEnsemblerSvc.Namespace)
```
ariefrahmansyah, line 405: Same as above on avoiding creating new currEnsemblerSvc object.

api/turing/service/router_deployment_service.go
krithika369: Fair point. I've refactored this part.

ariefrahmansyah (approved):
LGTM! Awesome job as always, @krithika369!

krithika369:
Thanks for the detailed review, @ariefrahmansyah ! Merging.

PR #351, opened 2023-07-13
fix: Missing experiment mappings in standard ensembler config

krithika369:
This PR adds a small fix in `sdk/turing/router/config/router_ensembler_config.py`, which was preventing experiment mappings in the Standard Ensembler Config from being saved correctly, in the result of a get/list router call.

Added unit tests to test the initialization of `RouterEnsemblerConfig` with different types of configs, and that the resulting structure has the all the relevant data stored.

### DIFF
--- a/sdk/turing/router/config/router_ensembler_config.py
+++ b/sdk/turing/router/config/router_ensembler_config.py
@@ -160,16 +160,14 @@ class RouterEnsemblerConfig(DataObject):
             self._standard_config = standard_config
         elif isinstance(standard_config, dict):
             openapi_standard_config = standard_config.copy()
-            openapi_standard_config["experiment_mappings"] = (
-                [
+            openapi_standard_config["experiment_mappings"] = None
+            if "experiment_mappings" in standard_config and standard_config["experiment_mappings"] is not None:
+                openapi_standard_config["experiment_mappings"] = [
                     turing.generated.models.EnsemblerStandardConfigExperimentMappings(
                         **mapping
                     )
                     for mapping in standard_config["experiment_mappings"]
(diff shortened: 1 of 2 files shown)


sdk/turing/router/config/router_ensembler_config.py
```diff
@@ -160,16 +160,14 @@ def standard_config(self, standard_config: Union[EnsemblerStandardConfig, Dict])
...
+            openapi_standard_config["experiment_mappings"] = None
+            if "experiment_mappings" in standard_config and standard_config["experiment_mappings"] is not None:
+                openapi_standard_config["experiment_mappings"] = [
                     turing.generated.models.EnsemblerStandardConfigExperimentMappings(
                         **mapping
                     )
                     for mapping in standard_config["experiment_mappings"]
                 ]
-                if openapi_standard_config is None
-                else None
```
deadlycoconuts, line 171: Just to make sure I understood what was happening previously, was `openapi_standard_config["experiment_mappings"]` set as `None` previously because `openapi_standard_config` is non-nil (in the case of the router that Roman was using)?

deadlycoconuts (approved):
LGTM! 🚀 Thanks a lot for investigating this and for the fixes + additional checks 🙏

sdk/turing/router/config/router_ensembler_config.py
krithika369: Correct. Experiment mappings should in fact have been non-nil if openapi_standard_config["experiment_mappings"] (or standard_config["experiment_mappings"]) was non-nil. Instead, it was being set to None.

PR #348, opened 2023-06-30
fix some styling on the deletion feature

zenovore:
This PR is a follow up regarding some issue from the #329 :
1. Different table alignment from turing and merlin
2. Hide Terminate/Delete Button when ensembling job is on `terminating` state

PR #352, opened 2023-07-18
Fix ensembler deletion UI bugs

deadlycoconuts:
## Context
Similar to PR #348 which introduced some UI fixes to the ensembler deletion workflow (which itself had been introduced in PR #329), this PR acts as a follow up to implement some additional minor fixes to the UI for the bugs below:
- when deleting multiple ensemblers in a row, the confirmation text box in the pop-up gets pre-filled with the ensembler name from the previous ensembler that had been deleted
image
- when opening and closing the deletion pop-up multiple times in a row (without actually deleting the ensembler), the information shown in the pop-up is not consistent (see the video at the bottom of this description for an illustration).

Besides these bug fixes, some other non-bug related improvements have been introduced:
- some minor changes to the formatting of the pop messages to better reflect the different types/quantities of router/router versions/ensembling jobs returned (see https://github.com/caraml-dev/turing/pull/352#issuecomment-1641552190)
- a modification to a change introduced in PR #348 whereby instead of hiding the 'Terminate/Delete' button if an ensembling job is in the 'terminating' state, we are now showing the button but greying that out instead

https://github.com/caraml-dev/turing/assets/36802364/9fa7e60d-1faf-47a5-9796-aebd47c95d45

### DIFF
--- a/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
+++ b/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
@@ -70,9 +70,9 @@ export const DeleteEnsemblerModal = ({
           {canDeleteEnsembler ? (
             <div>
               <p>
-              You are about to delete Ensembler <b>{ensembler.name}</b>. This action cannot be undone. 
+              You are about to delete the Ensembler <b>{ensembler.name}</b>. This action <b>cannot</b> be undone.
               </p>
-              To confirm, please type "<b>{ensembler.name}</b>" in the box below
+              To confirm, please type "<b>{ensembler.name}</b>" in the box below:
               <EuiFieldText     
                 fullWidth            
                 placeholder={ensembler.name}
@@ -80,16 +80,14 @@ export const DeleteEnsemblerModal = ({
                 onChange={(e) => setDeleteConfirmation(e.target.value)}
                 isInvalid={deleteConfirmation !== ensembler.name} />              
             </div>
-          ) : ensemblerUsedByCurrentRouterVersion ? (
-            <div>
-              You cannot delete this ensembler because it is associated with a router version that is currently being used by a router
-              <br/> <br/> If you still wish to delete this ensembler, please <b>Deploy</b> another version on this router.
-            </div>
           ) : (
             <div>
-              You cannot delete this ensembler because there are <b>Active Router Versions</b> or <b>Ensembling Jobs</b> that use this ensembler. 
-              <br/> <br/> If you still wish to delete this ensembler, please <b>Undeploy</b> router versions and <b>Terminate</b> ensembling jobs that use this ensembler.
-            </div> 
+              You cannot delete this ensembler because it is:
+              <ul>
+                <li>associated with one or more router versions <b>currently</b> used by one or more routers, and/or</li>
+                <li>used by one or more <b>Active Router Versions</b> or <b>Ensembling Jobs</b></li>
+              </ul>
+            </div>
           )}
           {/* Only show The Ensembling Table if ensembler is not used by current router version */}
           {!ensemblerUsedByCurrentRouterVersion && (
--- a/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
+++ b/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
@@ -61,7 +61,7 @@ export const DeleteEnsemblerModal = ({
     <ConfirmationModal
       title="Delete Ensembler"
       onCancel={() => modalClosed()}
-      onConfirm={submitForm}
+      onConfirm={(arg) => {submitForm(arg); modalClosed()}}
       isLoading={isLoading}
       content={
         <div>
--- a/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
+++ b/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
@@ -84,8 +84,11 @@ export const DeleteEnsemblerModal = ({
             <div>
               You cannot delete this ensembler because it is:
               <ul>
-                <li>associated with one or more router versions <b>currently</b> used by one or more routers, and/or</li>
-                <li>used by one or more <b>Active Router Versions</b> or <b>Ensembling Jobs</b></li>
+                {ensemblerUsedByCurrentRouterVersion &&
+                  <li>associated with one or more router versions <b>currently</b> used by one or more routers
+                    {ensemblerUsedByActiveRouterVersion && ", and"}</li>}
+                {ensemblerUsedByActiveRouterVersion &&
+                  <li>used by one or more <b>Active Router Versions</b> or <b>Ensembling Jobs</b></li>}
               </ul>
             </div>
           )}
--- a/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
+++ b/ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
@@ -54,6 +54,8 @@ export const DeleteEnsemblerModal = ({
   const modalClosed = () => {
     setCanDeleteEnsembler(true)
     setEnsemblerUsedByCurrentRouterVersion(false)
+    setEnsemblerUsedByActiveRouterVersion(false)
+    setEnsemblerUsedByActiveEnsemblingJob(false)
     setDeleteConfirmation("")
   }
 
--- a/ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
+++ b/ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
@@ -107,7 +107,7 @@ export const ListEnsemblingJobsForEnsemblerTable = ({
         <div>
           <br/>
           This Ensembler is used by {results.totalActiveCount} <b>Active Ensembling Job{results.totalActiveCount > 1 && "s"}</b>.
-          If any ensembling jobs are in the terminating state, please wait until they are complete to delete this ensembler.
+          If any ensembling jobs are in the terminating state, please wait until they are terminated to delete this ensembler.
           <EuiBasicTable
             items={results.activeItems}
             loading={!isLoaded}
--- a/ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
+++ b/ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
@@ -93,7 +93,8 @@ export const ListEnsemblingJobsForEnsemblerTable = ({
       {canDeleteEnsembler ? ( results.totalInactiveCount > 0 && (
         <div>
           <br/>
-          <p>Deleting this Ensembler will also delete {results.totalInactiveCount} <b>Failed</b> or <b>Completed</b> Ensembling Jobs that use this Ensembler </p>
+          Deleting this Ensembler will also delete {results.totalInactiveCount} <b>Failed</b> or <b>Completed</b>
+          &nbsp;Ensembling Job{results.totalInactiveCount > 1 && "s"} that use{results.totalInactiveCount === 1 && "s"} this Ensembler:
           <EuiBasicTable
             items={results.inactiveItems}
             loading={!isLoaded}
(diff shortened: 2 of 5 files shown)


ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
```diff
@@ -61,7 +61,7 @@ export const DeleteEnsemblerModal = ({
     <ConfirmationModal
       title="Delete Ensembler"
       onCancel={() => modalClosed()}
-      onConfirm={submitForm}
+      onConfirm={(arg) => {submitForm(arg); modalClosed()}}
```
deadlycoconuts, line 64: This change ensures that modalClosed gets called when the form submission is successful, which resets the confirmation text field:
```js
  const modalClosed = () => {
    setCanDeleteEnsembler(true)
    setEnsemblerUsedByCurrentRouterVersion(false)
    setDeleteConfirmation("")
  }
```

ui/src/jobs/components/modal/DeleteJobModal.js
```diff
@@ -45,7 +45,8 @@ export const DeleteJobModal = ({
   return (
     <ConfirmationModal
       title={isActiveJobStatus(job.status) ? 'Delete Ensembling Jobs' : 'Terminate Ensembling Jobs'}
-      onConfirm={submitForm}
+      onCancel={() => setDeleteConfirmation("")}
+      onConfirm={(arg) => {submitForm(arg); setDeleteConfirmation("")}}
```
deadlycoconuts, line 49: There is no `modalClosed` function here, so we're just calling `setDeleteConfirmation` directly to reset the deletion text field.

zenovore (approved):
LGTM 😁👍, sorry I didn't realize this case, I forgot to clear the state on success 🙏

deadlycoconuts:
Thanks for looking at this! I'm still trying to fix the inconsistent message bug though 😅 I'll update again this PR again when I find a fix!

zenovore:
> Thanks for looking at this! I'm still trying to fix the inconsistent message bug though 😅 I'll update again this PR again when I find a fix!

by inconsistent message bug do you mean, the error message shown on the popup?

deadlycoconuts:
> by inconsistent message bug do you mean, the error message shown on the popup?

Yep that's right, the error messages changes instead of remaining constant 😅

ui/src/ensembler/components/table/ListRouterVersionsForEnsemblerTable.js
```diff
@@ -115,7 +118,7 @@ export const ListRouterVersionsForEnsemblerTable = ({
     </EuiCallOut>
   ) : (
     <Fragment>
-      {ensemblerUsedByCurrentRouterVersion ? (
+      {(ensemblerUsedByCurrentRouterVersion && results.totalActiveCount === 0) ? (
```
deadlycoconuts, line 121: This condition has been added to ensure that the active router versions using an ensembler are listed first (if there are any) instead of listing the current router versions. This is because all active router versions are current router versions but not the other way round. 

In order to ensure that the UI displays these tables in this order of priority, the new condition is required to ensure that the current router versions are only listed when there are NO active router versions.

ui/src/ensembler/components/table/ListRouterVersionsForEnsemblerTable.js
```diff
@@ -51,6 +51,9 @@ export const ListRouterVersionsForEnsemblerTable = ({
   useEffect(() => {
     if (results.activeItems.length > 0){
       setEnsemblerUsedByActiveRouterVersion(true)
+      // If there is an active router version using this ensembler, it is definitely also a current router version using
+      // this ensembler
+      setEnsemblerUsedByCurrentRouterVersion(true)
```
deadlycoconuts, line 56: This change keeps the `ensemblerUsedByCurrentRouterVersion` state consistent with the actual status of the ensembler (as of present, there's a `modalClosed` function that unsets this state to `false` whenever the modal is closed - previously it was not guaranteed that this state would even get updated, especially if the condition `if (results.activeItems.length > 0)` was fulfilled), which allows the effect here https://github.com/caraml-dev/turing/pull/352/files#diff-2852d3e270c0d9ad9c03a7cfffb3edc5b9723d381090490fa33bd7496697eea7R26 to update the `canDeleteEnsembler` state correctly.

ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
```diff
@@ -54,14 +54,16 @@ export const DeleteEnsemblerModal = ({
   const modalClosed = () => {
     setCanDeleteEnsembler(true)
     setEnsemblerUsedByCurrentRouterVersion(false)
+    setEnsemblerUsedByActiveRouterVersion(false)
+    setEnsemblerUsedByActiveEnsemblingJob(false)
```
deadlycoconuts, line 58: These two setters are introduced to ensure the states are being reset to their default values.

deadlycoconuts:
Okay the fix for the 2nd bug pretty much just attempts to ensure all the states are set correctly because as of present some of them end up stale or do not get updated when we expect them to, causing the error message to be shown inconsistently. 

I guess in the future one way to make this component even simpler is to not use so many effects/hooks/states manipulation within the child classes of `ListEnsemblersTable` 😅  We can perhaps get all the API responses we need in a single component, say `DeleteEnsemblerModal` and then pass all the required data to display (current/active router versions for example) to its corresponding child components without calling effects and updating states across various levels.

There's also the additional benefit of not having to re-render the `DeleteEnsemblerModal` if its child components change some state (to the users this would appear as if the pop-up content got refreshed automatically) - look below for a sample.

But that said, that's for another time when we have more time to refactor this 😅 It's still pretty amazing how you came up with such a complex workflow in order to support such detailed information being shown, so thanks a lot for working on the initial version of this @zenovore!

https://github.com/caraml-dev/turing/assets/36802364/7c5e2ee2-d3b3-4376-bb97-57af0a36de6c

ui/src/ensembler/components/table/ListRouterVersionsForEnsemblerTable.js
```diff
@@ -51,6 +51,9 @@ export const ListRouterVersionsForEnsemblerTable = ({
   useEffect(() => {
     if (results.activeItems.length > 0){
       setEnsemblerUsedByActiveRouterVersion(true)
+      // If there is an active router version using this ensembler, it is definitely also a current router version using
```
krithika369, line 54: This is not necessarily true, right? A router's current version is only updated after a deployment succeeds. In the scenario where a router is being updated to a new version, say, from v4 to v5, it could be the case that v5 refers to the ensembler (which is an active version according to the logic above) but v4 (the current version) doesn't.

deadlycoconuts: Hmmm 🤔 that makes sense. Okay let me take a look at how to iron out all these conditional cases again. Thanks for pointing this out!

deadlycoconuts: Okay, just to share an update on this, we've decided to display *both* current and active router versions if an ensembler is associated with router versions of both types. In addition, if any of the associated router versions are *both* current and active, we'll only display that router version in the current router version table. This requires some additional refactoring which I will annotate shortly.

ui/src/ensembler/components/table/ListRouterVersionsForEnsemblerTable.js
```diff
@@ -50,16 +51,15 @@ export const ListRouterVersionsForEnsemblerTable = ({
...
-        }
+      setEnsemblerUsedByActiveRouterVersion(true);
+    }
+
+    if (currentRouterVersion.isLoaded && !currentRouterVersion.error) {
+      if (currentRouterVersion.data.length > 0){
+        setEnsemblerUsedByCurrentRouterVersion(true);
+        setFilteredActiveRouterVersions(results.activeItems.filter(item => currentRouterVersion.data.every((e => e.id !== item.id))));
+      } else {
+        setEnsemblerUsedByCurrentRouterVersion(false)
```
deadlycoconuts, line 62: These changes are made to allow the `ensemblerUsedByCurrentRouterVersion` and `ensemblerUsedByActiveRouterVersion` states to be set independently.

ui/src/ensembler/components/table/ListRouterVersionsForEnsemblerTable.js
```diff
@@ -115,44 +115,54 @@ export const ListRouterVersionsForEnsemblerTable = ({
     </EuiCallOut>
   ) : (
     <Fragment>
-      {ensemblerUsedByCurrentRouterVersion ? (
-        currentRouterVersion.data.length > 0 && (
+      {canDeleteEnsembler && results.totalInactiveCount > 0 && (
```
deadlycoconuts, line 118: The conditional logic here has been simplified - we first check if the ensembler can be deleted - if it can, we'll display the deletion warnings. Otherwise, we'll display the current router versions table, followed by the active router versions table.

deadlycoconuts:
Besides the actual fixes, I've also reformatted some of the error messages to reflect how we're displaying the current and active router versions in the confirmation pop up. There's also some other unrelated changes to the wording of the messages but they mostly have to do with managing the singular/plural form of nouns/verbs being shown, depending on the number of items that will be displayed:
Screenshot 2023-07-19 at 3 09 42 PM
Screenshot 2023-07-19 at 3 15 15 PM

ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
```diff
@@ -105,15 +106,16 @@ export const ListEnsemblingJobsForEnsemblerTable = ({
       )) : ( results.totalActiveCount > 0 && (
         <div>
           <br/>
-          <p>This Ensembler is being used by {results.totalActiveCount} <b>Active Ensembling Jobs</b>. 
-          If the ensembling jobs are in terminating state, please wait until the process complete to delete this ensembler</p>
+          This Ensembler is used by {results.totalActiveCount} <b>Active Ensembling Job{results.totalActiveCount > 1 && "s"}</b>.
+          If any ensembling jobs are in the terminating state, please wait until they are complete to delete this ensembler.
```
krithika369, line 110: Nit: `please wait until they are complete` -> `please wait until they are terminated` ?

ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
```diff
@@ -54,40 +54,43 @@ export const DeleteEnsemblerModal = ({
...
+                {ensemblerUsedByCurrentRouterVersion &&
+                  <li>associated with one or more router versions <b>currently</b> used by one or more routers
+                    {ensemblerUsedByActiveRouterVersion && ", and"}</li>}
+                {ensemblerUsedByActiveRouterVersion &&
+                  <li>used by one or more <b>Active Router Versions</b> or <b>Ensembling Jobs</b></li>}
+              </ul>
+            </div>
           )}
           {/* Only show The Ensembling Table if ensembler is not used by current router version */}
           {!ensemblerUsedByCurrentRouterVersion && (
```
krithika369, line 96: In case that the data in the modal becomes too large, should we make the content scrollable? Per [this issue](https://github.com/elastic/eui/discussions/4701), this should work automatically if wrapping the contents with `EuiModalBody`.

krithika369 (approved):
Left some small suggestions. LGTM. Thanks for the fix, @deadlycoconuts !

ui/src/ensembler/components/table/ListEnsemblingJobsForEnsemblerTable.js
deadlycoconuts: Sounds good! Just made that change.

ui/src/ensembler/components/modal/DeleteEnsemblerModal.js
deadlycoconuts: Hmm I think it's already scrollable right? Unless you were referring to something else? 🤔 Like thee tables being scrollable?
ezgif com-video-to-gif

krithika369: Oh I didn't realize this. This is great. 👍🏽

deadlycoconuts:
I've also included this additional change to grey out the 'Terminate/Delete' button instead of hiding it when an ensembling job is in the 'terminating' state:
Screenshot 2023-07-20 at 11 51 28 AM

deadlycoconuts:
Thanks a lot for all the comments! I'll be merging this now! 🙏

PR #353, opened 2023-07-27
fix: Add precautionary logic for PDB

krithika369:
PDB support for router components was introduced in https://github.com/caraml-dev/turing/pull/346. This PR fixes some bugs / corner cases:
* For ensembler, PDB should only be created for Docker / Pyfunc types (as Standard ensembler is embedded in the router deployment)
* Do not create PDB if the min replica is too low, which can prevent node drain. See related Merlin PR for more details: https://github.com/caraml-dev/merlin/pull/436

### DIFF
--- a/api/turing/service/router_deployment_service.go
+++ b/api/turing/service/router_deployment_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"os"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


deadlycoconuts (approved):
Thanks a lot for all these fixes! 🎉 Just have a tiny question but it looks good!

api/turing/service/router_deployment_service.go
```diff
@@ -771,13 +798,17 @@ func (ds *deploymentService) createPodDisruptionBudgets(
...
 
 	// Router's PDB
-	routerPdb := ds.svcBuilder.NewPodDisruptionBudget(
-		routerVersion,
-		project,
-		servicebuilder.ComponentTypes.Router,
-		ds.pdbConfig,
-	)
-	pdbs = append(pdbs, routerPdb)
+	if routerVersion.ResourceRequest != nil &&
```
deadlycoconuts, line 801: Is it possible that the resource request be `nil`?

krithika369: It's not 😅 and I don't think we've cared about this in many parts of the code. But simply because we are accessing a pointer, I thought it would be better to include the check.

krithika369:
Thanks for the review. Merging.

PR #354, opened 2023-07-27
fix: Invert logic of isActiveJobStatus helper function

deadlycoconuts:
## Context
This context inverts the logic of the `isActiveJobStatus` helper function used in the UI, since its name doesn't really represent what it actually does (it actually does the inverse 👀)

### DIFF
--- a/ui/src/jobs/list/ListEnsemblingJobsTable.js
+++ b/ui/src/jobs/list/ListEnsemblingJobsTable.js
@@ -17,7 +17,7 @@ import { useNavigate } from "react-router-dom";
 import { useConfig } from "../../config";
 import moment from "moment";
 import { DeploymentStatusHealth } from "../../components/status_health/DeploymentStatusHealth";
-import { JobStatus } from "../../services/job/JobStatus";
+import { isActiveJobStatus, JobStatus } from "../../services/job/JobStatus";
 import EnsemblersContext from "../../providers/ensemblers/context";
 import { DateFromNow } from "@caraml-dev/ui-lib";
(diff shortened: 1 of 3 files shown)


ui/src/jobs/list/ListEnsemblingJobsTable.js
```diff
@@ -70,8 +70,8 @@ export const ListEnsemblingJobsTable = ({
   }
   
   const isActiveJobStatus = function(jobStatus) {
-    return ["failed", "failed_submission", "failed_building", "completed"].includes(jobStatus);
```
krithika369, line 73: Shall we just reverse the name of the function to `isInactiveJobStatus` rather than change the list values? 😅 

Also, I think this function is defined in multiple places. I think we can move the definition to `ui/src/services/job/JobStatus.js` instead and re-use.

deadlycoconuts: > Shall we just reverse the name of the function to isInactiveJobStatus rather than change the list values? 😅 No preference on this eitherway though.

Ahaha I prefer to keep it this way; or at least to keep it consistent with Merlin 😅 

> Also, I think this function is defined in multiple places. I think we can move the definition to ui/src/services/job/JobStatus.js instead and re-use.

Ah good point, there's one more place where I forgot to refactor 👀

deadlycoconuts: Okay done, just made the change!

krithika369 (approved):
LGTM. Thanks for this fix!

PR #356, opened 2023-08-02
fix: Prevent calls to ensembling job service/controllers if batch ensembling is deactivated

deadlycoconuts:
## Context
When an ensembler gets deleted, checks are performed to see if there are any associated ensembling jobs that use it. However, these checks are unnecessary and should not be performed if batch ensembling has been deactivated in the first place. Doing so triggers bugs and other errors because the ensembling job service or controllers wouldn't be initialised during start up of the Turing API server. 

This PR thus introduces some additional checks to ensure that these checks are skipped if batch ensembling is turned off, so as to prevent such bugs:
- on the UI level: prevent the ensembler deletion pop up from sending any requests to the Turing API server to retrieve a list of related ensembling jobs if batch ensembling is turned off
image
- on the API level: prevent the delete ensembler handler from using the ensembling job service to retrieve ensembling jobs if batch ensembling is turned off

### DIFF
--- a/api/turing/api/ensemblers_api_test.go
+++ b/api/turing/api/ensemblers_api_test.go
@@ -912,7 +912,6 @@ func TestEnsemblerController_DeleteEnsembler(t *testing.T) {
 
 				return routerVersionSvc
 			},
-			ensemblingJobSvc: nil,
 			mlflowSvc: func() mlflow.Service {
 				mlflowSvc := &mlflowMock.Service{}
 				mlflowSvc.On("DeleteExperiment", mock.Anything, "1", true).Return(nil)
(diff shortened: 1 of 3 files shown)


api/turing/api/ensemblers_api_test.go
```diff
@@ -951,7 +981,7 @@ func TestEnsemblerController_DeleteEnsembler(t *testing.T) {
 				ensemblingJobSvc = tt.ensemblingJobSvc()
 			}
 			var routerVersionsSvc service.RouterVersionsService
-			if tt.ensemblingJobSvc != nil {
+			if tt.routerVersionsSvc != nil {
```
deadlycoconuts, line 984: Some typo that was left out previously; fixing this here.

ariefrahmansyah (approved):

deadlycoconuts:
Thanks for the quick review! Merging this now! 😅

PR #357, opened 2023-08-10
Add kaniko service account to image builder configs

deadlycoconuts:
## Context
Similar to how https://github.com/caraml-dev/merlin/pull/352 introduced a Kubernetes service account to be used for Kaniko jobs orchestrated by Merlin, this PR similarly introduces a similar change for Turing in its API server so that it is able to pass a configured Kaniko service account name to all the Kaniko jobs it orchestrates.

Note that not a lot of changes (if any) were made to the unit tests, despite the addition of a new service account (name) field, unlike in the Merlin repo. This is normal since Merlin is currently testing the image job creation workflows much more [extensively](https://github.com/caraml-dev/merlin/blob/da431b86a091b7966a65858f5961b432e8eeade6/api/pkg/imagebuilder/imagebuilder_test.go#L1294) in its unit tests, which isn't the case in Turing (we are only ascertaining that no error gets returned and the correct container image name gets returned). While it would be nice to introduce more extensive unit tests for Turing (so as to exhaustively test all the transformations performed in order to generate a K8s job manifest), this is out of scope of this PR and will probably be taken on at a later time.

### DIFF
--- a/api/turing/imagebuilder/imagebuilder.go
+++ b/api/turing/imagebuilder/imagebuilder.go
@@ -8,6 +8,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/caraml-dev/turing/api/turing/cluster"
+	"github.com/caraml-dev/turing/api/turing/config"
+	"github.com/caraml-dev/turing/api/turing/log"
+	"github.com/caraml-dev/turing/api/turing/models"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/google"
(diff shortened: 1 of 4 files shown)


api/turing/imagebuilder/imagebuilder.go
```diff
@@ -277,6 +276,34 @@ func (ib *imageBuilder) createKanikoJob(
 		annotations["cluster-autoscaler.kubernetes.io/safe-to-evict"] = "false"
 	}
 
+	var volumes []cluster.SecretVolume
+	var volumeMounts []cluster.VolumeMount
+	var envVars []cluster.Env
+
+	// If kaniko service account is not set, use kaniko secret
+	if ib.imageBuildingConfig.KanikoConfig.ServiceAccount == "" {
+		kanikoArgs = append(kanikoArgs,
```
krithika369, line 285: This is already set when `kanikoArgs` is initialised, that line should be removed since it's now being set conditionally.

krithika369 (approved):
Left one comment - the rest LGTM. Thanks, @deadlycoconuts !

api/turing/imagebuilder/imagebuilder.go
deadlycoconuts: Ah oh yes, nice catch 😅 Just removed it! :)

PR #358, opened 2023-08-15
fix: Add check before activating service accounts in Kaniko build jobs

deadlycoconuts:
## Context
With the changes in PR #357, there is a need to make additional changes to the Dockerfiles of the Kaniko image building templates - instead of authenticating immediately with `GOOGLE_APPLICATION_CREDENTIALS`, we now first check if that env var is defined, if so we will authenticate with the credentials stored in the file path pointed to by that env var; otherwise, the `gcloud auth activate-service-account` will not be run in the Dockerfile. In the second scenario, the Google service account credentials are expected to be passed to the Kaniko image building jobs via a mounted Kubernetes service account.

A similar change has been implemented in Merlin: https://github.com/caraml-dev/merlin/pull/352/files#diff-931d889c572814da3e253178b316471ac9878c45bf0497dd1792b5442eab359cR24

### DIFF
--- a/engines/pyfunc-ensembler-job/app.Dockerfile
+++ b/engines/pyfunc-ensembler-job/app.Dockerfile
@@ -4,7 +4,8 @@ FROM ${BASE_IMAGE}
 
 ARG MODEL_URL
 ARG GOOGLE_APPLICATION_CREDENTIALS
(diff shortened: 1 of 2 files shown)


krithika369 (approved):
Thanks for this fix! LGTM.

deadlycoconuts:
Thanks for the quick review! :D Merging this now!

PR #361, opened 2023-09-10
Remove multi-stage build from Pyfunc ensembler service

krithika369:
This PR upgrades the Kaniko image version used in the Turing infra chart.

With this Kaniko version, the Pyfunc ensembler service image builds require a lot more memory and often get OOMKilled (based on internal tests). Thus, to get around this, the multi-stage build is removed from the `Dockerfile` (which is the same approach as Merlin's Pyfunc models - [ref](https://github.com/caraml-dev/merlin/blob/v0.33.0/python/pyfunc-server/docker/Dockerfile)) to save build time and resources.

However, removing the multi-stage build does result in a larger built image size (about 500MB larger). Optimizing the image building process will be taken up separately.

### DIFF
--- a/engines/pyfunc-ensembler-service/Dockerfile
+++ b/engines/pyfunc-ensembler-service/Dockerfile
@@ -15,6 +15,3 @@ COPY ./temp-deps/sdk ./../../sdk
 
 RUN conda env create -f ./env-${PYTHON_VERSION}.yaml -n $CONDA_ENV_NAME &&  \
     rm -rf /root/.cache
-
-# Install conda-pack:
-RUN conda install conda-pack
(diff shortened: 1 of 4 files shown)


engines/pyfunc-ensembler-service/Dockerfile
```diff
@@ -15,6 +15,3 @@ COPY ./temp-deps/sdk ./../../sdk
 
 RUN conda env create -f ./env-${PYTHON_VERSION}.yaml -n $CONDA_ENV_NAME &&  \
     rm -rf /root/.cache
-
-# Install conda-pack:
-RUN conda install conda-pack
```
krithika369, line 20: Removing this from the Pyfunc ensembler base image as it will no longer be required.

krithika369:
@ariefrahmansyah @deadlycoconuts @tkpd-hafizhan, could I get a quick review on this? Thanks!

deadlycoconuts (approved):
LGTM! 🚀 (oh lol I actually clicked on submit review after the standup but I realised that tab never managed to load :/)

krithika369:
> LGTM! 🚀 (oh lol I actually clicked on submit review after the standup but I realised that tab never managed to load :/)

Haha cool. Thanks for the quick review, @deadlycoconuts !

PR #362, opened 2023-09-13
Remove Python 3.7 support in turing-sdk

krithika369:
Python 3.7 has reached its end of life ([ref](https://devguide.python.org/versions/)). This MR removes support for Python 3.7 from the SDK. For the time being the Pyfunc ensembler service and batch ensembler job apps will still support Python 3.7 - this is achieved by pinning their dependency on the `turing-sdk`. A future MR will also remove support for Python 3.7 from the ensembler service and batch jobs.

## Summary of changes
* `sdk/setup.py` - Bump up Python requirement to a minimum of `3.8`
* `engines/pyfunc-ensembler-service/requirements.txt`, `engines/pyfunc-ensembler-job/requirements.txt` - Pin `turing-sdk` dependency at `0.11.0` which is the latest SDK release that supports Python 3.7
* Update documentation and samples which use Python 3.7 to a different version

### DIFF
--- a/.github/workflows/pyfunc-ensembler-job.yaml
+++ b/.github/workflows/pyfunc-ensembler-job.yaml
@@ -30,6 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # TODO: Remove support for Python 3.7
         python-version: ["3.7", "3.8", "3.9", "3.10"]
     
     steps:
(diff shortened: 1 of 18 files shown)


ariefrahmansyah (approved):
LGTM! Thanks Krithika!

krithika369:
Thanks for the quick review, @ariefrahmansyah ! Merging.

PR #364, opened 2023-10-02
fix: Change delete ensembler api response from int to struct

deadlycoconuts:
## Context
The delete ensembler API endpoint currently returns an `int` (representing the id of the deleted ensembler) as part of a 200 response if the deletion occurs successfully. This is inconsistent with the specification in the Turing API server's OpenAPI specs which specifies a struct with the schema `{"id": [ensembler id]}` (this itself is consistent with the delete router API endpoint responses). See https://github.com/caraml-dev/turing/blob/main/api/api/specs/ensemblers.yaml#L147 and https://github.com/caraml-dev/turing/blob/main/api/api/specs/common.yaml#L14 for the exact schema definitions.

As such, using the Turing SDK to delete ensemblers would throw errors as the autogenerated OpenAPI classes would expect a struct in the response body instead of an `int`. This PR thus fixes the response returned by the API server when this endpoint is used.

### DIFF
--- a/api/turing/api/ensemblers_api.go
+++ b/api/turing/api/ensemblers_api.go
@@ -213,7 +213,7 @@ func (c EnsemblersController) DeleteEnsembler(
 		return Error(httpStatus, "failed to delete the ensembler", err.Error())
 	}
 
-	return Ok(ensembler.GetID())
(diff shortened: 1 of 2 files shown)


krithika369 (approved):
Thanks for this fix, Zi Yi! LGTM.

PR #368, opened 2024-01-30
fix(image-builder):Refactor ensembler image naming convention

deadlycoconuts:
## Context
This fix/enhancement revises the image naming convention for both ensembler service and ensembler job images built by the image builder:

### Before
```
{dockerRegistry}/{projectName}-{ensemblerName}-{runID}-service:{ensemblerID}
```
or
```
{dockerRegistry}/{projectName}-{ensemblerName}-{runID}-job:{ensemblerID}
```

### After
```
{dockerRegistry}/{projectName}/ensembler-services/{ensemblerName}:{runID}
```
or 
```
{dockerRegistry}/{projectName}/ensembler-jobs/{ensemblerName}:{runID}
```

By removing the `runID` from the image name, we now allow all images built for the same ensembler to be stored within the same image repository as opposed to storing them in separate repositories. Furthermore, since all built images of the same ensembler will now be stored in the same repository, we now tag them with their specific `runID`-s in order to distinguish between different versions of the same ensembler.

**_This PR contains changes similar to what is implemented for Merlin in https://github.com/caraml-dev/merlin/pull/526._** 

## Modifications
- `api/turing/imagebuilder/ensembler.go` - Change in template string for the ensembler image name
- `api/turing/imagebuilder/imagebuilder.go` - Change of image tag from `ensemblerID` to `runID`

### DIFF
--- a/api/turing/imagebuilder/ensembler.go
+++ b/api/turing/imagebuilder/ensembler.go
@@ -39,11 +39,8 @@ func (n *ensemblerJobNameGenerator) generateBuilderName(
 }
 
 // generateDockerImageName generate the name of docker image of prediction job that will be created from given model
-func (n *ensemblerJobNameGenerator) generateDockerImageName(projectName string,
-	modelName string,
-	runID string,
-) string {
-	return fmt.Sprintf("%s/%s-%s-%s-job", n.registry, projectName, modelName, runID)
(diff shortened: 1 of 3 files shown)


leonlnj (approved):
thanks!

PR #370, opened 2024-03-06
feat(api): Replace fluentd deployment with statefulset

deadlycoconuts:
## Context
In order to make the Fluentd deployment used by a Turing router deployment (to flush logs to BigQuery) more robust to cluster autoscaling events, which may remove the node on which the one and only Fluentd pod is scheduled and cause the Turing routers to return errors, we need to increase the number of replicas of the Fluentd deployment from **1 to 2**.

However, doing so involves replacing how Turing deploys the Fluentd deployment - from a Kubernetes Deployment resource to a StatefulSet. This is needed because deployments do not allow its pod replicas to specify volume/volume mount names that are different. This is crucial because the PVC that the current Fluentd pod uses is of the ReadWriteOnce type, which means that Fluentd replicas need to be scheduled on the same node in order to read/write on the same PVC. Other access mode exist - such as ReadOnlyMany and ReadWriteMany. However, ReadOnlyMany wouldn't allow Fluentd to write anything to the PVC, completely voiding the original purpose of even attaching a PVC to Fluentd, while ReadWriteMany isn't available in all managed Kubernetes services. 

By replacing the Deployment with a StatefulSet, we are able to specify a generic name for any of the PVCs that will be spun up in tandem with the number of pods the StatefulSet controls. This allows us to increase the number of Fluentd pods to 2, which allows for better uptime of the Fluentd service, especially when we introduce other configurations such as PodDisruptionBudgets.

As we are increasing the number of Fluentd pods, we are also halving the amount of CPU and memory resources requested for each pod to keep the total amount of resources consumed the same as before.

## Additional Changes
This PR also includes a tiny change to the pod disruption budget selector key used for Knative services (Turing routers, enrichers and ensemblers) from `app` to `serving.knative.dev/service`. This is done because the label value corresponding to the `app` key on these Knative service-managed pods have an additional suffix (e.g. `-0`) that indicates the revision number of the deployment. 

While this isn't a problem in most instances, any manual edits made to the Knative service manifest would undoubtedly cause the revision number to be bumped (e.g. from `0` to `1`), making the existing pod disruption budget resources created useless. This isn't the case however, for the label value corresponding to the `serving.knative.dev/service` key, which does not have the additional revision number suffix.

## Main Modifications
- `api/turing/cluster/controller.go` - Refactoring of the steps to deploy Fluentd as a stateful set instead of a deployment
- `api/turing/cluster/kubernetes_service.go` - Refactoring of the Fluentd deployment spec to a statefulset spec
- `api/turing/cluster/servicebuilder/fluentd.go` - Halving of Fluentd CPU and memory resources for each pod 
- `api/turing/cluster/servicebuilder/service_builder.go` - Addition of steps to configure PDBs differently for Fluentd
- `api/turing/service/router_deployment_service.go` - Removal of steps needed to manually create/delete PVCs


api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -353,18 +353,23 @@ func (sb *clusterSvcBuilder) NewSecret(
...
 func (sb *clusterSvcBuilder) NewPodDisruptionBudget(
 	routerVersion *models.RouterVersion,
 	project *mlp.Project,
 	componentType string,
 	pdbConfig config.PodDisruptionBudgetConfig,
 ) *cluster.PodDisruptionBudget {
+	var appNameSuffix string
+	if componentType != ComponentTypes.FluentdLogger {
+		appNameSuffix = "-0"
+	}
```
deadlycoconuts, line 366: This is needed because the Fluentd stateful set doesn't have a revision number as the suffix in its app name, which is the case for Knative apps (Turing router/enricher/ensembler).

api/turing/cluster/servicebuilder/fluentd.go
```diff
@@ -125,17 +125,8 @@ func buildFluentdVolumes(
...
 	})
 
-	volumes = append(volumes, corev1.Volume{
-		Name: ComponentTypes.CacheVolume,
-		VolumeSource: corev1.VolumeSource{
-			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-				ClaimName: cacheVolumePVCName,
-			},
-		},
-	})
```
deadlycoconuts, line 135: This is removed as the volumes get created automatically by the stateful set resource.

leonlnj: how does the stateful set "reuse" a previous volume then?

api/turing/cluster/controller.go
```diff
@@ -517,44 +517,23 @@ func (c *controller) GetKnativeServiceURL(ctx context.Context, svcName string, n
 	return url
 }
 
-// ApplyPersistentVolumeClaim creates a PVC in the given namespace.
-// If the PVC already exists, it will update the existing PVC.
-func (c *controller) ApplyPersistentVolumeClaim(
+// DeleteStatefulSetPersistentVolumeClaims deletes all PVCs belonging to the specified stateful set in the given
+// namespace.
+func (c *controller) DeleteStatefulSetPersistentVolumeClaims(
```
leonlnj, line 522: nit: could be more generic as `DeletePVC` and take in `listOptions` instead of `statefulSetName`

api/turing/cluster/servicebuilder/fluentd.go
leonlnj: nvm, on second thought, the volume are always deleted with statefulset

api/turing/service/router_deployment_service.go
```diff
@@ -563,27 +556,13 @@ func deleteSecret(controller cluster.Controller, secret *cluster.Secret, isClean
...
-func deletePVC(
+func deleteStatefulSetPVCs(
 	controller cluster.Controller,
 	namespace string,
-	pvc *cluster.PersistentVolumeClaim,
+	statefulSetName string,
 	isCleanUp bool,
 ) error {
-	return controller.DeletePersistentVolumeClaim(context.Background(), pvc.Name, namespace, isCleanUp)
+	return controller.DeleteStatefulSetPersistentVolumeClaims(context.Background(), statefulSetName, namespace, isCleanUp)
```
leonlnj, line 565: nit: would it be better to just use `ignoreNotFound` as per signature instead of `isCleanUp`?

api/turing/cluster/servicebuilder/service_builder.go
leonlnj: this got me thinking if the others should **not** be suffix with -0. During any patching, the suffix will not stay as `-0`and the pdb selector will not pick up anything

should we be using other key instead of `app` that will be consistent for the deployment even after patching?
something like 
`serving.knative.dev/service=abc-turing-ensembler-9` 
`serving.knative.dev/service=abc-turing-router-9` 
 `app=abc-turing-fluentd-logger-9`
 
 @ariefrahmansyah  any thoughts?

deadlycoconuts: Hmm, seems like we can take away the revision number for the label that has `app` as the key i.e. use `app=abc-turing-fluentd-logger` instead (the other labels can remain the same).

api/turing/cluster/controller.go
```diff
@@ -347,34 +347,34 @@ func (c *controller) GetKnativeServiceDesiredReplicas(
...
 	svcConf *KubernetesService,
 ) error {
-	desiredDeployment, desiredSvc := svcConf.BuildKubernetesServiceConfig()
+	desiredStatefulSet, desiredSvc := svcConf.BuildKubernetesServiceConfig()
 
-	// Deploy deployment
-	deployments := c.k8sAppsClient.Deployments(svcConf.Namespace)
+	// Deploy stateful set
+	statefulSets := c.k8sAppsClient.StatefulSets(svcConf.Namespace)
 	// Check if deployment already exists. If exists, update it. If not, create.
```
ariefrahmansyah, line 359: nit, `Check if statefulset ...`

api/turing/cluster/servicebuilder/service_builder.go
ariefrahmansyah: I just checked that `app` and `serving.knative.dev/revision` labels also have `-0` suffix for Turing router. But you can use `serving.knative.dev/service` as this one does not have the suffix. It should be fine using other label(s) that does not have `-0` suffix.

api/turing/cluster/servicebuilder/fluentd.go
```diff
@@ -77,7 +77,7 @@ func (sb *clusterSvcBuilder) NewFluentdService(
 			Volumes:               volumes,
 			VolumeMounts:          volumeMounts,
 		},
-		Replicas: fluentdReplicaCount,
+		Replicas: FluentdReplicaCount,
```
leonlnj, line 80: where is the flush rate configured?

api/turing/cluster/servicebuilder/fluentd.go
deadlycoconuts: Ah do you mean the PVCs? That's not done automatically when the stateful set is deleted, which is why there's this other function `DeleteStatefulSetPersistentVolumeClaims` that you commented on below.

api/turing/cluster/controller.go
deadlycoconuts: Alright sounds good; will refactor this.

api/turing/service/router_deployment_service.go
deadlycoconuts: `isCleanUp` is like the standard name of the such a `ignoreNotFound` flag passed around everywhere by the router deployment service though, so I just used the same name to keep things the same but I can refactor all of them too to use `ignoreNotFound`.

api/turing/cluster/controller.go
deadlycoconuts: Wow nice catch 😱

api/turing/cluster/servicebuilder/fluentd.go
deadlycoconuts: It's configured here: https://github.com/caraml-dev/turing/blob/40df8e9bb65849f396fc394c8f60f845571f4e59/api/turing/cluster/servicebuilder/fluentd.go#L43 but I think I initially didn't multiply the interval by the number of replicas because I thought it'd be better to have it configured manually in the API configs but now I've decided that it's probably simpler to have it configured automatically. One of my newer commits should have that value multiplied by the number of replicas.

api/turing/service/router_deployment_service.go
leonlnj: nah no strong opinion

deadlycoconuts: Okay then I'll leave it as is then.

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -353,19 +353,22 @@ func (sb *clusterSvcBuilder) NewSecret(
...
+// for the given service (router/enricher/ensembler/fluentd logger).
 func (sb *clusterSvcBuilder) NewPodDisruptionBudget(
 	routerVersion *models.RouterVersion,
 	project *mlp.Project,
 	componentType string,
 	pdbConfig config.PodDisruptionBudgetConfig,
 ) *cluster.PodDisruptionBudget {
+	var matchLabelKey string
+	if componentType != ComponentTypes.FluentdLogger {
+		matchLabelKey = cluster.KnativeServiceLabelKey
```
leonlnj, line 365: oh i didnt know we were already using this all along elsewhere. good to know

leonlnj (approved):
thanks, lgtm!

ariefrahmansyah (approved):

deadlycoconuts:
Thanks for the reviews everyone! Merging this now! 🚀

api/turing/cluster/servicebuilder/service_builder.go
deadlycoconuts: Thanks for the suggestion! :D Added that to the changes in this PR!

PR #371, opened 2024-03-12
chore(api): Remove authz middleware and update Turing API server, router, plugin dependencies

deadlycoconuts:
## Context
Similar to https://github.com/caraml-dev/merlin/pull/551, this MR basically bumps a couple of Go package dependencies used by the Turing API server, router, plugin, etc. The version of Go used in these components have also been bumped up from 1.20 to 1.22.

In addition to the general dependency upgrades, the authz middleware has been removed from the Turing API server entirely as it is no longer in use.

Main packages updated:
```
- github.com/GoogleCloudPlatform/spark-on-k8s-operator
- github.com/caraml-dev/mlp
- github.com/google/go-containerregistry 
- github.com/onsi/ginkgo/v2
- github.com/onsi/gomega
- k8s.io/api 
- k8s.io/apimachinery
- k8s.io/client-go
- knative.dev/networking
- knative.dev/pkg
- knative.dev/serving
- sigs.k8s.io/controller-runtime
```

Some minor refactoring was performed on structs that may have changed or on functions called due to the changes in their expected arguments - these are annotated directly on the diffs.


.github/workflows/turing.yaml
```diff
@@ -39,8 +39,8 @@ on:
 
 env:
   ARTIFACT_RETENTION_DAYS: 7
-  GO_VERSION: "1.20"
-  GO_LINT_VERSION: v1.51.2
+  GO_VERSION: "1.22"
+  GO_LINT_VERSION: v1.56.2
```
deadlycoconuts, line 43: Similar to what was explained in this comment [here](https://github.com/caraml-dev/merlin/pull/551/files#r1522415541), we need to use newer versions of the Go linter that actually supports Go 1.22, otherwise we'd run in OOM issues.

.github/workflows/turing.yaml
```diff
@@ -58,6 +58,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      # This is needed due to the issue raised here: https://github.com/golang/go/issues/61455
+      - name: Install Go ${{ env.GO_VERSION }}
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: false
```
deadlycoconuts, line 66: Not having this step causes the following errors to be thrown when `go mod vendor` is run in the CICD runners:
```
go: downloading go1.22 (linux/amd64)
go: download go1.22 for linux/amd64: toolchain not available
```
This issue is mentioned in https://github.com/golang/go/issues/61455, and one of the workarounds is to simply explicitly install go 1.22 https://github.com/shogo82148/actions-goveralls/issues/485.

api/Makefile
```diff
@@ -40,7 +40,7 @@ vendor:
 .PHONY: test
 test: tidy
 	@echo "Running tests..."
-	gotest -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
+	gotest -gcflags=-l -race -short -cover -coverprofile cover.out ${SRC_ROOT}/... -tags integration
```
deadlycoconuts, line 43: This new flag is needed because the package `bou.ke/monkey` we're using to patch certain functions called in our e2e tests is not able to properly patch certain functions. Essentially this flag turns off [inlining](https://dave.cheney.net/2020/05/02/mid-stack-inlining-in-go), which is what's recommended by the package's writer [here](https://github.com/bouk/monkey?tab=readme-ov-file#notes).

While we're on this, I found out that this patching package is apparently super old and there are probably newer alternatives that we might consider in the future 🤔

api/e2e/test/router_e2e_test.go
```diff
@@ -6,8 +6,8 @@ import (
 	"net/http"
 
 	"github.com/gavv/httpexpect/v2"
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
```
deadlycoconuts, line 10: Removed these dot imports because the linter complains about them.

api/e2e/test/cluster/cluster.go
```diff
@@ -41,7 +41,7 @@ var clients = &struct {
 type TuringRouterResources struct {
 	KnativeServices []knservingv1.Service
 	K8sServices     []coreV1.Service
-	IstioServices   []v1alpha3.VirtualService
+	IstioServices   []*v1alpha3.VirtualService
```
deadlycoconuts, line 44: This is now changed to a list of pointers because the new Istio Go client returns a list of pointers instead of `VirtualService` objects - see https://github.com/caraml-dev/turing/blob/9823c2d2fda90e1c21dbc2c8bf829145e0c97498/api/e2e/test/cluster/cluster.go#L154.

api/turing/api/appcontext.go
```diff
@@ -16,14 +16,12 @@ import (
...
 	"github.com/caraml-dev/turing/api/turing/config"
 	"github.com/caraml-dev/turing/api/turing/imagebuilder"
-	"github.com/caraml-dev/turing/api/turing/middleware"
 	"github.com/caraml-dev/turing/api/turing/service"
 	"github.com/caraml-dev/turing/engines/router/missionctl/errors"
 )
 
 // AppContext stores the entities relating to the application's context
 type AppContext struct {
-	Authorizer *middleware.Authorizer
```
deadlycoconuts, line 26: This is removed because the Turing API server no longer uses the Authorization middleware layer.

api/turing/cluster/knative_service_test.go
```diff
@@ -135,7 +135,7 @@ func TestBuildKnativeServiceConfig(t *testing.T) {
 				},
 				Resources: resources,
 				LivenessProbe: &corev1.Probe{
-					Handler: corev1.Handler{
+					ProbeHandler: corev1.ProbeHandler{
```
deadlycoconuts, line 138: Similar to https://github.com/caraml-dev/merlin/pull/551/files#r1521016971, this field has been renamed.

api/turing/cluster/pvc.go
```diff
@@ -29,7 +29,7 @@ func (pvc *PersistentVolumeClaim) BuildPersistentVolumeClaim() *corev1.Persisten
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
 			AccessModes: accessModes,
-			Resources: corev1.ResourceRequirements{
+			Resources: corev1.VolumeResourceRequirements{
```
deadlycoconuts, line 32: The name of this struct has been renamed in newer versions of the k8s Go client.

api/turing/cluster/virtual_service.go
```diff
@@ -29,16 +29,20 @@ func (cfg VirtualService) BuildVirtualService() *v1beta1.VirtualService {
...
-	httpMatches := make([]*networking.HTTPMatchRequest, len(cfg.MatchURIPrefixes))
-	for index, prefix := range cfg.MatchURIPrefixes {
-		uri := networking.HTTPMatchRequest{
-			Uri: &networking.StringMatch{
-				MatchType: &networking.StringMatch_Prefix{
-					Prefix: prefix,
+
+	var httpMatches []*networking.HTTPMatchRequest
+	if len(cfg.MatchURIPrefixes) > 0 {
+		httpMatches = make([]*networking.HTTPMatchRequest, len(cfg.MatchURIPrefixes))
```
deadlycoconuts, line 35: These additional lines are needed because newer versions of the Istio client specify `nil` in the `Spec.Http.[*].Match` field if there are no (`len(cfg.MatchURIPrefixes)==0`) http matches specified, instead of specifying a 0-length slice.

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -366,7 +366,8 @@ func getGCPSubDomains() []string {
 }
 
 func (ib *imageBuilder) checkIfImageExists(imageName string, imageTag string) (bool, error) {
-	keychain := authn.DefaultKeychain
+	var keychain authn.Keychain
+	keychain = authn.DefaultKeychain
```
deadlycoconuts, line 370: The reasoning behind this change is exactly the same as the one given in this comment: https://github.com/caraml-dev/merlin/pull/551#discussion_r1521027966

api/turing/service/mlp_service.go
```diff
@@ -48,7 +48,11 @@ type merlinClient struct {
 
 func newMerlinClient(googleClient *http.Client, basePath string) *merlinClient {
 	cfg := merlin.NewConfiguration()
-	cfg.BasePath = basePath
+	cfg.Servers = merlin.ServerConfigurations{
+		{
+			URL: basePath,
+		},
+	}
```
deadlycoconuts, line 55: These changes reflect the changes to the Merlin API server client (autogenerated using the OpenAPI generator) in the newer versions of Merlin.

api/turing/service/mlp_service.go
```diff
@@ -229,7 +233,7 @@ func (service mlpService) GetEnvironments() ([]merlin.Environment, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), mlpQueryTimeoutSeconds*time.Second)
 	defer cancel()
 
-	environments, resp, err := service.merlinClient.api.EnvironmentApi.EnvironmentsGet(ctx, nil)
+	environments, resp, err := service.merlinClient.api.EnvironmentAPI.EnvironmentsGet(ctx).Execute()
```
deadlycoconuts, line 236: `EnvironmentsGet` no longer returns the outputs of the API call directly - see https://github.com/caraml-dev/merlin/blob/main/api/client/api_environment.go#L47. We now need to call `Execute()` on the returned object - see https://github.com/caraml-dev/merlin/blob/main/api/client/api_environment.go#L35.

engines/router/missionctl/internal/safe_chan.go
```diff
@@ -63,11 +63,6 @@ func (sc *SafeChan) WriteAsync(data interface{}) {
 func (sc *SafeChan) Close() {
 	close(sc.closingCh)
 
-	go func() {
-		for range sc.ch {
-		}
-	}()
-
```
deadlycoconuts, line 70: Removed this since the linter complained that this go loop isn't doing anything.

api/turing/middleware/openapi_validation.go
```diff
@@ -105,3 +106,16 @@ func (openapi *OpenAPIValidation) Middleware(next http.Handler) http.Handler {
 	}
 	return handlerFunc
 }
+
+func jsonError(w http.ResponseWriter, msg string, status int) {
```
deadlycoconuts, line 110: This function is defined here since the file in which it was previously defined (`api/turing/middleware/authorization.go`) has been deleted.

api/turing/cluster/virtual_service_test.go
```diff
@@ -62,6 +60,8 @@ func TestBuildVirtualService(t *testing.T) {
 		},
 	}
 	got := cfg.BuildVirtualService()
-	err := tu.CompareObjects(expected, *got)
-	assert.NilError(t, err)
+
+	assert.Equal(t, expected.ObjectMeta, got.ObjectMeta)
+	assert.Equal(t, expected.Spec.String(), got.Spec.String())
+	assert.Equal(t, expected.Status.String(), got.Status.String())
```
deadlycoconuts, line 66: We cannot compare `expected` and `got` directly since `VirtualService`-s returned by the Istio client now contain unexported fields (e.g. `_.Status.state`) which makes `CompareObjects` unusable, as well as mutex locks. In order to work around this, we can compare the string generated from the `VirtualService`'s spec and status that prints out most of the fields in these fields.

The assert package has also been changed to `github.com/stretchr/testify/assert` to make the testing packages consistent with what is found in `api/turing/cluster/controller_test.go`. The previous `gotest.tools/assert` package was also unable to properly compare the `ObjectMeta` fields:

```
panic: runtime error: comparing uncomparable type v1.ObjectMeta
```

api/turing/cluster/controller_test.go
```diff
@@ -1737,8 +1738,19 @@ func TestApplyIstioVirtualService(t *testing.T) {
...
+						actualAction, ok := action.(k8stesting.CreateAction)
+						assert.True(t, ok)
+						actual, ok := actualAction.GetObject().(*v1beta1.VirtualService)
+						assert.True(t, ok)
+						expected, ok := expAction.GetObject().(*v1beta1.VirtualService)
+						assert.True(t, ok)
+
+						assert.Equal(t, expected.ObjectMeta, actual.ObjectMeta)
+						assert.Equal(t, expected.Spec.String(), actual.Spec.String())
+						assert.Equal(t, expected.Status.String(), actual.Status.String())
```
deadlycoconuts, line 1753: Similar to the comment [here](https://github.com/caraml-dev/turing/pull/371/files#r1555448202), we cannot compare expected and got directly since VirtualService-s returned by the Istio client now contain unexported fields as well as mutex locks. In order to work around this, we can compare the string generated from the VirtualService's spec and status that prints out most of the fields in these fields.

api/turing/service/mlp_service_test.go
```diff
@@ -67,40 +66,37 @@ func TestNewMLPService(t *testing.T) {
...
-	require.NoError(t, err)
 	// Create test projects and environments
 	projects := []mlp.Project{{ID: 1}}
 	environments := []merlin.Environment{{Name: "dev"}}
 
 	// Patch new Merlin and MLP Client methods
 	defer monkey.UnpatchAll()
 	monkey.Patch(newMerlinClient,
-		func(googleClient *http.Client, basePath string) *merlinClient {
-			assert.Equal(t, gc, googleClient)
```
deadlycoconuts, line 81: I removed this comparison because it is now failing due to the differences in the `googleClient.Transport` field, which may now be different, unsurprisingly because these two clients are initialised separately. It isn't immediately clear what we're comparing by having this check too so I just had it removed.

api/Makefile
leonlnj: its archived since 2020 😓

api/e2e/test/router_test.go
```diff
@@ -7,8 +7,8 @@ import (
 
 	upiv1 "github.com/caraml-dev/universal-prediction-interface/gen/go/grpc/caraml/upi/v1"
 	"github.com/gavv/httpexpect/v2"
-	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
+	"github.com/onsi/ginkgo/v2"
```
leonlnj, line 10: curious whats the motivation

api/turing/api/appcontext.go
leonlnj: this was used before the Keto upgrade?

api/turing/cluster/controller.go
```diff
@@ -680,7 +680,7 @@ func (c *controller) DeleteSparkApplication(ctx context.Context, namespace, appN
 
 func (c *controller) ApplyPodDisruptionBudget(
 	ctx context.Context,
-	namespace string,
+	_ string,
```
leonlnj, line 683: is namespace still required even? can it be remove from the method arg

leonlnj (approved):
thanks for the heavy refactoring, left some minor comment but LGTM

api/e2e/test/router_test.go
deadlycoconuts: It's one of the messages returned from our linter that dot imports are bad (but it seems like not everyone [agrees](https://github.com/golang/lint/issues/179) with them); I can't remember why I would've changed this if it was just a warning and not an error - lemme try removing this commit and see if it still works.

api/turing/cluster/controller.go
deadlycoconuts: Yeah that's true, it's not really needed anymore in this case. I'll remove it then; thanks for catching!

leonlnj (approved):
thanks LGTM

api/turing/api/appcontext.go
deadlycoconuts: Yes that's right.

PR #375, opened 2024-04-22
feat: support array in experiment engine json

leonlnj:
# Description

Update to the `engines/experiment/pkg/request/request.go` to support object and array type.

The underlying method does support object and array by presumingly older requirements then, these type are not supported. However due to evolving experiment segments that involves array and dictionary, these would required to be updated

image

### DIFF
--- a/engines/experiment/pkg/request/request.go
+++ b/engines/experiment/pkg/request/request.go
@@ -115,7 +115,7 @@ func getValueFromJSONPayload(body []byte, key string) (string, error) {
 	value, typez, _, _ := jsonparser.Get(body, strings.Split(key, ".")...)
 
 	switch typez {
-	case jsonparser.String, jsonparser.Number, jsonparser.Boolean, jsonparser.Array:
+	case jsonparser.String, jsonparser.Number, jsonparser.Boolean, jsonparser.Array, jsonparser.Object:
 		// See: https://github.com/buger/jsonparser/blob/master/bytes_unsafe.go#L31
 		return *(*string)(unsafe.Pointer(&value)), nil
 	case jsonparser.Null:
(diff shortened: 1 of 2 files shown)


engines/experiment/pkg/request/request.go
```diff
@@ -60,7 +60,7 @@ func GetValueFromHTTPRequest(
 	case PayloadFieldSource:
 		return getValueFromJSONPayload(bodyBytes, field)
 	case HeaderFieldSource:
-		value := reqHeader.Get(field)
+		value := strings.Join(reqHeader.Values(field), ",")
```
leonlnj, line 63: This is required for array/object. This is backward compatible (as seen from unit test) if there is only 1 element

engines/experiment/pkg/request/request_test.go
```diff
@@ -44,6 +44,17 @@ func TestGetValueFromHTTPRequest(t *testing.T) {
...
+			field:    "CustomerID",
+			fieldSrc: request.HeaderFieldSource,
+			header: func() http.Header {
+				header := http.Header{}
+				header.Set("CustomerID", "123")
+				header.Add("CustomerID", "321")
+				return header
+			}(),
+			expected: "123,321",
+		},
```
deadlycoconuts, line 57: Just curious, how would the header look like if its corresponding value is an object? :o

deadlycoconuts (approved):
LGTM! 🚀 Thanks for this PR!

engines/experiment/pkg/request/request_test.go
leonlnj: it depends on the library and how you create the headers. It will just be a string nevertheless. Eg ```header.add(`{"a":"b"}``` will result in ```123,{"a","b"}```

PR #376, opened 2024-04-29
Ensembler Image Building API

ariefrahmansyah:
# Description

The image-building process of the Turing ensembler jobs and services are tightly coupled with the serving flow. If the image-building failed, the serving would fail too.

This PR introduces a new Image Building API to trigger the image building process programmatically and separately from the serving flow.

# Modifications

1. Add new Image Building API which calls the new Image Building Service layer
   a. Add ListImages endpoint that will return the job and services images for a given ensembler.
   b. Add BuildImage endpoint that will build an job or service image
2. Add Image Building Service layer which calls the imagebuilder/imagebuilder.go

# Tests

# Checklist
- [x] Added PR label
- [x] Added unit test, integration, and/or e2e tests
- [x] Tested locally
- [ ] Updated documentation
- [x] Update Swagger spec if the PR introduce API changes
- [x] Regenerated Golang and Python client if the PR introduces API changes

# Release Notes

```release-note
NONE
```


sdk/requirements.dev.txt
```diff
@@ -2,7 +2,8 @@ black==22.6.0
 setuptools>=21.0.0
 wheel
 twine
-pytest
+# The next release 8.2.0 of pytest breaks the unit tests
+pytest<=8.1.2
```
ariefrahmansyah, line 6: I ended up porting your changes here @deadlycoconuts

sdk/turing/version.py
```diff
@@ -1,2 +1 @@
-# Do not update this field manually; this value gets updated automically during build time
```
leonlnj, line 1: what the motivation for this change?

leonlnj:
LGTM, dont any other comments, will let ziyi take a look as well

sdk/turing/version.py
ariefrahmansyah: Hmm, I don't remember updating this. I think it is because I do pip install --editable and somehow it got overwritten. Will revert it.

api/api/openapi.bundle.yaml
```diff
@@ -1441,6 +1512,64 @@ components:
...
+        $ref: '#/components/schemas/EnsemblerImage'
+      type: array
+    EnsemblerImage:
+      example:
+        image_ref: image_ref
+        project_id: 0
+        image_building_job_status:
+          message: message
+        exists: true
+        ensembler_id: 6
```
deadlycoconuts, line 1526: Wow there's even an example generated. I wonder if this was generated by the openapitools form 🤔

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -486,3 +560,50 @@ func (ib *imageBuilder) DeleteImageBuildingJob(
...
+	if err != nil {
+		return EnsemblerImage{}, err
+	}
+
+	imageRef := fmt.Sprintf("%s:%s", imageName, ensembler.RunID)
+
+	image := EnsemblerImage{
+		ProjectID:           models.ID(project.ID),
+		EnsemblerID:         models.ID(ensembler.GetID()),
+		EnsemblerRunnerType: ib.runnerType,
```
deadlycoconuts, line 579: I can't tell whether it's cleaner to reuse the existing `ib.nameGenerator` field to perform the check `_, isJobType := ib.nameGenerator.(*ensemblerJobNameGenerator)` to find out the type (service/job) of the image builder **_or_**, to introduce a new field `runnerType` to the `imageBuilder` struct and then potentially remove all the `nameGenerator`-derived structs (ensemblerServiceNameGenerator/ensemblerJobNameGenerator) because they can use the new `runnerType` field to determine which naming conventions to use. 

Still sitting on the fence on this for now so what do you all think?

deadlycoconuts (approved):
LGTM! Thanks for this super long PR. I've left some comments which might no longer be relevant anymore after your latest pushes 😅

api/api/openapi.bundle.yaml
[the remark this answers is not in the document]
ariefrahmansyah: Yup, this was generated by open api tool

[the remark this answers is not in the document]
ariefrahmansyah: Yup, this one was generated automatically by open api tool

api/turing/imagebuilder/imagebuilder.go
ariefrahmansyah: I agree that there should be a better design pattern for this to simplify the nameGenerator and runnerType while at the same time make it consistent. However, I think we should revisit this in separate PR. I'm resolving this comment for now 🙏🏻

deadlycoconuts: Sounds good! Thanks!

PR #378, opened 2024-05-03
fix(turing): Add new autoscaling target processing for concurrency

deadlycoconuts:
## Context
While some changes were implemented in https://github.com/caraml-dev/merlin/pull/576 to allow concurrency autoscaling targets of Merlin models to be set below 1, the Turing API server does not require that many changes because it has already been allowing users to set float values for autoscaling targets. However, to keep the behaviour of both applications consistent, this PR will be introducing some simple additional float-value rounding as well as error handling.

1. Float values for concurrency targets (only) will be rounded to 2 decimal places
3. Float values for concurrency targets above below 0.01 after rounding will cause an error to be returned, so as to prevent the scenario whereby the Turing API server attempts deployment, only for it to fail after some timeout because the Knative admission webhook rejects targets below 0.01.


leonlnj (approved):
lgtm thanks!

deadlycoconuts:
Thanks for the quick review, merging this! 🚀

PR #379, opened 2024-05-08
fix(api): Refactor JobState to string

ariefrahmansyah:
The previous JobState is `int` and uses `iota` together in the same const block with other values. This lead to JobState is starting from 3 instead of 0.

Using `int` for `JobState` requires mapping implementation in SDK.

This PR refactors JobState from `int` to `string` so that it easier to parse by SDK and to make it consistent with Merlin implementation (https://github.com/caraml-dev/merlin/blob/main/api/models/version_image.go#L15-L20).


leonlnj (approved):
thanks!

PR #380, opened 2024-05-09
fix(api,ui): Fix autoscaling target parsing

deadlycoconuts:
## Context
This is a follow up on PR #378, which introduces additional fixes and some minor refactoring to make the UI and API server behaviour consistent with that of Merlin's, especially with regards to the handling of autoscaling target values.

In short, this PR does the following:
1. On the UI, removal of all integer parsing/verification of all autoscaling targets (this will now be handled in a more sophisticated manner in the API server; see below for more information)
2. On the API server, introduction of additional handling to parse CPU and RPS autoscaling targets to integers

### DIFF
--- a/ui/src/router/components/form/components/autoscaling_policy/AutoscalingPolicyPanel.js
+++ b/ui/src/router/components/form/components/autoscaling_policy/AutoscalingPolicyPanel.js
@@ -24,10 +24,10 @@ export const AutoscalingPolicyPanel = ({
   errors = {},
 }) => {
   const { onChange } = useOnChangeHandler(onChangeHandler);
-  // Parse the integer portion of the value
+  // Parse the float portion of the value
   const onChangeTarget = (value) => {
-    const parsedInt = parseInt(value);
-    onChange("target")(isNaN(parsedInt) ? "" : parsedInt.toString());
+    const parsedFloat = parseFloat(value);
+    onChange("target")(isNaN(value) ? "" : parsedFloat.toString());
   };
   // Update default target if the metric is changing
   const onChangeMetric = (value) => {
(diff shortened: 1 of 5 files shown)


ui/src/router/components/form/components/autoscaling_policy/AutoscalingPolicyPanel.js
```diff
@@ -93,8 +93,9 @@ export const AutoscalingPolicyPanel = ({
               onChange={(e) => onChangeTarget(e.target.value)}
               isInvalid={!!errors.target}
               name="memory"
-              min={1}
-              step={1}
+              // The min value is set as 0.005 because it's the smallest value, when rounded to 2 decimal places, gives
+              // 0.01, the smallest value accepted as an autoscaling target (concurrency).
+              min={0.005}
```
deadlycoconuts, line 98: The removal of the step size is to fix this issue whereby a red line appears when a value that is outside the configured step size is set:
Screenshot 2024-05-09 at 11 56 32 AM

The min value has simply been updated to allow values below 0 (e.g. 0.5) to be set, which are valid values when the concurrency autoscaling target is used.

ui/src/router/components/form/components/autoscaling_policy/AutoscalingPolicyPanel.js
```diff
@@ -93,8 +93,10 @@ export const AutoscalingPolicyPanel = ({
...
+              step={"any"}
```
deadlycoconuts, line 99: The setting the step size to `any` (the default value is `1`) is to fix this issue whereby a red line appears when a value that is outside the configured step size (e.g. 1.003 when the step size is 1) is set:
Screenshot 2024-05-09 at 11 56 32 AM

The min value has also been changed,  to allow values below 0 (e.g. 0.5) to be set, which are valid values when the concurrency autoscaling target is used.

ariefrahmansyah (approved):

deadlycoconuts:
Thanks for the quick review @ariefrahmansyah ! Merging this now (oops I totally forgot about this hanging PR for over 2 weeks 🥲)

PR #381, opened 2024-05-21
feat(api,ui,sdk): Make CPU limits configurable

deadlycoconuts:
## Context
Similar to https://github.com/caraml-dev/merlin/pull/586, this PR aims to make CPU limits configurable for the end user.

As of present, users are not able to configure the CPU limits of the pods in which Turing routers/enrichers/ensemblers (docker and pyfunc) are deployed in - they are instead determined automatically on the platform-level (Turing API server). Depending on how the API server has been configured, one of the following happens:
- the CPU limit of a component is set as its CPU request value, multiplied by a [scaling factor](https://github.com/caraml-dev/turing/blob/0d6f5780491390f68c407ab2529cd83d905858eb/api/turing/config/config.go#L266) (e.g. 2 CPU * 1.5) **or,**
    - Note that this is the existing way memory limits are automatically set by the Turing API server
- the CPU limit is left unset 

This PR introduces a new workflow which would allow users to instead override the platform-level CPU limits (described in the paragraph above) set on a component. This workflow is available via the UI, SDK and by extension, directly calling the API endpoint of the API server. 

UI:
Screenshot 2024-05-24 at 4 27 46 PM

Screenshot 2024-05-24 at 3 29 57 PM

SDK:
```python
config = router.config
config.resource_request = ResourceRequest(
    min_replica=1,
    max_replica=3,
    cpu_request="100m",
    cpu_limit="2",
    memory_request="512Mi",
)

router.update(config)
```

In addition, this PR adds a new configuration, `DefaultEnvVarsWithoutCPULimits`, which is a list of env vars that automatically get added to all Turing routers/enrichers/ensemblers (docker and pyfunc) when CPU limits are not set. This allows the Turing API server's operators to set env vars platform-wide that can potentially improve these deployments' performance, e.g. env vars involving concurrency. 

## Modifications
- `api/turing/cluster/knative_service.go` - Removal of platform-level fields from the `KnativeService` struct
- `api/turing/cluster/servicebuilder/service_builder.go` - Addition of platform-level configs to `clusterSvcBuilder` and new helper methods to set default env vars when cpu limits are not explicitly set and when the cpu limit scaling factor is set as 0
- `api/turing/config/config.go` - Addition of the new field `DefaultEnvVarsWithoutCPULimits`
- `sdk/turing/router/config/resource_request.py` - Addition of a new cpu limit field to the resource request class
- `ui/src/router/components/form/components/CPULimitsFormGroup.js` - Addition of a new form group to allow cpu limits to be specified on the UI


api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -128,6 +119,9 @@ type clusterSvcBuilder struct {
 	MaxMemory                 resource.Quantity
 	MaxAllowedReplica         int
 	TopologySpreadConstraints []corev1.TopologySpreadConstraint
+
+	// Knative service configs
+	knativeServiceConfig *config.KnativeServiceDefaults
```
deadlycoconuts, line 124: I've also refactored the `clusterSvcBuilder` struct a little to carry the `knativeServiceConfig` field, which in turn contains platform-level fields like the cpu/memory limit scaling factor. These are fields that are common to a single API server and do not need to be drilled all the way to each individual `KnativeService`.

api/turing/config/config.go
```diff
@@ -263,8 +263,9 @@ type KubernetesLabelConfigs struct {
 // Knative services
 type KnativeServiceDefaults struct {
 	QueueProxyResourcePercentage          int
-	UserContainerCPULimitRequestFactor    float64 `json:"userContainerLimitCPURequestFactor"`
-	UserContainerMemoryLimitRequestFactor float64 `json:"userContainerLimitMemoryRequestFactor"`
+	UserContainerCPULimitRequestFactor    float64
+	UserContainerMemoryLimitRequestFactor float64
+	DefaultEnvVarsWithoutCPULimits        []corev1.EnvVar
```
deadlycoconuts, line 268: I have no idea why we needed the `json` tags in the first place (these are the API server-level configs and aren't converted to json anywhere), so I'm removing them.

api/turing/internal/testutils/validation.go
```diff
@@ -16,21 +15,7 @@ func CompareObjects(actual interface{}, expected interface{}) error {
...
-			actualString = string(jsonActual)
-		}
-		jsonExpected, err := json.Marshal(expected)
-		if err == nil {
-			expectedString = string(jsonExpected)
-		}
-
-		return fmt.Errorf("Did not get expected configuration.\nEXPECTED:\n%v\nGOT:\n%v",
-			expectedString, actualString)
+		return fmt.Errorf(cmp.Diff(actual, expected, cmp.AllowUnexported(allowUnexportedOn)))
```
deadlycoconuts, line 18: I simplified all these steps because the error message it generates isn't precise/specific/useful enough (e.g. for pointing out the presence of null fields especially if those fields get omitted when converted to json) and replaced it with the output of the `Diff` method of the sample `cmp` package.

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -434,3 +460,23 @@ func buildLabels(
 	}
 	return labeller.BuildLabels(r)
 }
+
+// mergeEnvVars merges multiple sets of environment variables and return the merging result.
+// All the EnvVars passed as arguments will be not mutated.
+// EnvVars to the right have higher precedence.
+func mergeEnvVars(left []corev1.EnvVar, rightEnvVars ...[]corev1.EnvVar) []corev1.EnvVar {
```
deadlycoconuts, line 467: Added this helper function that came from https://github.com/caraml-dev/merlin in order to merge env vars with replacement (for those with the same name) instead of appending blindly (which can create duplicated env vars in the list).

api/turing/cluster/models.go
```diff
@@ -62,22 +64,19 @@ type BaseService struct {
...
-	UserContainerCPULimitRequestFactor float64,
-	UserContainerMemoryLimitRequestFactor float64,
-) corev1.ResourceRequirements {
+func (cfg *BaseService) buildResourceReqs() corev1.ResourceRequirements {
 	reqs := map[corev1.ResourceName]resource.Quantity{
 		corev1.ResourceCPU:    cfg.CPURequests,
 		corev1.ResourceMemory: cfg.MemoryRequests,
 	}
 
 	// Set resource limits to request * userContainerCPULimitRequestFactor or UserContainerMemoryLimitRequestFactor
```
leonlnj, line 73: this is invalid now right?

api/turing/cluster/models.go
```diff
@@ -62,22 +64,19 @@ type BaseService struct {
 	InitContainers []Container `json:"init_containers"`
 }
 
-func (cfg *BaseService) buildResourceReqs(
-	UserContainerCPULimitRequestFactor float64,
-	UserContainerMemoryLimitRequestFactor float64,
-) corev1.ResourceRequirements {
+func (cfg *BaseService) buildResourceReqs() corev1.ResourceRequirements {
```
leonlnj, line 67: can you help add some comment on this? feel like i will forget the order/flow of in the future

api/turing/cluster/servicebuilder/service_builder.go
```diff
@@ -414,6 +404,42 @@ func (sb *clusterSvcBuilder) getTopologySpreadConstraints() ([]corev1.TopologySp
 	return topologySpreadConstraints, nil
 }
 
+func (sb *clusterSvcBuilder) getCPULimit(resourceRequest *models.ResourceRequest) *resource.Quantity {
```
leonlnj, line 407: do you think this would be easier to read 
```
if resourceRequest == nil {
    return nil
}

# use user set limit as priority
if resourceRequest.CPULimit != nil && !resourceRequest.CPULimit.IsZero() {
    return resourceRequest.CPULimit
}

if sb.knativeServiceConfig.UserContainerCPULimitRequestFactor == 0 {
    return nil
}

# apply platform multiplier if limit is not set
cpuLimit := cluster.ComputeResource(resourceRequest.CPURequest, sb.knativeServiceConfig.UserContainerCPULimitRequestFactor)
return &cpuLimit
```

leonlnj (approved):
thanks LGTM. +1 for moving platform default out and putting it as part of the service builder

api/turing/cluster/models.go
deadlycoconuts: Ahh yes nice good catch

api/turing/cluster/models.go
deadlycoconuts: Alright just added something. Thanks!

PR #386, opened 2024-07-17
fix(ui): Revert upgrade to react-lazylog

deadlycoconuts:
## Context
For some still yet to be understood reason, replacing the forked version of `react-lazylog` from https://github.com/gojekfarm/react-lazylog#master with the latest official version of the same package caused the logs viewer to crash due to some styling issues that get triggered when we manually set the height of the logs viewer (or the component containing it):

Screenshot 2024-07-17 at 13 48 53

In the Turing UI, this height is set by the custom style in `PodLogsViewer.scss` (removed in this PR) while in the Merlin UI, it is set by passing it as a prop to the `LazyLog` component. This PR also standardises the way both UIs are setting this value.

Strangely enough, the Merlin UI isn't experiencing these styling issues even after having its `react-lazylog` replaced with the official reason but a deeper look into this will be done at a later time.

### DIFF
--- a/.github/workflows/turing.yaml
+++ b/.github/workflows/turing.yaml
@@ -180,7 +180,7 @@ jobs:
           cache-dependency-path: ui/yarn.lock
 
       - name: Install
-        run: yarn install --network-concurrency 1
+        run: NODE_OPTIONS=--openssl-legacy-provider yarn install --network-concurrency 1
 
       - name: Lint code
         run: yarn lint
--- a/ui/package.json
+++ b/ui/package.json
@@ -24,7 +24,7 @@
     "react-collapsed": "^4.1.2",
     "react-diff-viewer": "^3.1.1",
     "react-dom": "^18.3.1",
-    "react-lazylog": "^4.5.3",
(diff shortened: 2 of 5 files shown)


leonlnj (approved):

.github/workflows/turing.yaml
```diff
@@ -180,7 +180,7 @@ jobs:
           cache-dependency-path: ui/yarn.lock
 
       - name: Install
-        run: yarn install --network-concurrency 1
+        run: NODE_OPTIONS=--openssl-legacy-provider yarn install --network-concurrency 1
```
deadlycoconuts, line 183: Somehow I need an older SSL provider https://github.com/caraml-dev/turing/actions/runs/9969015761/job/27547281483#step:5:35, but I have no idea why since the dependencies in https://github.com/gojekfarm/react-lazylog/blob/master/package.json and in https://github.com/mozilla-frontend-infra/react-lazylog/blob/master/package.json are the same.

I simply followed one of the steps suggested in https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported to fix this.

PR #384, opened 2024-06-28
chore(ui): Upgrade UI dependencies

ariefrahmansyah:
## Context
Similar to what had been done in MLP https://github.com/caraml-dev/mlp/pull/100 and in Merlin https://github.com/caraml-dev/merlin/pull/591, this PR updates certain UI dependencies in Turing, most notably:
- node.js from v16 to v20 (an engines.node field is added in the `package.json` file to pin the engine to >=20.13.1)
- react from v17 to v18
- elastic UI from v66 to v88.2.0* 🚨 **this was intentionally left at this version because updating it to v88.3.0 breaks all the `EuiAccordion`-related components; this is likely due a massive rework of the component in this PR https://github.com/elastic/eui/pull/7161 but I've no idea why this change didn't affect the accordions in Merlin**

Other minor changes have also been made to the GitHub CICD files (more details are covered below and in the comments left on the diff itself).

## Main Changes
- Upgrading of UI dependencies
- Removal of the unused GitHub workflow to release the Turing Helm chart (this chart has already been moved to the [caraml-dev/helm-charts](https://github.com/caraml-dev/helm-charts/tree/main) repo a long time ago)
- Upgrading of multiple GitHub actions; see the subsection below for more information

### GitHub Actions Updates
- `actions/upload-artifact` - from v2 to v4 since the former has already been [deprecated](https://github.com/actions/upload-artifact?tab=readme-ov-file#actionsupload-artifact) since the 30th of June this year
- `actions/download-artifact` - from v2 to v4 since the former has already been [deprecated](https://github.com/actions/download-artifact?tab=readme-ov-file#actionsdownload-artifact) since the 30th of June this year (the versions of this action and `upload-artifact` need to be compatible to work)
- `actions/checkout` - from v3 to v4 since it's similarly done in the MLP repo
- `actions/setup-go` - from v4 to v5 since it's similarly done in the MLP repo (also provides caching)
- `actions/setup-python` - from v2 to v5 to keep it updated to the latest version (also provides caching)
- `docker/login-action` - from v1 to v3 just to keep it updated to the latest version
- `jupyterhub/action-k8s-await-workloads` - from v1 to v3 just to keep it updated to the latest version

PR #387, opened 2024-07-23
fix(ui): Fix yup validation bugs

deadlycoconuts:
## Context
In PR #384, the `yup` package had been updated from `0.29.1` to `1.4.0`. However, this huge upgrade had a lot of [breaking changes](https://github.com/jquense/yup/blob/master/CHANGELOG.md) (see the [latest](https://github.com/jquense/yup) README.md vs the [old](https://github.com/jquense/yup/tree/v0.29.1) one) and effectively broke most of the existing validation schemas in the Turing UI. This PR serves to fix those broken validation schemas - these changes can largely be grouped into the following:

- The `Schema.when` function has been [updated](https://github.com/jquense/yup?tab=readme-ov-file#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema), such that when the builder object (one with the fields `is`, `then`, and sometimes `otherwise` defined) is passed as an argument, the `then` field accepts ONLY functions of the following signature `(schema: Schema) => Schema` instead of `Schema` previously. 

  Existing schemas that do not follow this convention have been updated and turned into arrow functions. 
  
  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", {
    is: "nop",
    then: yup.string().required("this is needed")
  }),

  // after
  some_field: yup.string().when("some_other_field", {
    is: "nop",
    then: (_) => yup.string().required("this is needed")
  }),
  ```
- When using the same `Schema.when` function but passing a function directly as an argument instead of the builder object, the signature of the expected function has also been updated from `(value, schema)=> Schema): Schema` to `(values: any[], schema) => Schema): Schema`.

  Existing schemas that do not follow this convention now have been updated to reflect the new expected array:

  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", (some_other_field, schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),

  // after
  some_field: yup.string().when("some_other_field", ([some_other_field], schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),
  ```
- The `mixed` schema has been severely nerfed and is now no longer the base class for all the other schemas, and as such no longer supports a lot of the other functions that we were previously calling. Schemas that have fields which use the `mixed` schema have been updated to use a schema (e.g. `string`) corresponding to its expected primitive data type.

## Unrelated Update
This PR also bumps up the version of the `caraml-dev/ui-lib` library to the latest version.

### DIFF
--- a/ui/src/router/components/form/components/experiment_config/validation/schema.js
+++ b/ui/src/router/components/form/components/experiment_config/validation/schema.js
@@ -15,21 +15,17 @@ const experimentSchema = yup.object().shape({
 
 const variableSchema = yup.object().shape({
   required: yup.boolean(),
-  field_source: yup.mixed().when("required", {
+  field_source: yup.string().when("required", {
     is: true,
-    then: fieldSourceSchema.required(
-      "Select the field source for the required variable"
-    ),
+    then: fieldSourceSchema,
   }),
-  field: yup.mixed().when("required", {
+  field: yup.string().when("required", {
     is: true,
-    then: fieldSchema("field_source").required(
-      "Specify the field name for the required variable"
-    ),
+    then: fieldSchema("field_source"),
   }),
 });
 
-const standardExperimentConfigSchema = (engineProps) =>
+const standardExperimentConfigSchema = (engineProps) => (_) =>
   yup.object().shape({
     client: engineProps?.standard_experiment_manager_config
       ?.client_selection_enabled
--- a/ui/src/router/components/form/validation/schema.js
+++ b/ui/src/router/components/form/validation/schema.js
@@ -130,7 +130,7 @@ const routeSchema = yup.object().shape({
 });
 
 const validRouteSchema = yup
-  .mixed()
+  .object()
   .test("valid-route", "Valid route is required", function(value) {
     const configSchema = this.from.slice(-1).pop();
     const { routes } = configSchema.value.config;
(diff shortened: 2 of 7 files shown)


ui/src/router/components/form/validation/schema.js
```diff
@@ -364,10 +365,9 @@ const schema = (maxAllowedReplica) => [
         .test("unique-rule-names", validateRuleNames),
       routes: yup
         .array(routeSchema)
-        .required()
```
deadlycoconuts, line 367: `Array.required` has been deprecated and the same functionality is performed by `min(1, ...)`.

ui/src/router/components/form/components/experiment_config/validation/schema.js
```diff
@@ -15,21 +15,17 @@ const experimentSchema = yup.object().shape({
...
-    ),
+    then: fieldSourceSchema,
   }),
-  field: yup.mixed().when("required", {
+  field: yup.string().when("required", {
     is: true,
-    then: fieldSchema("field_source").required(
-      "Specify the field name for the required variable"
-    ),
+    then: fieldSchema("field_source"),
```
deadlycoconuts, line 24: These `required` conditions were removed because we can no longer call `required` on a function (`fieldSchema("field_source")` is a function). This isn't a problem because the schema returned by `fieldSchema("field_source")` itself contains very similar error prompts:

leonlnj (approved):

deadlycoconuts:
Thanks for the quick approval! Merging this after the pipeline passed (I temporarily removed codecov for the API server since it's constantly failing the past PRs even though there are no API server related changes; the threshold doesn't seem to work either which is weird)! 🚀

PR #390, opened 2024-08-02
fix(ui): Replace deprecated overgear/yup-ast package with demvsystems/yup-ast

deadlycoconuts:
## Context
It seems like after upgrading yup from `0.29.1` to `1.4.0`, we have broken the its compatibility with the `@overgear/yup-ast` [package](https://github.com/WASD-Team/yup-ast). This package is ancient; it's last published 6 years ago and no longer receives any updates because the original contributors have left the company and do not have any permissions/licenses to continue contributing to them. Their [replacement](https://github.com/WASD-Team/yup-ast/issues/10#issuecomment-610873654) is the `@demvsystems/yup-ast` [package](https://github.com/demvsystems/yup-ast), which works _almost_ exactly the same as before.

This package is used when parsing the `yup`-like schema that's served over an API server endpoint (see https://github.com/caraml-dev/xp/blob/cb06edaaf409077b5abf7fed0b803984b6decf7e/plugins/turing/manager/experiment_manager.go#L38C1-L38C36). We only use this package in one place, when parsing the `yup`-like schema off the experiment engines endpoint on the Turing API server (this in turn gets its value from the XP experiment manager plugin, which explains why the file quoted above is found in the XP repository and not in Turing). This parsed schema is then used to validate the XP experiment engine config together with the other `yup` schemas of the Turing router config.

More concretely, this broken package is currently causing this problem whereby the `yup` validation that gets triggered when a user clicks on the 'Next' page of an accordion form, to end up permanently stuck:
image-20240801-142115

A separate PR in XP has also been created to fix a related bug in the `yup`-like schema quoted above: https://github.com/caraml-dev/xp/pull/81

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -12,7 +12,7 @@
     "@elastic/eui": "88.2.0",
     "@emotion/css": "^11.11.2",
     "@emotion/react": "^11.11.4",
-    "@overgear/yup-ast": "^1.0.3",
+    "@demvsystems/yup-ast": "^1.2.2",
     "@sentry/browser": "^7.118.0",
     "js-yaml": "^4.1.0",
     "mitt": "^3.0.1",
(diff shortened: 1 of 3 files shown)


tiopramayudi (approved):
LGTM ! Thanks @deadlycoconuts

deadlycoconuts:
Thanks a lot for the quick review @tiopramayudi ! Merging this now!

PR #391, opened 2024-08-07
fix(ui): Fix min replica yup validation schema

deadlycoconuts:
## Context 
This PR introduces an additional fix to the yup validation schema, as a follow up from #387. In particular, the fix in this PR addresses this change:

- When using the same `Schema.when` function but passing a function directly as an argument instead of the builder object, the signature of the expected function has also been updated from `(value, schema)=> Schema): Schema` to `(values: any[], schema) => Schema): Schema`.

  Existing schemas that do not follow this convention now have been updated to reflect the new expected array:

  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", (some_other_field, schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),

  // after
  some_field: yup.string().when("some_other_field", ([some_other_field], schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),

### DIFF
--- a/ui/src/router/components/form/validation/schema.js
+++ b/ui/src/router/components/form/validation/schema.js
@@ -221,7 +221,7 @@ const resourceRequestSchema = (maxAllowedReplica) =>
         maxAllowedReplica,
         "Max Replicas value has exceeded allowed number of replicas: ${max}"
       )
-      .when("min_replica", (minReplica, schema) =>
+      .when("min_replica", ([minReplica], schema) =>
         minReplica === 0
           ? schema.positive("Max Replica should be positive")
           : schema


ui/src/router/components/form/validation/schema.js
```diff
@@ -221,7 +221,7 @@ const resourceRequestSchema = (maxAllowedReplica) =>
         maxAllowedReplica,
         "Max Replicas value has exceeded allowed number of replicas: ${max}"
       )
-      .when("min_replica", (minReplica, schema) =>
+      .when("min_replica", ([minReplica], schema) =>
```
tiopramayudi, line 224: what is the impact if we don't make this change?

tiopramayudi (approved):
Add question, the rest is LGTM!

ui/src/router/components/form/validation/schema.js
deadlycoconuts: Ah, the direct impact would be `minReplica` being a list with 1 integer (the actual `min_replica` value, e.g. `[0]`), which will cause the comparison operator `minReplica === 0` to fail, thus allowing `max_replica` to be set as 0. In other words, our users will get to set min_replica=max_replica=0 which isn't a big problem but probably something we can block upfront.

tiopramayudi (approved):

deadlycoconuts:
Thanks for the second approval @tiopramayudi 😂

PR #392, opened 2024-09-06
chore: add managed by label to namespace

bthari:
# Context
When creating a namespace object in Kubernetes, add label “{prefix}/managed: true”. This will help when we want to modify resources created by CaraML services in a shared cluster, e.g. configuring log for the managed namespace only

# Changes

- add `NamespaceLabelPrefix` in configuration, so `InitKubernetesLabeller` can use the configured namespace prefix
- add the label when creating namespace


tiopramayudi (approved):
LGTM! Thanks @bthari

PR #394, opened 2024-09-18
feat(api): Refactor API server to introduce support for Docker image registries and s3 bucket support

deadlycoconuts:
## Note 🚨
This PR should not be merged without the changes in https://github.com/caraml-dev/mlp/pull/116 being merged, published and imported as dependencies first. This PR is currently only using a branch of a fork (the source of the quoted PR) of the MLP repository.

## Context
In order to provide support for using image registries that use [Docker registry credentials](https://docs.docker.com/reference/cli/docker/login/#configure-the-credential-store) as well as [AWS S3-based blob storage services](https://docs.aws.amazon.com/code-library/latest/ug/go_2_s3_code_examples.html), this PR refactors the API server to support these additional image registry and bob storage options. More concretely, these are the following changes made:

- Set up the workflow needed to allow platform maintainers to configure the image registry that the API server will push images to (Docker or Google Cloud/Artifact Registry) as well as the blob storage service that it should read ensembler artifacts from/write files to (S3-based store/Google Cloud Storage)
- Allow the API server to access a configured Docker registry to check if an image is available
- Allow Kaniko jobs spun up by the API server to use load ensembler artifacts from a configured S3-based store when building model images
- Allow Kaniko jobs spun up by the API server to push images to the configured Docker registry

## Modifications
- `api/turing/api/appcontext.go` - Add additional argument to be passed to the job builder so that it knows what file blob type storage it needs to use
- `api/turing/config/config.go` - Introduce new configs for platform maintainers to specify the `PushRegistryType` and the `DockerCredentialSecretName`
- `api/turing/imagebuilder/imagebuilder.go` - Make changes to the image builder to configure the Kaniko job spec correctly depending on the selected registry type and blob storage type
- `engines/pyfunc-ensembler-job/Dockerfile` -  Add steps to the base batch ensembler image to install the AWS CLI
- `engines/pyfunc-ensembler-job/app.Dockerfile` - Add steps to the batch ensembler docker image to authenticate and pull ensembler artifacts correctly depending on the configured blob storage type
- `engines/pyfunc-ensembler-service/Dockerfile` - Add steps to the base real-time ensembler image to install the AWS CLI
- `engines/pyfunc-ensembler-service/app.Dockerfile` - Add steps to the real-time ensembler docker image to authenticate and pull ensembler artifacts correctly depending on the configured blob storage type
- `infra/charts/turing/values.yaml` - Added configs needed for the e2e test in the CICD job

## Some tiny caveats found
- Adding the AWS CLI and building an ensembler image using Kaniko makes Kaniko OOM is it only has 4GB of memory set as its limits
- The current naming convention of ensembler images (see [this](https://github.com/caraml-dev/turing/blob/358bea86dc00880b4ccb9256e843a2bffad073da/api/turing/imagebuilder/ensembler.go#L81) and [this](https://github.com/caraml-dev/turing/blob/358bea86dc00880b4ccb9256e843a2bffad073da/api/turing/imagebuilder/ensembler.go#L44)), with its slashes `/`, seems to not work when pushing to Docker Hub, as Docker Hub doesn't seem to accept slashes in a repository (image)'s name.


api/turing/config/config.go
```diff
@@ -426,8 +430,8 @@ type MLPConfig struct {
 }
 
 type MlflowConfig struct {
-	TrackingURL         string
-	ArtifactServiceType string
+	TrackingURL         string `validate:"required_if=ArtifactServiceType gcs ArtifactServiceType s3"`
+	ArtifactServiceType string `validate:"required,oneof=nop gcs s3"`
```
bthari, line 434: I'm wondering what will happen if we have a different pair of `ArtifactServiceType` and `PushRegistryType`. Like, `PushRegistryType=docker` & `ArtifactServiceType=gcs`, can still it run? I assume it can’t because then we will not have the `GOOGLE_APPLICATION_CREDENTIALS` env var for the `gcloud auth activate-service-account` command in [dockerfile](https://github.com/caraml-dev/turing/blob/2f5cc7c83eeb5bf8d1c7c79771af0453e869ab60/engines/pyfunc-ensembler-service/app.Dockerfile#L18), but CMIIW. 

Should we validate this config first?

bthari (commented):
Thank you for this PR and the detailed changelog! A bit of question, you mentioned that Docker Hub doesn’t support image name with slash, are we planning to change the name format later, or what's the plan for it?

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -288,12 +293,23 @@ func (ib *imageBuilder) createKanikoJob(
...
 		fmt.Sprintf("--build-arg=BASE_IMAGE=%s", baseImage),
+		fmt.Sprintf("--build-arg=MLFLOW_ARTIFACT_STORAGE_TYPE=%s", ib.artifactServiceType),
 		fmt.Sprintf("--build-arg=FOLDER_NAME=%s", folderName),
 		fmt.Sprintf("--destination=%s", imageRef),
 		"--cache=true",
 		"--single-snapshot",
 	}
 
+	if ib.artifactServiceType == "s3" {
+		kanikoArgs = append(
```
tiopramayudi, line 304: can we read it from environment files?

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -288,12 +293,23 @@ func (ib *imageBuilder) createKanikoJob(
 		fmt.Sprintf("--context=%s", ib.imageBuildingConfig.KanikoConfig.BuildContextURI),
 		fmt.Sprintf("--build-arg=MODEL_URL=%s", artifactURI),
 		fmt.Sprintf("--build-arg=BASE_IMAGE=%s", baseImage),
+		fmt.Sprintf("--build-arg=MLFLOW_ARTIFACT_STORAGE_TYPE=%s", ib.artifactServiceType),
```
tiopramayudi, line 296: can you extract out the kaniko build argument code in separate function?

api/turing/config/config.go
deadlycoconuts: > I'm wondering what will happen if we have a different pair of `ArtifactServiceType` and `PushRegistryType`. Like, `PushRegistryType=docker` & `ArtifactServiceType=gcs`, can still it run?

Yes it's expected that such a configuration will still run. In other words, the `ArtifactServiceType` and `PushRegistryType` are supposed to be independent from each other. The `ArtifactServiceType` only specifies the blob storage type where Merlin model artifacts are stored whereas `PushRegistryType` specifies the image registry Merlin should push a built model image to.

> I assume it can’t because then we will not have the `GOOGLE_APPLICATION_CREDENTIALS` env var for the `gcloud auth activate-service-account` command in [dockerfile](https://github.com/caraml-dev/turing/blob/2f5cc7c83eeb5bf8d1c7c79771af0453e869ab60/engines/pyfunc-ensembler-service/app.Dockerfile#L18)
Ah yes, good catch! If `PushRegistryType=docker` & `ArtifactServiceType=gcs`, then we still need to have some form of GCS credentials in the kaniko image builder (either via a service account or via those `GOOGLE_APPLICATION_CREDENTIALS` but I've removed them from this change [here](https://github.com/caraml-dev/turing/pull/394/files#diff-4fbe33f74c84401285db6174c70b566fd9be74c9fcb30ff90a20019ff21f25e3R326). Thanks a lot for spotting this! Let me fix this!

deadlycoconuts: I've refactored the code a little (just like in Merlin) in order to add the GCP credentials (via a mounted secret and env var) if `gcs` is configured as an artifact store (this is independent from whatever is configured as the push registry (docker or gcr). You can find those changes here: https://github.com/caraml-dev/turing/pull/394/files#diff-4fbe33f74c84401285db6174c70b566fd9be74c9fcb30ff90a20019ff21f25e3R382

api/turing/imagebuilder/imagebuilder.go
deadlycoconuts: Okay I've done this like in Merlin [here](https://github.com/caraml-dev/merlin/pull/605#discussion_r1807766235) where there's a config field in the config value files where platform-level additional kaniko args can be added, instead of using an env file to keep things consistent. If you think an env file is cleaner we can change it to that too.

api/turing/imagebuilder/imagebuilder.go
deadlycoconuts: Alright, I just refactored all the credentials-related modifications to kaniko args, volumes, etc. into separate functions (just like the changes I did for Merlin. I didn't create a helper function to add non-credential related build arguments though.

deadlycoconuts:
> A bit of question, you mentioned that Docker Hub doesn’t support image name with slash, are we planning to change the name format later, or what's the plan for it?

@bthari hmm I'm thinking/hoping that it might just be a problem specific to Docker Hub but if we aren't pushing to Docker Hub then it should be okay. However, I think a more comprehensive solution would be to allow the naming convention to be configurable on a platform-level (via the API server configs) but that might need a little more time to expose as a config.

deadlycoconuts:
Thanks a lot for the comments @tiopramayudi and @bthari ! I should've covered and addressed all of the outstanding comments. If all looks good feel free to approve it again otherwise let me know what additional changes we should include and I'll get back to it. Thanks once again!

api/turing/config/config.go
bthari: Thank you so much for clarifying this and making the changes! Just one last question (promise!), in [here](https://github.com/caraml-dev/turing/blob/611615f299ee1c0ab8143a415c3b833deccd7c21/api/turing/imagebuilder/imagebuilder.go#L388-L409), it will only mount the google credential if `artifactServiceType == gcs`. But if `artifactServiceType == gcs & pushRegistryType == docker`, don't we need both the gcs & docker credential, or not?

deadlycoconuts: Ah yes that's right! Nice catch again! Hahaha thanks for giving it one last check (I must've been falling asleep when implementing that change) ><" I'll update the `if-else` block and turn it into 2 `if` blocks.

api/turing/imagebuilder/imagebuilder.go
```diff
@@ -384,6 +370,58 @@ func (ib *imageBuilder) createKanikoJob(
 	)
 }
 
+func (ib *imageBuilder) configureKanikoArgsToAddCredentials(kanikoArgs []string) []string {
+	if ib.imageBuildingConfig.KanikoConfig.PushRegistryType == "gcr" || ib.artifactServiceType == "gcs" {
```
tiopramayudi, line 374: sorry i missed this previously, but can you add the `gcr` and `gcs` in the constant variable?

engines/pyfunc-ensembler-job/Dockerfile
```diff
@@ -1,8 +1,13 @@
 FROM apache/spark-py:v3.1.3
 
+# Switch to user root so we can add additional jars and configuration files.
+USER root
```
tiopramayudi, line 4: q: won't it possible using the user 0?

tiopramayudi (approved):
Adding one more feedback, the rest is LGTM! Feel free to merge this, thanks @deadlycoconuts !

api/turing/imagebuilder/imagebuilder.go
deadlycoconuts: Ah good point, done! :D

engines/pyfunc-ensembler-job/Dockerfile
deadlycoconuts: Ah oops my bad, didn't notice there was user 0 line below ><" I removed the `USER 0` line instead so as to make this file look similar to the batch predictor Dockerfile in Merlin: https://github.com/caraml-dev/merlin/blob/9e78611f3343cda6fa7f82b72b96051d065214ad/python/batch-predictor/docker/base.Dockerfile#L18

deadlycoconuts:
Thanks for the re-review @tiopramayudi! I'll merge this soon! :D

PR #395, opened 2024-10-09
feat: add router and ensembler webhook

bayu-aditya:
# Context
This PR to adding webhooks for Router and Ensembler for this action and event type.
- Router Webhook
  - Creating Router (`OnRouterCreated`)
  - Updating Router (`OnRouterUpdated`)
  - Deleting Router (`OnRouterDeleted`)
- Router Deployment Webhook
  - Deploying Router (`OnRouterDeployed`)
  - Undeploying Router (`OnRouterUndeployed`)
- Ensembler Webhook
  - Creating Ensembler (`OnEnsemblerCreated`)
  - Updating Ensembler (`OnEnsemblerUpdated`)
  - Deleting Ensembler (`OnEnsemblerDeleted`)

For router, Turing will send webhook event with this body payload
```json5
{
  "event_type": "OnRouterCreated",

  // models.Router
  "router": {
    "id": 123,
    "created_at": "string",
    "updated_at": "string",
    "project_id": 456,
    "environment_name": "string",
    "name": "string",
    "status": "string",
    "endpoint": "string",
    "config": {},
    "monitoring_url": "string"
  }
}
```

Then for router deployment, Turing will send webhook event with this body payload
```json5
{
  "event_type": "OnRouterDeployed",
  "project_id": 456,

  // models.RouterVersion
  "router_version": {
    "id": 123,
    "created_at": "string",
    "updated_at": "string",

    "router": {},
    "version": 123,
    "status": "deployed"
  }
}
```

And for ensembler, Turing will send webhook event with this body payload
```json5
{
  "event_type": "OnEnsemblerCreated",

  // models.GenericEnsembler or models.PyFuncEnsembler
  "ensembler": { 
    "id": 123,
    "created_at": "string",
    "updated_at": "string",
    "project_id": 456,
    "type": "string",
    "name": "string"
  }
}
```

# Changes
- Create webhook client package and unittest
- Add webhook client in `BaseController`
- Invoke webhook in this HTTP handler functions:
  - POST `/projects/{project_id}/routers`
  - PUT `/projects/{project_id}/routers/{router_id}`
  - DELETE `/projects/{project_id}/routers/{router_id}`
  - POST `/projects/{project_id}/routers/{router_id}/deploy`
  - POST `/projects/{project_id}/routers/{router_id}/undeploy`
  - POST `/projects/{project_id}/ensemblers`
  - PUT `/projects/{project_id}/ensemblers/{ensembler_id}`
  - DELETE `/projects/{project_id}/ensemblers/{ensembler_id}`


api/turing/api/ensemblers_api.go
```diff
@@ -149,14 +161,20 @@ func (c EnsemblersController) UpdateEnsembler(
 		}
 	}
 
+	// call webhook for ensembler update event
+	if errWebhook := c.webhookClient.TriggerEnsemblerEvent(ctx, webhook.OnEnsemblerUpdated, ensembler); errWebhook != nil {
```
tiopramayudi, line 165: what happens if the trigger is failing but the operation is successful?

api/turing/webhook/webhook.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"context"
+
+	"github.com/caraml-dev/mlp/api/pkg/webhooks"
+
+	"github.com/caraml-dev/turing/api/turing/models"
+)
+
+type Client interface {
+	TriggerRouterEvent(ctx context.Context, eventType webhooks.EventType, router *models.Router) error
```
tiopramayudi, line 12: what happen if this method called with eventType `OnEnsemblerCreated` ?

api/turing/webhook/webhook.go
```diff
@@ -0,0 +1,73 @@
...
+	TriggerEnsemblerEvent(ctx context.Context, eventType webhooks.EventType, ensembler models.EnsemblerLike) error
```
tiopramayudi, line 13: what happen if this method is called with eventType `OnRouterDeployed` ?

api/turing/api/ensemblers_api.go
bayu-aditya: Webhook will retry to send the event because the mechanism already implemented in MLP package https://github.com/caraml-dev/mlp/blob/9b920121bf6af8bab1922efdaaa8c0411afbd522/api/pkg/webhooks/client.go#L81-L110

api/turing/webhook/webhook.go
bayu-aditya: Ah got it, thanks for your detail, will be updated

api/turing/webhook/webhook.go
bayu-aditya: Fixing in https://github.com/caraml-dev/turing/pull/395/commits/7760b00e620e5386fa545f67363f000043a48e2f

api/turing/webhook/webhook.go
bayu-aditya: Fixing in https://github.com/caraml-dev/turing/pull/395/commits/7760b00e620e5386fa545f67363f000043a48e2f

api/turing/api/ensemblers_api.go
```diff
@@ -80,14 +86,20 @@ func (c EnsemblersController) CreateEnsembler(
 		return InternalServerError("unable to save the ensembler", err.Error())
 	}
 
+	// call webhook for ensembler creation event
+	if errWebhook := c.webhookClient.TriggerEnsemblerEvent(ctx, webhook.OnEnsemblerCreated, ensembler); errWebhook != nil {
+		log.Warnf("Error triggering webhook for event %s, %v", webhook.OnEnsemblerCreated, errWebhook)
```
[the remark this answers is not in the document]
bayu-aditya: fixed 🙏

api/turing/api/routers_api.go
```diff
@@ -119,6 +123,11 @@ func (c RoutersController) CreateRouter(
 		return InternalServerError("unable to create router", strings.Join(errorStrings, ". "))
 	}
 
+	// call webhook for router creation event
+	if errWebhook := c.webhookClient.TriggerRouterEvent(ctx, webhook.OnRouterCreated, router); errWebhook != nil {
+		log.Warnf("Error triggering webhook for event %s, %v", webhook.OnRouterCreated, errWebhook)
```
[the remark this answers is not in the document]
bayu-aditya: fixed 🙏

api/turing/api/routers_api.go
```diff
@@ -267,6 +298,11 @@ func (c RoutersController) DeployRouter(
 		return NotFound("router version not found", err.Error())
 	}
 
+	// call webhook for router deployment event
+	if errWebhook := c.webhookClient.TriggerRouterEvent(ctx, webhook.OnRouterDeployed, router); errWebhook != nil {
```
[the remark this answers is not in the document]
bthari: I think need to add this event on router creation/update too, because in that endpoint there's also a mechanism to deploy a router, or maybe add this event inside the function of `deployOrRollbackRouter()` 🤔 What do you think?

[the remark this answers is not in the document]
bayu-aditya: Fixed, thanks @bthari for your detail 🙏

api/turing/webhook/constant.go
```diff
@@ -0,0 +1,41 @@
+package webhook
+
+import (
+	"errors"
+
+	"github.com/caraml-dev/mlp/api/pkg/webhooks"
+)
+
+var (
+	OnRouterCreated    = webhooks.EventType("OnRouterCreated")
```
tiopramayudi, line 10: the event type format is different with the one in merlin, can we agree on the same format?

bayu-aditya: Fixed on https://github.com/caraml-dev/turing/pull/395/commits/5ac941e0028de90e2cea6279ceb9fc89c7626b52, thank you for your detail 🙏

api/turing/api/ensemblers_api.go
```diff
@@ -62,12 +64,16 @@ func (c EnsemblersController) GetEnsembler(
...
 	vars RequestVars,
 	body interface{},
 ) *Response {
-	var errResp *Response
-	var project *mlp.Project
+	var (
+		ctx     = req.Context()
+		errResp *Response
+		project *mlp.Project
+	)
```
deadlycoconuts, line 75: Super mega nit: why did you bother with grouping the initialisation of `ctx` with the other variables into the brackets `()` here but not for the others? 😅

api/turing/api/routers_api.go
```diff
@@ -182,27 +204,45 @@ func (c RoutersController) UpdateRouter(_ *http.Request, vars RequestVars, body
...
 		}
+
+		// call webhook for router deployment event
+		if errWebhook := c.webhookClient.TriggerRouterEvent(ctx, webhook.OnRouterDeployed, router); errWebhook != nil {
+			log.Warnf(
+				"Error triggering webhook for event %s, router id: %d, %v",
+				webhook.OnRouterDeployed, router.ID, errWebhook,
+			)
+		}
 	}()
```
deadlycoconuts, line 228: Unlike Merlin, when Turing updates a router, it undeploys the old version and deploys the old one in a single operation. As such, we might need to insert another webhook call here for the `OnRouterUndeployed` for the old router version, preferably just before the webhook call for the `OnRouterDeployed` event. 

You might have to pay attention to the `CurrRouterVersion` [field](https://github.com/caraml-dev/turing/blob/358bea86dc00880b4ccb9256e843a2bffad073da/api/turing/models/router.go#L38) that's being sent to the webhook server (you'd want to send a request body with the old router version for the undeployment event and another request body with the new router version with the deployment event).

An additional point to take note is that you might want to log the router version ID in addition to the router ID so that's it's easier for debugging.

deadlycoconuts (changes requested):
Thanks a lot for the PR! I've left a couple of tiny comments here and there but the most important one is the one about calling the webhooks whenever a router is **_updated_**, because that involves 2 steps, the deployment of a new router version and subsequently, the undeployment of the previous router version.

Another (equally tricky) change you might've left out is to add webhook calls to the 'DeployRouterVersion' [method](https://github.com/caraml-dev/turing/blob/358bea86dc00880b4ccb9256e843a2bffad073da/api/turing/api/router_versions_api.go#L171). This is similarly (or even more) confusing because there are two possible scenarios here - 1) there's just a simple deployment of a specific router version (no undeployment involved) and 2) there's a deployment of the specified router version + the undeployment of an existing deployed router version. Scenario 1 happens when there are no existing router versions deployed but scenario 2 happens when another router version was already deployed when that API endpoint gets called. You might need to poke a little more to find out some way to determine if you need 1 additional webhook call for the undeployment event for this API endpoint.

api/go.mod
```diff
@@ -251,6 +252,7 @@ replace (
 
 	github.com/caraml-dev/turing/engines/experiment => ../engines/experiment
 	github.com/caraml-dev/turing/engines/router => ../engines/router
+	github.com/go-gota/gota => github.com/gojekfarm/gota v0.12.1-0.20230221101638-6cd9260bd598
```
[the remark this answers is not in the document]
bayu-aditya: Because I got this error while install dependency
image

api/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/GoogleCloudPlatform/spark-on-k8s-operator v0.0.0-20221025152940-c261df66a006
 	github.com/antihax/optional v1.0.0
 	github.com/caraml-dev/merlin v0.0.0
-	github.com/caraml-dev/mlp v1.12.0
+	github.com/caraml-dev/mlp v1.13.2-rc2
```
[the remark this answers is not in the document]
bayu-aditya: Fixed, thanks for your detail 🙏

api/go.mod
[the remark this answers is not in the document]
bthari: I think this is an error from intellij only, I got this too, idk why tho 😅 but the app can run fine without explicitly replace the `gota` in `go.mod`, I just run `go mod vendor` manually (not using intellij dependencies sync). Does this not work for you?

api/turing/webhook/webhook.go
```diff
@@ -0,0 +1,120 @@
...
+	if !w.isEventConfigured(eventType) {
+		return nil
+	}
+
+	if err := w.manager.InvokeWebhooks(
+		ctx,
+		eventType,
+		body,
+		webhooks.NoOpCallback,
+		webhooks.NoOpErrorHandler,
```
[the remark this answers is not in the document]
deadlycoconuts: Just to confirm if I understood this correctly, are you referring to the case where we might be having different `onSuccess` and `onError` handlers for different `eventType`-s?

api/turing/webhook/request.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+type routerRequest struct {
+	EventType webhooks.EventType `json:"event_type"`
+	Router    *models.Router     `json:"router"`
+}
+
+type routerDeploymentRequest struct {
+	EventType     webhooks.EventType    `json:"event_type"`
+	ProjectID     uint                  `json:"project_id"`
```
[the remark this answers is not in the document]
deadlycoconuts: Good point, `routerVersion.router` should have a project id field in `routerVersion.router.ProjectID`. I'll remove this field from the request struct.

api/turing/webhook/webhook.go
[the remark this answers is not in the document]
deadlycoconuts: This is resolved by removing the use-case specific webhook functions `TriggerRouterEvent`, `TriggerRouterDeploymentEvent` and `TriggerEnsemblerEvent` and instead replacing them with the generic `TriggerWebhooks` function.

api/turing/api/routers_api.go
```diff
@@ -305,6 +345,14 @@ func (c RoutersController) UndeployRouter(
 		return InternalServerError("unable to undeploy router", err.Error())
 	}
 
+	// call webhook for router un-deployment event
+	if errWebhook := c.webhookClient.TriggerRouterEvent(ctx, webhook.OnRouterUndeployed, router); errWebhook != nil {
```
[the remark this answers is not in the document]
deadlycoconuts: This is resolved by removing the use-case specific webhook functions `TriggerRouterEvent`, `TriggerRouterDeploymentEvent` and `TriggerEnsemblerEvent` and instead replacing them with the generic `TriggerWebhooks` function. Each webhook event is no longer tied to a specific webhook function and can be used directly with the `TriggerWebhooks` function.

deadlycoconuts:
Okay I've done some not-so-minor changes to the changes in this PR. Here's a summary of them:

1. Removed the use-case specific webhook functions `TriggerRouterEvent`, `TriggerRouterDeploymentEvent` and `TriggerEnsemblerEvent` and instead replaced them with the **generic `TriggerWebhooks` function**
   - Webhooks are now called directly with the `TriggerWebhooks` function without any additional checks to ensure that the passed event type 'belongs' to the correct 'use-case' (i.e. router, router deployment, ensembler)
   - This simplifies the `webhook` package in Turing and makes it leaner
2. Refactored and repositioned the webhook events such that **each event corresponds to an explicit endpoint handler** on the API server
   - This means **only the webhook for a specific event** (e.g. OnRouterUpdated) gets called when its corresponding endpoint handler successfully executes completely (to ensure atomicity). If the endpoint handler errors out midway during execution, the webhook will not be called.
   - Previously, multiple webhooks (e.g. deploying the new router version, undeploying the old router version, undeploying a failed new version, etc.) can potentially be called even during the execution of a single endpoint (e.g. OnRouterUpdated). While this allows for simpler and smaller-sized webhook handlers to be designed on the webhook service, as well as similar events to be reused, this unfortunately makes it super confusing to know where to insert these webhooks in the API server, to easily understand how and which webhook calls will be made, so as to understand the entire effect of all these webhook calls (e.g. what happens if out of 3 webhook calls expected from a single endpoint's execution, only 2 get triggered because of an error midway?).
   - This change means that more effort and thought needs to go into implementing each webhook handler on the webhook service, but I believe this will make it easier for maintenance

That's all for the changes I have in this PR, please help me take a look and leave any comments you may have! 🙏 Thanks!

api/go.mod
```diff
@@ -267,7 +268,6 @@ replace (
 
 	github.com/caraml-dev/merlin => github.com/caraml-dev/merlin/api v0.0.0-20240313065547-6778bd14c119
 	github.com/caraml-dev/merlin-pyspark-app => github.com/caraml-dev/merlin/python/batch-predictor v0.0.0-20240313065547-6778bd14c119
-	github.com/caraml-dev/mlp => github.com/deadlycoconuts/mlp v0.0.0-20240917090435-d94d92572eac
```
deadlycoconuts, line 270: I unfortunately merged this change in my previous PR [here](https://github.com/caraml-dev/turing/pull/394/files#diff-ba0784a950902285760f839a4ad720a381b2981e8a67147f5f43f7508e6343ed), so I'm removing this line in this PR.

api/go.mod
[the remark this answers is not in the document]
deadlycoconuts: Yeah I seem to get that error too on my IDE 😅 and I also run `go mod vendor` manually 🧐 Not sure why this happens though, but I'll just remove it for the time being unless it causes some problem on the app.

api/turing/webhook/webhook.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (w webhook) TriggerWebhooks(ctx context.Context, eventType webhooks.EventType, body interface{}) error {
+	if !w.isEventConfigured(eventType) {
+		return nil
+	}
+
+	return w.webhookManager.InvokeWebhooks(
+		ctx,
+		eventType,
+		body,
```
bthari, line 66: Can we also add the event name in the body? I imagine for some webhooks it might be easier to have one endpoint for several events, then differentiate what to do based on `eventType`. Like with our sync use case, the webhook might simply put the request in a queue (so no need to have several endpoint), and there would be another worker that will process the queue later on (but it needs to know the `eventType`

api/turing/webhook/webhook.go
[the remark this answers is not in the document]
bthari: Yes, I wanted to comment the same. I think we can always get back to this when Turing have an actual usecase where some of the events use different callback, but for now we can keep it. Thank you for addressing this!

api/turing/webhook/webhook.go
deadlycoconuts: Yeah sure that's possible. Should we be making this change for Merlin too?

bthari: Merlin already include the `eventType` in the payload body. It will be nice if it can be added by MLP side, but isn't it going to change the structure of the payload, like how do we insert new field to an `interface{}` variable? 🤔 Or we can modify Turing without changing MLP, it will be okay too for me

deadlycoconuts: > Merlin already include the eventType in the payload body.

Ah yes you're right I just saw that 😮  

> but isn't it going to change the structure of the payload, like how do we insert new field to an interface{} variable? 🤔

Hmm, what do you mean? I was thinking of just adding something like this (just as in Merlin) in the MLP package (assuming that Merlin hadn't already added the event type to the payload `p`):

```go
type WebhookRequest struct {
	EventType EventType     `json:"event_type"`
	Data      interface{} `json:"data"`
}

...

func (w *SimpleWebhookManager) InvokeWebhooks(
	ctx context.Context,
	event EventType,
	p interface{},
	onSuccess func([]byte) error,
	onError func(error) error,
) error {
	...
	originalPayload, err := json.Marshal(p)

	newRequestBody := &WebhookRequest{
		EventType: event,
		Data: p,
	}
	originalPayload, err := json.Marshal(newRequestBody)
	...
}

```

But yeah, since Merlin already includes the event type in the payload body, I'll just add it here to Turing without changing MLP then :D

bthari (approved):
LGTM and thank you so much for the changes @deadlycoconuts!

deadlycoconuts:
Thanks a lot for looking at this PR again! Thanks a lot @bayu-aditya also for most of the work! :D I'll merge this PR now!

deadlycoconuts (approved):

deadlycoconuts:
Damn I had to approve this PR myself even though I have pushed changes to it (that's because I selected the 'request changes' option when reviewing this PR 😑

PR #396, opened 2024-11-11
Switch to use <EuiSuperselect /> component to list docker registery in router deployment

vinoth-gojek:
At the time of writing, clicking Docker Image Registry list in Turing UI leads to an error page. This happens because of "Maximum update depth exceeded" which happens usually as infinte re-rendering is blocked. While toggling the popover, the state re-rendering is causing infinte reload. I couldn't find any extra value addition in using "EuiPopover" over the simple EuiSuperSelect. 

The MR fixes the issue by choosing switching from "EuiPopover" to use "EuiSuperSelect" component

Test plan - https://drive.google.com/file/d/1CLl-RoVwO8dhiDmv8YUMJRNT3IJdMiK3/view?usp=sharing (limited access)

### DIFF
--- a/ui/src/components/form/docker_image_combo_box/DockerRegistryPopover.js
+++ b/ui/src/components/form/docker_image_combo_box/DockerRegistryPopover.js
@@ -1,14 +1,14 @@
-import {  EuiSuperSelect } from "@elastic/eui";
+import { EuiSuperSelect } from "@elastic/eui";
 
 export const DockerRegistryPopover = ({ value, registryOptions, onChange }) => {
+
   return (
     <EuiSuperSelect
-          fullWidth
-          options={registryOptions}
-          valueOfSelected={value}
-          itemLayoutAlign="top"
-          onChange={onChange}
-          isInvalid={false}
-        />
+      fullWidth
+      options={registryOptions}
+      valueOfSelected={value}
+      itemLayoutAlign="top"
+      onChange={onChange}
+    />
   );
 };


deadlycoconuts:
Thanks for this PR! Hmm do you have any idea what's causing the infinite re-rendering? The popover's supposed to have an effect that's similar to what's shown in the video below:

https://github.com/user-attachments/assets/bf9078f1-a1d5-4b3d-90f6-36f862e22daf

You can take a look at the code snippet [here](https://github.com/caraml-dev/turing/blob/39865c08a53ac25018ad60488eb7e3a7c34cd59e/ui/src/router/components/request_field_source/FieldSourceFormLabel.js#L36) corresponding to that part above (which looks super similar to what you're working on) and maybe you might figure out what's causing the popover to break 🤔

vinoth-gojek:
> ###
> **[deadlycoconuts](/deadlycoconuts) ** commented [Nov 14, 2024](#issuecomment-2475299909)

@deadlycoconuts yeah valid point, I was unsuccessful cornering the setstate which is causing the exception. I suspect, the dropdown updates registry, on change of registry, it re-renders the parent component, but as mentioned I was not able to corner it. So I moved away to use "EuiSuperselect" which helped

```
Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
    at checkForNestedUpdates (http://localhost:3001/turing/static/js/vendors-node_modules_react-dom_index_js.chunk.js:23602:15)
```

vinoth-gojek:
Attaching more manual testing details for adding new registry

1. New registry could be configured from [privateDockerRegistries config](https://source.golabs.io/data-science/mlp/vino/-/blob/master/environments/gojek/dev/clusters/d-gods-id-products/meta/caraml/local.yaml?ref_type=heads#L719)
2. Adding a new registry will look like - [image](https://drive.google.com/drive/folders/1-svyXCH6VKdneZKI8PlEYGXWi1uMvQso)
3. Successful deployment of router with new registry with publicly available docker image - [link](https://console.d.ai.golabs.io/turing/projects/150/routers/1099/edit)

ui/src/components/form/docker_image_combo_box/DockerRegistryPopover.js
```diff
@@ -1,40 +1,14 @@
...
-      panelPaddingSize="s">
-      <EuiContextMenu initialPanelId={0} panels={panels} />
-    </EuiPopover>
+    <EuiSuperSelect
+          fullWidth
+          options={registryOptions}
+          valueOfSelected={value}
+          itemLayoutAlign="top"
+          onChange={onChange}
+          isInvalid={false}
```
deadlycoconuts, line 11: Can we remove this prop since its default value is `false`?

ui/src/components/form/docker_image_combo_box/DockerRegistryPopover.js
```diff
@@ -1,40 +1,14 @@
-import React, { useState } from "react";
-import { EuiButtonEmpty, EuiContextMenu, EuiPopover } from "@elastic/eui";
-import { flattenPanelTree } from "@caraml-dev/ui-lib";
+import {  EuiSuperSelect } from "@elastic/eui";
```
deadlycoconuts, line 1: nit: there's an extra space ` ` before EuiSuperSelect

ui/src/components/form/docker_image_combo_box/DockerRegistryPopover.js
```diff
@@ -1,40 +1,14 @@
...
-      closePopover={togglePopover}
-      panelPaddingSize="s">
-      <EuiContextMenu initialPanelId={0} panels={panels} />
-    </EuiPopover>
+    <EuiSuperSelect
+          fullWidth
+          options={registryOptions}
+          valueOfSelected={value}
+          itemLayoutAlign="top"
+          onChange={onChange}
```
deadlycoconuts, line 10: nit: Can we keep the indentation style the same as before?

deadlycoconuts (approved):
Thanks a lot for finding an alternative component from the EUI library to replace the EuiPopover/EuiContextMenu combination that somehow isn't working. I believe there's some issue in the parent component using the DockerRegistryPopover that's probably triggering unnecessary re-renders but we'll probably have to come back to this another time :/

vinoth-gojek:
hi @deadlycoconuts thanks for sharing the comments. I have addressed your comments with the latest commit. I have kept those open for you to take a look again. Thanks and let me know your thoughts in case there are any more to address!

deadlycoconuts:
Thanks a lot for addressing them! I think we're good to merge this PR! Once this is done we can cut a patch release like this one here https://github.com/caraml-dev/turing/releases/tag/v1.20.2!

PR #397, opened 2024-11-13
chore: change webhook payload structure

bthari:
## Summary

Changing the structure of the payload send from Turing to a webhook call, so we have the same structure across [Merlin](https://github.com/caraml-dev/merlin/blob/main/api/webhook/request.go) and Turing. From
```
{
    "event_type": "on-model-endpoint-created",
    "data": {} // the data being either ensembler/router/router_version 
}
```
to
```
{
    "event_type": "on-model-endpoint-created",
    "data": {
        "ensembler": {}, // omit if empty
        "router": {}, // omit if empty
        "router_version": {} // omit if empty
    }
}
```
I make the payload with this structure in case in the future we want to put several object in the payload (instead of just one object like now). I’m open to any advice if you have other opinions

### DIFF
--- a/api/turing/webhook/request.go
+++ b/api/turing/webhook/request.go
@@ -2,9 +2,31 @@ package webhook
 
 import (
 	"github.com/caraml-dev/mlp/api/pkg/webhooks"
+	"github.com/caraml-dev/turing/api/turing/models"
 )
 
 type Request struct {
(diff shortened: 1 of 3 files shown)


deadlycoconuts (approved):
Ah oops I didn't know you needed it to be in exactly the same schema as the request body in Merlin, my bad. Thanks for fixing this!

PR #399, opened 2025-01-03
fix(api): Add configmaps as a resource to the spark driver role rules

deadlycoconuts:
## Context
Similar to https://github.com/caraml-dev/merlin/pull/623, currently, the Spark driver service account that is created for every batch ensembling job does not have permissions to create ConfigMaps that its Spark executor(s) require(s). This is a problem in clusters with RBAC enabled and this PR simply adds the missing rule on ConfigMap resources for the Spark driver service account.

## Modifications
- `api/turing/cluster/spark.go` - Addition of configmaps as an additional rule

### DIFF
--- a/api/turing/cluster/spark.go
+++ b/api/turing/cluster/spark.go
@@ -71,6 +71,18 @@ var DefaultSparkDriverRoleRules = []apirbacv1.PolicyRule{
 			"*",
 		},
 	},
+	{
+		// Allow driver to manage configmaps
+		APIGroups: []string{
+			"", // indicates the core API group
+		},
+		Resources: []string{
+			"configmaps",
+		},
+		Verbs: []string{
+			"*",
+		},
+	},
 }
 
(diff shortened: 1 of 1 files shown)


api/turing/cluster/spark.go
```diff
@@ -71,6 +71,18 @@ var DefaultSparkDriverRoleRules = []apirbacv1.PolicyRule{
 			"*",
 		},
 	},
+	{
+		// Allow driver to manage configmaps
+		APIGroups: []string{
+			"", // indicates the core API group
+		},
+		Resources: []string{
+			"services",
```
mbruner, line 80: should be configmaps

mbruner (changes requested):

api/turing/cluster/spark.go
deadlycoconuts: Ah nice catch!

mbruner (approved):

PR #401, opened 2025-01-16
chore: increase log volume size

mbruner:
In certain environments there is a constraint for logger volume, setting the default value to 20Gi


deadlycoconuts (approved):
Thanks for this change!

PR #405, opened 2025-04-24
feat(sdk): migrate unit test from `urllib3-mock` to `MagicMock`

naufalandika:
# Description
Right now we use `urllib3-mock` to `merlin-sdk`'s http request. But `urllib3-mock` is not maintained anymore since 2015 and doesn't work with python 3.11++ because they still use some functions that's removed in python 3.11. Because we want to support python 3.11++ in our sdk we need to migrate the unit test away from `urllib3-mock` to `MagicMock`

# Modifications
- Remove all function calls related to `urllib3-mock`
- Refactor unit test to use `MagicMock`


vinoth-gojek (approved):
overall lgtm

deadlycoconuts (commented):
Thanks a lot for the extensive changes! I just left a couple of tiny questions but they aren't urgent or anything.

PR #406, opened 2025-06-13
chore: change caraml-auth-google to 0.0.0.post9 to avoid using dev tag

naufalandika:


deadlycoconuts (approved):
Thanks a lot for the change! 🚀

PR #408, opened 2025-07-14
Append enricher response to the turing router response

vinoth-gojek:
Right now, turing passes the enricher response only to the router and experiments call. As a new requirement, the MR enables the enricher response to be passed to the ensembler also

The combinedRouterResp which is the turing router response, currently has response from experiments and the router responses. The enricher response is appended to this combinedRouterResp and this enables turing sdk to read the enricher response from combinedRouterResp

### DIFF
--- a/api/e2e/test/router_e2e_test.go
+++ b/api/e2e/test/router_e2e_test.go
@@ -139,6 +139,7 @@ var _ = DeployedRouterContext("testdata/create_router_nop_logger_proprietary_exp
 					})
 				})
 			})
+
 		})
 
 		ginkgo.Describe("1.4. Undeploying router", func() {
--- a/api/e2e/test/router_e2e_test.go
+++ b/api/e2e/test/router_e2e_test.go
@@ -139,7 +139,6 @@ var _ = DeployedRouterContext("testdata/create_router_nop_logger_proprietary_exp
 					})
 				})
 			})
-
 		})
 
 		ginkgo.Describe("1.4. Undeploying router", func() {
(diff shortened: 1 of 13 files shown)


sdk/turing/ensembler.py
```diff
@@ -117,10 +118,11 @@ def _ensemble_request(self, model_input: Dict[str, Any]) -> Any:
                 input=request_body["request"],
                 predictions=routes_to_response,
                 treatment_config=request_body["response"]["experiment"],
+                enricher_response=request_body["response"]["enricher_response"],
                 headers=model_input["headers"],
             )
         except TypeError as e:
-            if "got an unexpected keyword argument 'headers'" in str(e):
+            if "got an unexpected keyword argument 'enricher_response'" in str(e):
```
vinoth-gojek, line 125: To maintain backward compatibility with older versions of turing router which don't pass enricher_response, the ensemble object is init without enricher_response

With my minimal knowledge, other option would be use enricher_response=request_body["response"].get("enricher_response", {}). I think this would avoid the exception in the first place

api/e2e/test/router_e2e_test.go
```diff
@@ -139,7 +139,6 @@ var _ = DeployedRouterContext("testdata/create_router_nop_logger_proprietary_exp
 					})
 				})
 			})
-
```
deadlycoconuts, line 142: nit: we can remove this

deadlycoconuts (approved):
Thanks a lot for these changes! Really appreciate them! :D

PR #409, opened 2025-07-16
Set ensembler log level to WARNING from INFO

vinoth-gojek:
The default turing ensembler log level is set to INFO but this spams with too many logs thus causing memory consumption in downstream logging platform.

The goal of this very simple MR is to modify this log level to WARNING from INFO - [python logging loglevels](https://docs.python.org/3/library/logging.html#logging-levels)

Also there is an added capability for DS folks to change the ensembler log levels to INFO using env

### DIFF
--- a/engines/pyfunc-ensembler-service/app.Dockerfile
+++ b/engines/pyfunc-ensembler-service/app.Dockerfile
@@ -52,4 +52,4 @@ RUN if [ "${MLFLOW_ARTIFACT_STORAGE_TYPE}" = "gcs" ]; then  \
 ENV FOLDER_NAME=$FOLDER_NAME
 SHELL ["/bin/bash", "-c"]
 ENTRYPOINT . activate ${CONDA_ENV_NAME} && \
-  python -m pyfunc_ensembler_runner --mlflow_ensembler_dir ./${FOLDER_NAME} -l WARNING
+  python -m pyfunc_ensembler_runner --mlflow_ensembler_dir ./${FOLDER_NAME}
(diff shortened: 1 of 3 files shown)


engines/pyfunc-ensembler-service/app.Dockerfile
```diff
@@ -52,4 +52,4 @@ RUN if [ "${MLFLOW_ARTIFACT_STORAGE_TYPE}" = "gcs" ]; then  \
 ENV FOLDER_NAME=$FOLDER_NAME
 SHELL ["/bin/bash", "-c"]
 ENTRYPOINT . activate ${CONDA_ENV_NAME} && \
-  python -m pyfunc_ensembler_runner --mlflow_ensembler_dir ./${FOLDER_NAME} -l INFO
+  python -m pyfunc_ensembler_runner --mlflow_ensembler_dir ./${FOLDER_NAME} -l WARNING
```
anantadwi13, line 55: i prefer we remove this `-l WARNING` flag. since it isn't easy to update the value unless we update the whole container entrypoint
```bash
docker run --entrypoint ". activate ${CONDA_ENV_NAME} && python -m pyfunc_ensembler_runner --mlflow_ensembler_dir ./${FOLDER_NAME} -l DEBUG"
```

engines/pyfunc-ensembler-service/pyfunc_ensembler_runner/__main__.py
```diff
@@ -28,14 +29,14 @@
     dest="log_level",
     choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"],
     help="Set the logging level",
-    default=logging.DEBUG,
+    default=logging.WARNING,
 )
```
anantadwi13, line 33: just remove this `--log-level ` flag and set the log level via env var only (`WARNING` as the default value)

anantadwi13: also, usually the command flag has higher priority than the ENV var. if we replace the flag with ENV, it will create confusion to our users.

anantadwi13 (approved):
LGTM. thanks for fixing this!

PR #413, opened 2025-07-17
fix(sdk): Fix ensembler ensemble method

deadlycoconuts:
## Context
This PR fixes the way the Pyfunc ensembler calls the `ensemble` method, by first checking if the `enricher_response` is first present in the request body, before passing it (or a `None` value by default) to the `ensemble` method, failing which a `KeyError` will be thrown.


deadlycoconuts:
Okay I've added a lot more lines to this PR. Mainly just tests and some more reformatting. Merging this now.

